PR #15, opened 2021-08-05
add julia compat entry

trontrytel:


charleskawczynski (commented):
I might suggest setting the compat to 1.5, since we don't test on 1.3. I'm not sure why some of the checks failed, can you re-force-push?

charleskawczynski (approved):

PR #17, opened 2021-08-11
Julia style logo :)

trontrytel:


charleskawczynski (approved):
omg, I love it! You should share this on Julia slack! Also, how’d you make this? Tikz? This makes me think that Thermodynamics.jl needs a logo revamp!

Issue #20, opened 2021-08-16
Qualify inputs


trontrytel:
As suggested here https://github.com/CliMA/CloudMicrophysics.jl/pull/18#discussion_r689787075

Do this:

```
import Thermodynamics
TD = Thermodynamics
TD.air_temperature(ts)
```
throughout the package.

jakebolewski:
In Julia 1.6 (I'm unsure of 1.5) you can achieve this with `import Thermodynamics as TD`  similar to python.

trontrytel:
resolved

PR #26, opened 2021-08-19
using Blah -> import Blah as BL

trontrytel:
This solves #20

### DIFF
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -3,11 +3,15 @@
 """
 module Common
 
-import Thermodynamics as TD
+import Thermodynamics
+import CLIMAParameters
+import CLIMAParameters.Planet
+import CLIMAParameters.Atmos.Microphysics
 
-import CLIMAParameters as CP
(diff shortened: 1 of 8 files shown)


trontrytel:
Looks like the nice syntax doesn't work in julia 1.5. Will fix that

src/Common.jl
```diff
@@ -3,11 +3,15 @@
...
-
-import CLIMAParameters as CP
-import CLIMAParameters.Planet as CP_planet
-import CLIMAParameters.Atmos.Microphysics as CP_micro
+import Thermodynamics
+import CLIMAParameters
+import CLIMAParameters.Planet
+import CLIMAParameters.Atmos.Microphysics
+
+TD = Thermodynamics
```
jakebolewski, line 11: you want to declare these as `const`

trontrytel: Makes sense! done

charleskawczynski (approved):
LGTM

PR #29, opened 2021-09-02
Add Kappa

trontrytel:
This adds the kappa parameter for describing hygroscopicity of aerosol particles

### DIFF
--- a/src/AerosolActivation.jl
+++ b/src/AerosolActivation.jl
@@ -23,10 +23,12 @@ const CP = CLIMAParameters
 const CP_planet = CLIMAParameters.Planet
 const APS = CP.AbstractParameterSet
 
-export mean_hygroscopicity
+export mean_hygroscopicity_parameter
 export max_supersaturation
+
 export N_activated_per_mode
 export M_activated_per_mode
+
 export total_N_activated
 export total_M_activated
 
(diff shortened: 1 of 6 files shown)


src/AerosolActivation.jl
```diff
@@ -81,6 +90,19 @@ function mean_hygroscopicity(param_set::APS, ad::AM.AerosolDistribution)
...
+    ad::AM.AerosolDistribution{NTuple{N, T}},
+) where {N, T <: AM.Mode_κ}
+
+    return ntuple(length(ad.Modes)) do i
+
+        modei = ad.Modes[i]
+        sum(1:(modei.n_components)) do i
+            modei.vol_mix_ratio[i] * modei.kappa[i]
+        end
+    end
```
charleskawczynski, line 104: I would avoid using a nested `do` loop with both indices being the same--at least for clarity since the behavior depends on the scope rules.

trontrytel: good catch! Thanks!

charleskawczynski (approved):
I can't say much about the physics modeling here, but on the software side, I left one comment about changing indices.

Overall, this looks fine to me with that fix.

trontrytel:
> I can't say much about the physics modeling here, but on the software side, I left one comment about changing indices.
> 
> Overall, this looks fine to me with that fix.

Thanks! Me and @imdula  will keep working on it and will compare it with a different model. So hopefully we will have more knowledge about the physics soon.  I'll merge this PR right now because it will make our work easier.

PR #34, opened 2021-09-07
Upgrade to new version of CLIMAParameters

trontrytel:
@charleskawczynski  - this passes all the tests and builds the docs on my local. 

Unfortunately something mysterious happened and I had to update the dependencies in Manifest file for docs. But we need to revert the dependencies in all toml files anyway, so I left it for now. Lets see how it goes in the CI

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,9 +12,8 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 
 [compat]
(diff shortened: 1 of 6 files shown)


charleskawczynski:
Awesome. I'll register a new CLIMAParameters version, update Thermo, and we can update this PR.

trontrytel:
> Awesome. I'll register a new CLIMAParameters version, update Thermo, and we can update this PR.

Great. Many thanks!

charleskawczynski (approved):
Looks great!

trontrytel:
So now we should register a new release here and then fix ClimateMachine.jl?

PR #36, opened 2021-09-08
bump version because of CLIMAParameters changes

trontrytel:


charleskawczynski (approved):
This looks like the correct version number bump, assuming there's been API changes (which I think we said was the case) for the upgrade CLIMAParameters PR.

trontrytel:
> This looks like the correct version number bump, assuming there's been API changes (which I think we said was the case) for the upgrade CLIMAParameters PR.

Yep - many of the function signatures have changed because we only need to pass in one parameter set

PR #41, opened 2021-09-14
Update Thermodynamics compat, bump version

charleskawczynski:
We should have bumped the thermo compat, right now, no packages can `resolve` to use the latest version because the version of thermodynamics that this package limits us to uses an older version of CLIMAParameters :(

I found this out when trying to update ClimateMachine.jl..


trontrytel (approved):

trontrytel:
Sorry about that!

charleskawczynski:
No problem :)

PR #49, opened 2021-11-04
use latent heat const with T

trontrytel:
Changing to use latent heats constant with temperature. Done to be consistent with saturation adjustment in Thermodynamics.jl and TC.jl

### DIFF
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -32,7 +32,8 @@ function G_func(param_set::APS, T::FT, ::TD.Liquid) where {FT <: Real}
     _R_v::FT = CP_planet.R_v(param_set)
     _D_vapor::FT = CP_micro.D_vapor(param_set)
 
-    L = TD.latent_heat_vapor(param_set, T)
+    L::FT = CP_planet.LH_v0(param_set)
+
     p_vs = TD.saturation_vapor_pressure(param_set, T, TD.Liquid())
 
     return FT(1) / (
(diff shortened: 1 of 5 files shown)


src/Common.jl
```diff
@@ -32,7 +32,8 @@ function G_func(param_set::APS, T::FT, ::TD.Liquid) where {FT <: Real}
     _R_v::FT = CP_planet.R_v(param_set)
     _D_vapor::FT = CP_micro.D_vapor(param_set)
 
-    L = TD.latent_heat_vapor(param_set, T)
+    L = CP_planet.LH_v0(param_set)
```
charleskawczynski, line 35:
```suggestion
    L::FT = CP_planet.LH_v0(param_set)
```
Previously the type could be promoted from `T`, now we need to type-cast for type stability

src/Common.jl
```diff
@@ -45,7 +46,8 @@ function G_func(param_set::APS, T::FT, ::TD.Ice) where {FT <: Real}
     _R_v::FT = CP_planet.R_v(param_set)
     _D_vapor::FT = CP_micro.D_vapor(param_set)
 
-    L = TD.latent_heat_sublim(param_set, T)
+    L = CP_planet.LH_s0(param_set)
```
charleskawczynski, line 49:
```suggestion
    L::FT = CP_planet.LH_s0(param_set)
```

src/Microphysics_1M.jl
```diff
@@ -759,7 +759,7 @@ function snow_melt(param_set::APS, q_sno::FT, ρ::FT, T::FT) where {FT <: Real}
         _D_vapor::FT = CP_micro.D_vapor(param_set)
         _K_therm::FT = CP_micro.K_therm(param_set)
 
-        L = TD.latent_heat_fusion(param_set, T)
+        L = CP_planet.LH_v0(param_set)
```
charleskawczynski, line 762:
```suggestion
        L::FT = CP_planet.LH_v0(param_set)
```

charleskawczynski (approved):

PR #51, opened 2021-12-10
add simple snow autoconversion

trontrytel:
This PR adds a simple snow autoconversion rate for simulation setups that don't allow for supersaturation to be present (like our current version of TC.jl with saturation adjustment).

It also bumps Julia version to 1.7


charleskawczynski (approved):
LGTM

trontrytel:
bors try

PR #61, opened 2022-04-20
Upgrade to latest thermodynamics

charleskawczynski:
I tried upgrading out of curiosity, and it looks like the changes are actually relatively minor on this end! The resolver is having some issues on Julia 1.5 (I saw the same issue in SurfaceFluxes). I think the easiest thing is to drop support for 1.5, since this is a breaking change anyway.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.3.4"
+version = "0.4.0"
 
 [deps]
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -13,5 +13,5 @@ Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 CLIMAParameters = "0.2, 0.3, 0.4"
 DocStringExtensions = "0.8"
 SpecialFunctions = "1, 2"
-Thermodynamics = "0.5, 0.6"
+Thermodynamics = "0.5, 0.6, 0.7"
```
trontrytel, line 16: Shouldn't it be that now we only support v0.7? This `q_sat  = TD.q_vap_saturation(param_set, ts_neq)` won't work with older versions?

charleskawczynski: Ah, yes, thank you!

trontrytel: Right. Lets be conservative and bump to 0.7.

I'm honestly surprised how minor the changes are.

charleskawczynski: Ah, we'll need to do a breaking change since we're dropping previous versions.

charleskawczynski: Hope that's okay!

trontrytel: 👍

PR #66, opened 2022-05-10
prep for more params boxes

trontrytel:
This PR moves NonEquilibrium cloud formation into a separate module and improves on some file names, following suggestions from @odunbar .

Additionally it introduces `InternalClimaParams.jl` as an interface to parameters and `CommonTypes.jl` as a common file for all type definitions.


trontrytel:
@charleskawczynski @odunbar @costachris 

Here is a draft of the changes for your boxing efforts. Let me know if it looks reasonable. 

I also wanted to add InternalClimaParams following what I see in Themodynamics.jl (https://github.com/CliMA/Thermodynamics.jl/pull/91/files) Would that be reasonable?

odunbar (approved):
Hi @trontrytel Thanks for the rearrangement! Small points
1. I am happy with the move of code into a common type. I can't comment on the move of the `AbstractAerosolDistribution` out however.
1. Rename `md` files to reflect rename of `.jl` files

You can merge once these are resolved.

More generally: Are types like `precipitation` and `moisture` always going to be created outside of Microphysics? (I was going to ask about renaming types to `Precipitation0M`, `Precipitation1M` and `MoistureEq` ,`MoistureNoneq`. But i see now that none of these types are created here, it would just be renaming modules.)

trontrytel:
> I also wanted to add InternalClimaParams following what I see in Themodynamics.jl

No. I'm adding internal parameters file as mentioned in my comment above.

odunbar (approved):
LGTM. 

Changing to the new `ICP.xyz` from the `CPP.` and `CP.` looks fine and smooths out the interface with the current setting. 

One line that creates some kind of pointer, causes a codecov issue. `Base.broadcastable(param_set::APS) = Ref(param_set)`
If there is a 1 line test - this could be added. (If not, don't worry about it - as this will be replaced in the new "parameter box" interface soon)

trontrytel:
> LGTM.

:+1:

> Changing to the new `ICP.xyz` from the `CPP.` and `CP.` looks fine and smooths out the interface with the current setting.

:+1:

> One line that creates some kind of pointer, causes a codecov issue. `Base.broadcastable(param_set::APS) = Ref(param_set)` If there is a 1 line test - this could be added. (If not, don't worry about it - as this will be replaced in the new "parameter box" interface soon)

What I learned from @charleskawczynski - This line is here to make broadcasting possible with parameters. So that we can write `@. x = alpha(param_set, y)` instead of `@. x = alpha(Ref(param_set), y)` when `x` and `y` are fields or arrays. 

I don't think I can make code coverage happy with it.

PR #71, opened 2022-06-17
Upgrade to latest Thermo, use new ClimaParameters

charleskawczynski:
This PR applies the clima parameters overhaul:

 - Defines and uses `CloudMicrophysicsParameters`
 - Upgrades to the latest version of Thermodynamics, which requires using its own thermodynamics parameters


trontrytel (approved):
Lets do this!

PR #76, opened 2022-08-03
smooth transitioning at thresholds of autoconversion rates

sajjadazimi:
# PULL REQUEST

## Purpose and Content
Auto conversion rates in CM1 and CM2 are modified by using logistic functions to make transition between different parameter regimes at thresholds smooth.

## Benefits and Risks
This will prevent abrupt changes of process rates in different parameter regimes. No risk: functions by default return the old process rates unless a boolean is changed, so the code modification will not cause contradictory results.

## PR Checklist
- This PR has a corresponding issue OR is linked to an SDI.
- I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- I linted my code on my local machine prior to submission OR N/A.
- Unit tests are included OR N/A.
- Code used in an integration test OR N/A.
- All tests ran successfully on my local machine OR N/A.
- All classes, modules, and function contain docstrings OR N/A.
- Documentation has been added/updated OR N/A.

### DIFF
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -51,4 +51,62 @@ function G_func(param_set::APS, T::FT, ::TD.Ice) where {FT <: Real}
     )
 end
 
+"""
+    logistic_function(x, x_0, k)
+
+ - `x` - independent variable
+ - `x_0` - threshold value for x
+ - `k` - growth rate of the curve, characterizing steepness of the transition
+
+Returns the value of the logistic function for smooth transitioning at thresholds. This is
+a normalized curve changing from 0 to 1 while x varies from 0 to Inf (for positive k). For
+x < 0 the value at x = 0 (zero) is returned. For x_0 = 0 H(x) is returned.
+"""
+function logistic_function(x::FT, x_0::FT, k::FT) where {FT <: Real}
+
+    @assert k > 0
+    @assert x_0 >= 0
+    x = max(0, x)
+
+    if abs(x) < eps(FT)
+        return FT(0)
+    elseif abs(x_0) < eps(FT)
+        return FT(1)
+    end
+
+    return 1 / (1 + exp(-k * (x / x_0 - x_0 / x)))
+end
+
+"""
+    logistic_function_integral(x, x_0, k)
+
+ - `x` - independent variable
+ - `x_0` - threshold value for x
+ - `k` - growth rate of the logistic function, characterizing steepness of the transition
+
+Returns the value of the indefinite integral of the logistic function, for smooth transitioning
+of piecewise linear profiles at thresholds. This curve smoothly transition from y = 0
+for 0 < x < x_0 to y = x - x_0 for x_0 < x.
+"""
+function logistic_function_integral(x::FT, x_0::FT, k::FT) where {FT <: Real}
+
+    @assert k > 0
+    @assert x_0 >= 0
+    x = max(0, x)
+
+    if abs(x) < eps(FT)
+        return FT(0)
+    elseif abs(x_0) < eps(FT)
+        return x
+    end
+
+    # translation of the curve in x and y to enforce zero at x = 0
+    _trnslt::FT = -log(1 - exp(-k)) / k
+
+    _kt::FT = k * (x / x_0 - 1 + _trnslt)
+    _result::FT =
+        (_kt > 40.0) ? x - x_0 : (log(1 + exp(_kt)) / k - _trnslt) * x_0
+    return _result
+end
+
 end
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -277,12 +277,24 @@ end
 Returns the q_rai tendency due to collisions between cloud droplets
 (autoconversion), parametrized following Kessler (1995).
 """
-function conv_q_liq_to_q_rai(param_set::APS, q_liq::FT) where {FT <: Real}
+function conv_q_liq_to_q_rai(
+    param_set::APS,
+    q_liq::FT;
+    smooth_transition::Bool = false,
+) where {FT <: Real}
 
     _τ_acnv_rai::FT = CMP.τ_acnv_rai(param_set)
     _q_liq_threshold::FT = CMP.q_liq_threshold(param_set)
 
-    return max(0, q_liq - _q_liq_threshold) / _τ_acnv_rai
+    _output::FT = FT(0)
+    if smooth_transition
+        _k::FT = CMP.k_thrshld_stpnss(param_set)
+        _output = CO.logistic_function_integral(q_liq, _q_liq_threshold, _k)
+    else
+        _output = max(0, q_liq - _q_liq_threshold)
+    end
+    return _output / _τ_acnv_rai
+
 end
 
 """
@@ -298,13 +310,22 @@ simulations where there is no supersaturation
 """
 function conv_q_ice_to_q_sno_no_supersat(
     param_set::APS,
-    q_ice::FT,
+    q_ice::FT;
+    smooth_transition::Bool = false,
 ) where {FT <: Real}
 
     _τ_acnv_sno::FT = CMP.τ_acnv_sno(param_set)
     _q_ice_threshold::FT = CMP.q_ice_threshold(param_set)
 
-    return max(0, q_ice - _q_ice_threshold) / _τ_acnv_sno
+    _output::FT = FT(0)
+    if smooth_transition
+        _k::FT = CMP.k_thrshld_stpnss(param_set)
+        _output = CO.logistic_function_integral(q_ice, _q_ice_threshold, _k)
+    else
+        _output = max(0, q_ice - _q_ice_threshold)
+    end
+    return _output / _τ_acnv_sno
+
 end
 
 """
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -7,6 +7,9 @@
 """
 module Microphysics2M
 
+import ..Common
+const CO = Common
+
 import ..Parameters
 const CMP = Parameters
 const APS = CMP.AbstractCloudMicrophysicsParameters
@@ -45,7 +48,7 @@ function conv_q_liq_to_q_rai_KK2000(
     N_d::FT = 1e8,
 ) where {FT <: Real}
 
-    q_liq = max(0.0, q_liq)
+    q_liq = max(0, q_liq)
 
     A::FT = CMP.A_acnv_KK2000(param_set)
     a::FT = CMP.a_acnv_KK2000(param_set)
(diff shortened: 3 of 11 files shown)


src/Microphysics1M.jl
```diff
@@ -46,6 +46,35 @@ export accretion_snow_rain
...
+
+Returns the value of the indefinite integral of the logistic function, for smooth transitioning
+of piecewise linear profiles at thresholds. This curve smoothly transition from y = 0 
+for 0 < x < x_0 to y = x - x_0 for x_0 < x.
+"""
+function logistic_function_integral(x::FT, x_0::FT, k::FT) where {FT <: Real}
+
+    @assert k > 0
+    @assert x_0 >= 0
+    x = max(0.0, x)
```
charleskawczynski, line 64:
```suggestion
    x = max(0, x)
```
`max` follows [promotion rules](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/), so using `0` here will make the code type-stable for `FT`.

Similar changes can be made below in this function.

src/Microphysics1M.jl
```diff
@@ -46,6 +46,35 @@ export accretion_snow_rain
...
+
+    # translation of the curve in x and y to enforce zero at x = 0
+    _trnslt = -log(1 - exp(-k)) / k
+
+    # When both x and x_0 are zero t is nan, and need to be replaced by Inf so that 
+    # the result for x = 0 is zero.
+    _t = (x_0 == 0.0) ? Inf : x / x_0 - 1 + _trnslt
```
charleskawczynski, line 71: `== 0.0` is usually a risky way to compare floating point numbers. It's probably safer to use something like `abs(x_0) < eps(FT)` or `abs(x_0) < sqrt(eps(FT))` depending on the situation.

src/Microphysics1M.jl
charleskawczynski: @trontrytel is pretty familiar with type promotions. And if CloudMicrophysics has type-stability tests, then this will be caught by them.

sajjadazimi: I fixed this everywhere in the file

src/Microphysics1M.jl
sajjadazimi: Actually I just wanted to capture NaN cases. I changed the condition and used isnan

src/Microphysics1M.jl
```diff
@@ -46,6 +46,35 @@ export accretion_snow_rain
 export evaporation_sublimation
 export snow_melt
 
+"""
+    logistic_function_integral(x, x_0, k)
+
```
trontrytel, line 51: Lets put that function into common (https://github.com/CliMA/CloudMicrophysics.jl/blob/main/src/Common.jl) This way we won't have to define it twice.

src/Microphysics1M.jl
```diff
@@ -277,12 +306,24 @@ end
...
+    param_set::APS,
+    q_liq::FT;
+    smooth_transition::Bool = false,
+) where {FT <: Real}
 
     _τ_acnv_rai::FT = CMP.τ_acnv_rai(param_set)
     _q_liq_threshold::FT = CMP.q_liq_threshold(param_set)
 
-    return max(0, q_liq - _q_liq_threshold) / _τ_acnv_rai
+    if smooth_transition
```
trontrytel, line 318: Can we change it to something like below? I think it better shows the logic of what is different between the two choices.
```
    if smooth_transition
        _k = CMP.k_thrshld_stpnss(param_set)
        _output = logistic_function_integral(q_liq, _q_liq_threshold, _k)
    else
        _output = max(0, q_liq - _q_liq_threshold)
    end
    return _output  / _τ_acnv_rai
```

src/Microphysics1M.jl
```diff
@@ -298,13 +339,22 @@ simulations where there is no supersaturation
...
+    q_ice::FT;
+    smooth_transition::Bool = false,
 ) where {FT <: Real}
 
     _τ_acnv_sno::FT = CMP.τ_acnv_sno(param_set)
     _q_ice_threshold::FT = CMP.q_ice_threshold(param_set)
 
-    return max(0, q_ice - _q_ice_threshold) / _τ_acnv_sno
+    if smooth_transition
+        k = CMP.k_thrshld_stpnss(param_set)
```
trontrytel, line 350: Same as the above comment

src/Microphysics1M.jl
```diff
@@ -277,12 +306,24 @@ end
...
+    q_liq::FT;
+    smooth_transition::Bool = false,
+) where {FT <: Real}
 
     _τ_acnv_rai::FT = CMP.τ_acnv_rai(param_set)
     _q_liq_threshold::FT = CMP.q_liq_threshold(param_set)
 
-    return max(0, q_liq - _q_liq_threshold) / _τ_acnv_rai
+    if smooth_transition
+        k = CMP.k_thrshld_stpnss(param_set)
```
trontrytel, line 319:
```suggestion
        _k::FT = CMP.k_thrshld_stpnss(param_set)
```

src/Microphysics1M.jl
```diff
@@ -277,12 +306,24 @@ end
...
+        _output =
```
trontrytel, line 320:
```suggestion
        _output::FT =
```

src/Microphysics1M.jl
```diff
@@ -277,12 +306,24 @@ end
...
+            logistic_function_integral(q_liq, _q_liq_threshold, k) / _τ_acnv_rai
+    else
+        _output = max(0, q_liq - _q_liq_threshold) / _τ_acnv_rai
```
trontrytel, line 323:
```suggestion
        _output::FT = max(0, q_liq - _q_liq_threshold) / _τ_acnv_rai
```

src/Microphysics2M.jl
```diff
@@ -71,18 +97,29 @@ function conv_q_liq_to_q_rai_B1994(
...
 
     C::FT = CMP.C_acnv_B1994(param_set)
     a::FT = CMP.a_acnv_B1994(param_set)
     b::FT = CMP.b_acnv_B1994(param_set)
     c::FT = CMP.c_acnv_B1994(param_set)
     N_0::FT = CMP.N_0_B1994(param_set)
-    d::FT =
-        N_d >= N_0 ? CMP.d_low_acnv_B1994(param_set) :
-        CMP.d_high_acnv_B1994(param_set)
+    d::FT = 0.0
```
trontrytel, line 110: Do we have to declare d here?

src/Microphysics2M.jl
```diff
@@ -118,7 +156,14 @@ function conv_q_liq_to_q_rai_TC1980(
 
     q_liq_threshold::FT = m0_liq_coeff * N_d / ρ * r_0^me_liq
 
-    return D * q_liq^a * N_d^b * heaviside(q_liq - q_liq_threshold)
+    _output = D * q_liq^a * N_d^b
```
trontrytel, line 159:
```
    if smooth_transition
        _k::FT = CMP.k_thrshld_stpnss(param_set)
        _output::FT = logistic_function(q_liq, q_liq_threshold, k)
    else
        _output::FT = heaviside(q_liq - q_liq_threshold)
    end
    return D * q_liq^a * N_d^b * _output

    
    
```

trontrytel: Maybe it improves the readability?

src/Microphysics2M.jl
```diff
@@ -155,7 +201,14 @@ function conv_q_liq_to_q_rai_LD2004(
...
-        return E * (q_liq * ρ)^3 / N_d * heaviside(R_6 - R_6C) / ρ
+        _output = E * (q_liq * ρ)^3 / N_d / ρ
+        if smooth_transition
+            k = CMP.k_thrshld_stpnss(param_set)
+            _output = _output * logistic_function(R_6, R_6C, k)
+        else
+            _output = _output * heaviside(R_6 - R_6C)
+        end
+        return _output
     end
```
trontrytel, line 212: Same as the above comment

trontrytel (changes requested):
Thank you for adding this and adding the tests.

Would it also be possible to add a documentation page explaining the option and showing how the smoothing works? (I can help explain how the docs are build)

src/Microphysics1M.jl
sajjadazimi: logistic_function and logistic_function_integral are two different function, but I agree that putting both of them in Common is cleaner

src/Microphysics1M.jl
```diff
@@ -46,6 +46,35 @@ export accretion_snow_rain
...
+    # translation of the curve in x and y to enforce zero at x = 0
+    _trnslt = -log(1 - exp(-k)) / k
+
+    # When both x and x_0 are zero t is nan, and need to be replaced by Inf so that 
+    # the result for x = 0 is zero.
+    _t = isnan(x / x_0) ? Inf : x / x_0 - 1 + _trnslt
+    _kt = k * _t
+    _result = (_kt > 50.0) ? x - x_0 : (log(1 + exp(_kt)) / k - _trnslt) * x_0
+
+    return (_result < eps(FT) ? 0.0 : _result)
```
charleskawczynski, line 75:
```suggestion
    return (_result < eps(FT) ? FT(0) : _result)
```

src/Microphysics1M.jl
```diff
@@ -46,6 +46,35 @@ export accretion_snow_rain
...
+    @assert k > 0
+    @assert x_0 >= 0
+    x = max(0, x)
+
+    # translation of the curve in x and y to enforce zero at x = 0
+    _trnslt = -log(1 - exp(-k)) / k
+
+    # When both x and x_0 are zero t is nan, and need to be replaced by Inf so that 
+    # the result for x = 0 is zero.
+    _t = isnan(x / x_0) ? Inf : x / x_0 - 1 + _trnslt
```
charleskawczynski, line 71:
```suggestion
    _t = isnan(x / x_0) ? FT(Inf) : x / x_0 - 1 + _trnslt
```

`Inf` is a `Float64` by default, so we'll need to wrap it in `FT` make it type-stable (in the case that we use it for calculations).

Won't `_t = Inf` result in `_result = Inf`? (is that okay?) If so, we could short-circuit to return `Inf` to avoid calling `log` and `exp` on `Inf` (`exp` is *exp*ensive).

src/Microphysics2M.jl
sajjadazimi: cpp experience effect :)

src/Microphysics1M.jl
sajjadazimi: declaring type of _output twice is apparently not allowed although the line will be hit once!

src/Microphysics2M.jl
sajjadazimi: Maybe I did it because you should declare type of a variable only once and I wanted to declare before if statement. I don't know which option is clearer!

src/Microphysics1M.jl
sajjadazimi: I agree

src/Microphysics2M.jl
trontrytel: I didn't know that. Lets just go with what was here before then

src/Microphysics1M.jl
sajjadazimi: _t = Inf results in _kt = Inf so _kt > 50 and x-x0 is returned. For large values of _kt instead of computing log and exp I directly return x-x0 to avoid extra computations and also to avoid overflow.

sajjadazimi:
@trontrytel @charleskawczynski 
Thanks both of you for your reviews. I applied your suggestions in my local repo. I'm starting the documentation page and push after it's done

src/Common.jl
```diff
@@ -51,4 +51,58 @@ function G_func(param_set::APS, T::FT, ::TD.Ice) where {FT <: Real}
...
+    @assert k > 0
+    @assert x_0 >= 0
+    x = max(0, x)
+
+    # When both x and x_0 are zero t is nan, and need to be replaced by -Inf so that 
+    # the result for x = 0 is zero.
+    _t = (x == 0.0 && x_0 == 0.0) ? -Inf : x / x_0 - x_0 / x
+    _kt = k * _t
+
+    return 1 / (1 + exp(-_kt))
```
charleskawczynski, line 76:
```suggestion
    if x == 0.0 && x_0 == 0.0
        return FT(0) # avoid computing `exp(-(-Inf))`
    else
        _t = x / x_0 - x_0 / x
        _kt = k * _t
        return 1 / (1 + exp(-_kt))
    end
    _kt = k * _t

    return 1 / (1 + exp(-_kt))
```
I think my suggestion earlier was to use branching (`if`-statements) like this to (sometimes) avoid calling `exp`. It's somewhat minor, and really only performance related.

The more important issue I have is with `x == 0.0 && x_0 == 0.0`. I'm just not sure that this is how we want to do the comparison. Let's consider what happens here:

```julia
julia> compare(num::FT) where {FT} = num == 0.0
compare (generic function with 1 method)

julia> compare(Float32(1e-7))
false

julia> compare(Float32(1e-15))
false

julia> compare(Float32(1e-60))
true

julia> compare(Float32(1e-30))
false

julia> compare(Float64(1e-30))
false

julia> compare(Float64(1e-300))
false

julia> compare(Float64(1e-600))
true

julia> compare(Float64(0.0))
true
```

sajjadazimi: Oh yes you're right. I had this in two functions, fixed one and missed this one. Again here I wanted to catch NaN cases, I guess it would be better to use isnan function instead. But considering performance as you suggested, we can avoid computing exp in some more cases. I'll push new commits.

trontrytel (approved):

PR #77, opened 2022-08-23
Define broadcastable for parameters

charleskawczynski:
This PR defines `Base.broadcastable` for cloud microphysics parameters. We need this so that broadcasting works with these parameters


trontrytel (approved):

trontrytel:
Do we want to bump the patch version and release?

Issue #74, opened 2022-07-26
Refactor the 2-moment microphysics module


trontrytel:
## Purpose

As pointed out [here](https://github.com/CliMA/TurbulenceConvection.jl/pull/1244) we should dispatch over different autoconversion and accretion rates inside the `CM2` module.

I'll have a draft of the changes here for review first.

PR #79, opened 2022-09-06
2-moment scheme refactor

trontrytel:
This PR introduces dispatch over different rain autoconversion and accretion schemes.

Closes https://github.com/CliMA/CloudMicrophysics.jl/issues/74

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.8.2"
+version = "0.9.0"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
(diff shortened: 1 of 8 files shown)


charleskawczynski (commented):
This definitely looks like a step in the right direction. I think this is a clear improvement over what's in main. I think it'd be fine to merge this and incrementally move one step further, or incorporate a followup step in this PR, and that would be to combine the singleton `Abstract2MPrecipType` subtypes with the parameter set inputs.

We do this sort of thing in SurfaceFluxes-- the type of input parameters determine which type of universal functions are used

charleskawczynski (approved):

trontrytel:
> This definitely looks like a step in the right direction. I think this is a clear improvement over what's in main. I think it'd be fine to merge this and incrementally move one step further, or incorporate a followup step in this PR, and that would be to combine the singleton `Abstract2MPrecipType` subtypes with the parameter set inputs.
> 
> We do this sort of thing in SurfaceFluxes-- the type of input parameters determine which type of universal functions are used

Sounds great. Lets do another refactor when we add 2-moment scheme with prognostic Nd. Should be more clear how to improve the dispatch.

PR #84, opened 2022-11-07
Aerosol nucleation

### LINKED ISSUE #93: Add CLOUD-parameterized Nucleation
Implement and verify correctness of nucleation pathways as parameterized by:
E. M. Dunne et al., Science, 10.1126/science.aaf2649 (2016).

Pathways:

- Binary sulfuric acid nucleation
- Ternary sulfuric acid, ammonia nucleation
- Organic
- Sulfuric acid + organic

For our aerosol model implementation, this provides a more full and accurate picture than the MAM3 specification.

nefrathenrici:
## Purpose 

This PR implements aerosol nucleation as parameterized by the CLOUD eperiments and from Vehkamaki et al 2002.

## Content

Parameterizations from Dunne et al., 2016, Kirkby et al., 2016, and Riccobno et al., 2014.
Reproductions of figures from Dunne et al., 2016.
A basic nucleation parameterization is included in src/Nucleation.jl. Self-coagulation of particles will be included once I figure out the discrepancy between the CAM5 paper and code.
A plotting script to reproduce figures from Vehkamaki et. al in docs/src/VehkamakiPlots.jl

## Review checklist

I have:

----

Closes #93


src/Nucleation.jl
```diff
@@ -0,0 +1,97 @@
...
+temp: Temperature (K)
+so4: Total gas phase concentration of sulfuric acid (cm^-3)
+
+# Examples
+```julia-repl
+julia> nucleation_timestep(0.55, 236, 1e7)
+162.96173344575243
+```
+"""
+
```
charleskawczynski, line 23:
```suggestion
```
I think doc strings cannot have newlines separating the functions / types

nefrathenrici: @charleskawczynski Sorry, I'm a little confused by the suggestion. How's this?
```
"""
    nucleation_timestep(rh, temp, so4)

 - `rh` - Relative humidity, expressed as a percentage between 0 and 1
 - `temp` - Temperature (K)
 - `so4` - Total gas phase concentration of sulfuric acid (cm^-3)

Returns the nucleation rate for a single timestep under the given conditions.
All newly formed particles will be placed into the Aitken mode.
The initial nucleation rate comes from Vehkamaki et al, 2002 doi:10.1029/2002JD002184
Adjustment factor comes from Kerminen and Kulmala, 2002 doi:10.1029/2002JD002184

# Examples
```julia-repl
julia> nucleation_timestep(0.55, 236, 1e7)
162.96173344575243
```
I copied the format from [remove_precipitation](https://github.com/CliMA/CloudMicrophysics.jl/blob/main/src/Microphysics0M.jl#L21).

One more thing - the PR doesn't include any tests, but I have a plotting script to make sure that it's correct. What should I do about codecov failing?

charleskawczynski: Hi @nefrathenrici, the general format is fine. The difference is that [here](https://github.com/CliMA/CloudMicrophysics.jl/blob/main/src/Microphysics0M.jl#L35-L36), there is no line between the function and the doc string. For example

##### Broken doc string:
```julia
"""
    Foo

"""

Foo() = 1
```

##### Working doc string:
```julia
"""
    Foo

"""
Foo() = 1
```

charleskawczynski: > One more thing - the PR doesn't include any tests, but I have a plotting script to make sure that it's correct. What should I do about codecov failing?

I think there's two things here:

 - First, our CI doesn't depend on CodeCov-- it's more of a warning signal if, for example, coverage suddenly drops 90%. The reason being that CodeCov is [not 100% accurate](https://github.com/JuliaCI/Coverage.jl#a-note-for-advanced-users) due to how it's tracked and compiler optimizations.
 - Second, the plotting script should probably be moved to the `test/` or `docs/`, so that we don't have to make CloudMicrophysics.jl depend on (and therefore ship with) Plots.jl, since this is a pretty heavy-weight package. The other advantage of this is that, if we add it to the docs, which I believe already has plots, is that we can run this script as a part of CloudMicrophysics.jl's CI, to ensure that it consistently works as changes are made

nefrathenrici: OK, should be fixed now. Thanks for clarifying everything!

src/Nucleation.jl
```diff
@@ -0,0 +1,94 @@
...
+"""
+    nucleation_timestep(rh, temp, so4)
+
+ - `rh` - Relative humidity, expressed as a percentage between 0 and 1
+ - `temp` - Temperature (K)
+ - `so4` - Total gas phase concentration of sulfuric acid (cm^-3)
+
+Returns the nucleation rate for a single timestep under the given conditions.
+All newly formed particles will be placed into the Aitken mode.
```
trontrytel, line 11: Would be good to say (probably in the docs) that we are talking here about the sulfuric acid only

src/Nucleation.jl
```diff
@@ -0,0 +1,94 @@
+module Nucleation
+
+"""
+    nucleation_timestep(rh, temp, so4)
+
+ - `rh` - Relative humidity, expressed as a percentage between 0 and 1
+ - `temp` - Temperature (K)
+ - `so4` - Total gas phase concentration of sulfuric acid (cm^-3)
```
trontrytel, line 8: We should be using SI units as inputs only

src/Nucleation.jl
```diff
@@ -0,0 +1,94 @@
...
+The initial nucleation rate comes from Vehkamaki et al, 2002 doi:10.1029/2002JD002184
+Adjustment factor comes from Kerminen and Kulmala, 2002 doi:10.1029/2002JD002184
+
+# Examples
+```julia-repl
+julia> nucleation_timestep(0.55, 236, 1e7)
+162.96173344575243
+```
+"""
+function nucleation_timestep(rh, temp, so4)
```
trontrytel, line 21: Would it be correct to name this `SO4_nucleation_rate`? Timestep may be confusing, for example when we will use some multi stage timesteppers

src/Nucleation.jl
```diff
@@ -0,0 +1,94 @@
...
+    i_T = (
+        -1.05611 + 0.00903378 * temp - 0.0000198417 * temp^2 +
+        2.46048e-8 * temp^3 - 0.0579087 / x_crit
+    )
+    j_T = (
+        -0.148712 + 0.00283508 * temp - 9.24619e-6 * temp^2 +
+        5.00427e-9 * temp^3 - 0.0127081 / x_crit
+    )
+
+    # Nucleation rate 1/(cm^3 s)
```
trontrytel, line 78: We should be using SI units as output too. So `1 / m3 /s`

src/Nucleation.jl
```diff
@@ -0,0 +1,94 @@
...
+
+# Examples
+```julia-repl
+julia> nucleation_timestep(0.55, 236, 1e7)
+162.96173344575243
+```
+"""
+function nucleation_timestep(rh, temp, so4)
+    # Vehkamaki et al 2002 for initial parameterization of nucleation rate J
+    ln_so4 = log(so4)
```
trontrytel, line 23: In the docs it would be great to underline that this is all an empirical fit to data. Do they mention any physics/chemistry resolving model in the original paper. That we could use as a reference? Or at least to understand how the process works and why do we need to approximate it here?

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Climate Modeling Alliance"]
 version = "0.9.2"
 
 [deps]
+CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
```
trontrytel, line 7: Could we keep the CliMAParameters dependency only in tests?

nefrathenrici: Yep, I forgot to remove it.

docs/src/VehkamakiPlots.jl
```diff
@@ -0,0 +1,174 @@
+using Plots
+
+include("../../src/Nucleation.jl")
+using .Nucleation
+
+# Reproduces figures from Vehkamaki et al, 2002 doi:10.1029/2002JD002184
+# In order verify that the implementation is correct
+# Nucleation rate shown here is without any adjustment due to self-coagulation
+
+function plot_vehk(
```
trontrytel, line 10: Do we want to keep this one in plots to compare with our nucleation rates? Or is this not needed?

nefrathenrici: I think it would be good to keep for now in order to compare. The other code doesn't rely on it, so it would be easy to remove later.

trontrytel:
Could we do the package updates in a separate PR?

nefrathenrici:
> Could we do the package updates in a separate PR? This would also then resolve the conflict in the Project.toml file. I think because of that conflict we don't see the documentation being deployed for this PR?

@trontrytel Should be fixed now. I'll be more careful about pushing package updates in the future

nefrathenrici:
Should be good to go except for the missing ion concentration profiles. See documentation [here](https://clima.github.io/CloudMicrophysics.jl/previews/PR84/AerosolNucleation/).

docs/src/lehtinen_apparent_nucleation_rate.jl
```diff
@@ -1,6 +1,8 @@
 import CLIMAParameters as CP
 using Plots
-include("../../src/Nucleation.jl")
+import CloudMicrophysics
+const pkg_dir = dirname(dirname(pathof(CloudMicrophysics)))
```
charleskawczynski, line 4:
```suggestion
const pkg_dir = pkgdir(CloudMicrophysics)
```

charleskawczynski (commented):
Left a comment

src/CommonTypes.jl
```diff
@@ -139,4 +139,9 @@ Base.broadcastable(x::IceType) = tuple(x)
 Base.broadcastable(x::RainType) = tuple(x)
 Base.broadcastable(x::SnowType) = tuple(x)
 
+Base.broadcastable(x::LiquidType) = tuple(x)
+Base.broadcastable(x::IceType) = tuple(x)
+Base.broadcastable(x::RainType) = tuple(x)
+Base.broadcastable(x::SnowType) = tuple(x)
```
trontrytel, line 145: Looks like the above lines are duplicated

trontrytel:
This is really great! Thank you for not giving up on this. 

Aside from the duplicated lines in `CommonTypes`, could we also add some test from aerosol nucleation module to the gpu tests and to performance test? I'm happy to do it s a separate PR though. Lets merge it in.

PR #103, opened 2023-02-28
Ice deposition on dust

trontrytel:
This PR adds the activated fraction parameterisation for Arizona Test Dust and Sahara Dust particles. This is an empirical parameterisation that represents a heterogeneous nucleation process of water vapour deposition on dust particles. This PR also adds a 0-dimensional adiabatic parcel example for visualising and testing different ice nucleation mechanisms.

### DIFF
--- a/docs/src/IceNucleationBox.md
+++ b/docs/src/IceNucleationBox.md
@@ -1,0 +1,177 @@
+# 0-dimensional ice nucleation model
+
+## Equations
+
+This is a 0-dimensional adiabatic parcel model for testing nucleation schemes.
+It is based on [KorolevMazin2003](@cite), as well as the cirrus box model
+  [Karcher2006](@cite), [Tully2022](@cite).
+The model solves for saturation in a 0-dimensional
+  adiabatic parcel raising with constant velocity.
+
+!!! note
+    For now, the equations are written assuming
+    that only the water vapor and cloud ice
+    are present in the system.
+
+We define ice saturation ratio ``S_i``
+```math
+\begin{equation}
+S_i = \frac{e}{e_{si}}
+\end{equation}
+```
+where:
+- ``e`` - is the partial pressure of water vapor,
+- ``e_{si}`` - is the partial pressure of water vapor at saturation over ice.
+
+The change in saturation can be described as
+```math
+\begin{equation}
+\frac{dS_i}{dt} = \frac{1}{e_{si}} \frac{de}{dt} - \frac{e}{e_{si}} \frac{de_{si}}{dt}
+\end{equation}
+```
+From ideal gas law the partial pressure of water vapor can be written as
+```math
+\begin{equation}
+e = q_v p \frac{R_v}{R_a}
+\end{equation}
+```
+where:
+- ``q_v`` is the water vapor specific humidity
+- ``p`` is the air pressure
+- ``R_v``, ``R_a`` are the gas constants for water vapor and air.
+
+The change in partial pressure can be written as
+```math
+\begin{equation}
+\frac{de}{dt} = \frac{dq_v}{dt} p \frac{R_v}{R_a} + q_v \frac{dp}{dt}\frac{R_v}{R_a}
+\end{equation}
+```
+
+From the Clausius–Clapeyron relation
+```math
+\begin{equation}
+\frac{de_{si}}{dt} = \frac{L_s e_{si}}{R_v T^2} \frac{dT}{dt}
+\end{equation}
+```
+where:
+- ``L_s`` is the latent heat of sublimation
+- ``T`` is the temperature.
+
+From the moist adiabatic assumption
+```math
+\begin{equation}
+\frac{dT}{dt} = \frac{R_a T}{c_p p} \frac{dp}{dt} + \frac{L_w}{c_p} \frac{dq_l}{dt} + \frac{L_s}{c_p} \frac{d q_i}{dt}
+\end{equation}
+```
+where:
+- ``L_w`` is the latent heat of evaporation.
+
+From hydrostatic balance and assuming constant vertical velocity:
+```math
+\begin{equation}
+\frac{dp}{dt} = -\frac{p g}{R_a T} w
+\end{equation}
+```
+where:
+- ``g`` is the gravitational acceleration
+- ``w`` is the constant vertical velocity.
+
+Assuming only water vapor and ice are present ``q_l = 0``, ``\frac{dq_v}{dt} = -\frac{dq_i}{dt}``,
+and rearranging the terms
+```math
+\begin{equation}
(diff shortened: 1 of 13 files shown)


trontrytel:
@amylu00 - I moved the code and updated the dependencies. 

The next steps would be to:
- [ ] Update the documentation to only include the processes we are adding now.
- [ ] Add documentation on how the cirrus box model works
- [ ] Finish coding the cirrus box model

parcel/parcel.jl
```diff
@@ -0,0 +1,216 @@
...
+import Thermodynamics as TD
+import CloudMicrophysics as CM
+import CLIMAParameters as CP
+
+const CMT = CM.CommonTypes
+const CMO = CM.Common
+const CMI = CM.HetIceNucleation
+const CMP = CM.Parameters
+
+FT = Float64
```
charleskawczynski, line 13: I would try avoiding this. I'll sprinkle some notes below on things you may need to change

trontrytel: You mean avoiding adding this example to the documentation? Or avoiding defining FT here?

charleskawczynski: Avoid defining FT like this

charleskawczynski: Ah, maybe just delete this definition and use `toml_dict = CP.create_toml_dict(Float64; dict_type = "alias")` below?

parcel/parcel.jl
```diff
@@ -0,0 +1,216 @@
...
+include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))
+toml_dict = CP.create_toml_dict(FT; dict_type = "alias")
+const param_set = cloud_microphysics_parameters(toml_dict)
+const thermo_params = CMP.thermodynamics_params(param_set)
+
+"""
+    ODE problem definition
+"""
+function cirrus_box(dY, Y, p, t)
+
```
charleskawczynski, line 25:
```suggestion
    (; param_set) = p
```
I'd recommend passing `param_set` through `p` to avoid the closure. We can also grab `FT`: `FT = eltype(Y)` (?) to avoid the closure over the global variable (which `FT` is a global non-constant, so this will _significantly_ slow things down).

test/heterogeneous_ice_nucleation_tests.jl
```diff
@@ -0,0 +1,53 @@
...
+import CLIMAParameters
+const CP = CLIMAParameters
+
+const TT = Test
+const TD = Thermodynamics
+const CMT = CloudMicrophysics.CommonTypes
+const CMI = CloudMicrophysics.HetIceNucleation
+
+include(joinpath(pkgdir(CloudMicrophysics), "test", "create_parameters.jl"))
+FT = Float64
```
charleskawczynski, line 14: I'd try to avoid defining a global non-constant type variable, since this opens the door for major performance hits

charleskawczynski (commented):
The docs look great! I'd probably just try to avoid closures over `FT`

parcel/parcel.jl
```diff
@@ -0,0 +1,216 @@
...
+    @assert isapprox(e, e_check; rtol = 0.01)
+
+    # Adiabatic parcel coefficients
+    a1 = L_subl * grav / cp_a / T^2 / R_v - grav / R_a / T
+    a2 = p_a / e_si * R_v / R_a
+    a3 = L_subl^2 / R_v / T^2 / cp_a
+
+    # Activating new crystals
+    AF = CMI.dust_activated_number_fraction(S_i, T, CMT.DesertDustType())
+    τ_relax = const_dt
```
sajjadazimi, line 66: Maybe pass `const_dt` to the function as a parameter through `p` instead of using it as a global variable.

parcel/parcel.jl
```diff
@@ -0,0 +1,216 @@
...
+    dN_act_dt = max(FT(0), AF * N_aerosol - N_act) / τ_relax
```
sajjadazimi, line 67: Updating `N_act` by using a first order approximation for the rate works only for some of the ODE solver schemes. Here our goal is that `N_act` reaches `AF * N_aerosol` after `dt`. It will perfectly work if the ODE solver is, for example, a first order Euler scheme. But if the scheme uses sub-stepping (like Range-Kutta schemes), the rate in the substeps will be smaller than what we want (because of diving by dt instead of dt_substep), and the result will be under predicted. In think updating Y instead of returning a rate should be more robust, unless you cautiously use a scheme that works.

sajjadazimi (approved):
It looks very good to me. I just left a couple of small comments plus a speculation about updating `N_act` to think about.

docs/src/IceNucleationBox.md
```diff
@@ -0,0 +1,177 @@
...
+
+We define ice saturation ratio ``S_i``
+```math
+\begin{equation}
+S_i = \frac{e}{e_{si}}
+\end{equation}
+```
+where:
+- ``e`` - is the partial pressure of water vapor,
+- ``e_{si}`` - is the partial pressure of water vapor at saturation over ice.
```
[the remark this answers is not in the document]
trontrytel: Some people do that. For example it is used in the Pruppacher and Klett book on Microphysics of Clouds and Precipitation. But I take your point that we haven't been using it like that. I'll switch to `p_v`.

parcel/parcel.jl
trontrytel: Agreed. However, the problem with that approach is that you can't overwrite your state vector `Y` when computing the tendencies for the ODE. We have run into this with EDMF code.  (I'll see if I can generate a bad example with this simple code.) But my understanding is that the ODE solver assumes that `Y` is unchanged and bad things happen if you try to change it behind it's back. Since I want to use the ODE solver instead of writing my own, I decided to stick with the simple Euler scheme. 

But it's probably something we would want to revisit once we start adding more competing processes here.

test/heterogeneous_ice_nucleation_tests.jl
trontrytel: This is inside our tests. Maybe it's ok to do `const FT = Float64`?

parcel/parcel.jl
trontrytel: I'll try to see if I can make an example later. I'll merge it in using Euler and const `dt` for now

docs/src/IceNucleationBox.md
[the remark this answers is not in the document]
trontrytel: Actually, I'm going to backtrack on that. In the docs switching `e_{si}` to `p_{vap}^{si}` is quite painful. I'll leave the current notation for now

test/heterogeneous_ice_nucleation_tests.jl
trontrytel: I'll try to do something better. But it will be a separate PR

PR #115, opened 2023-05-05
Add some performance monitoring to GHA

trontrytel:


trontrytel:
@charleskawczynski - could you take a look? Would that be meaningful to do for a couple of CM.jl functions?

I don't understand why the benchmark trial shows no memory and no allocations
```

BenchmarkTools.Trial: 10000 samples with 323 evaluations.
 Range (min … max):  258.514 ns … 437.774 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     258.827 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   259.503 ns ±   4.222 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
 
 Memory estimate: 0 bytes, allocs estimate: 0.
```

charleskawczynski (approved):
> Would this draft PR be meaningful to do for a couple of CM.jl functions?

Looks good! It can't hurt to have an idea of what the performance looks like. In thermodynamics, we're mostly interested in the performance of the thermo constructors since that's the main entry point and is the only place where we have an iterative solve.

You may want to call the function you're benchmarking once before calling `@benchmark` to make sure that the compile time is out of the way.

I think something simple here is most useful.

Also, from discussion on Julia slack, the _minimum_ time from the trials seems to be the "best" to use because (from what I understand) its most consistent.

PR #117, opened 2023-05-09
Run GHA tests for Float32 and Float64

trontrytel:
Closes https://github.com/CliMA/CloudMicrophysics.jl/issues/111

GitHub comparison makes this PR look much longer than it actually is. I just replaced

```
FT=Float64
function test_all_the_things()
```
with

```
function test_all_the_things(FT)
test_all_the_things(Float32)
test_all_the_things(Float64)
```

and fixed all the issues that I found along the way. (Like for example the fact that `2 * \pi` ends up being a `Float64`)

### DIFF
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -82,8 +82,8 @@ function raindrops_limited_vars(
     L_rai = ρ * q_rai
     xr_0 = L_rai / N_rai
     xr_hat = max(xr_min, min(xr_max, xr_0))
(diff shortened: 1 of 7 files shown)


trontrytel:
bors try

charleskawczynski:
> Like for example the fact that 2 * \pi ends up being a Float64)

True, but π does promote otherwise!

```julia
julia> Float32(1) * π
3.1415927f0
```

charleskawczynski (approved):

nefrathenrici (approved):

Issue #46, opened 2021-09-22
Improve unpack_parameters


trontrytel:
Currently we are inefficiently unpacking many parameters. See for example: 

https://github.com/CliMA/CloudMicrophysics.jl/blob/ce20220aa95a6699e64aae2a03758f9b7138d2b9/src/Microphysics_1M.jl#L163

@charleskawczynski  - could you provide some minimal example of how you thought this could be changed?

charleskawczynski:
Yep, I was thinking, instead of

```julia
function unpack_params(
    param_set::APS,
    snow::SnowType,
    ρ::FT,
    q_sno::FT,
) where {FT <: Real}
```

Where we unpack 

```julia
    _r0_sno::FT = CP_micro.r0_sno(param_set)

    _m0_sno::FT = CP_micro.m0_sno(param_set)
    _me_sno::FT = CP_micro.me_sno(param_set)
    _a0_sno::FT = CP_micro.a0_sno(param_set)
    _ae_sno::FT = CP_micro.ae_sno(param_set)
    _v0_sno::FT = CP_micro.v0_sno(param_set)
    _ve_sno::FT = CP_micro.ve_sno(param_set)

    _χm_sno::FT = CP_micro.χm_sno(param_set)
    _Δm_sno::FT = CP_micro.Δm_sno(param_set)
    _χa_sno::FT = CP_micro.χa_sno(param_set)
    _Δa_sno::FT = CP_micro.Δa_sno(param_set)
    _χv_sno::FT = CP_micro.χv_sno(param_set)
    _Δv_sno::FT = CP_micro.Δv_sno(param_set)
```
And dump this into a big NamedTuple. We could instead define

```julia
r0(param_set, ::SnowType) = CP_micro.r0_sno(param_set)

m0(param_set, ::SnowType) = CP_micro.m0_sno(param_set)
me(param_set, ::SnowType) = CP_micro.me_sno(param_set)
a0(param_set, ::SnowType) = CP_micro.a0_sno(param_set)
ae(param_set, ::SnowType) = CP_micro.ae_sno(param_set)
v0(param_set, ::SnowType) = CP_micro.v0_sno(param_set)
ve(param_set, ::SnowType) = CP_micro.ve_sno(param_set)

χm(param_set, ::SnowType) = CP_micro.χm_sno(param_set)
Δm(param_set, ::SnowType) = CP_micro.Δm_sno(param_set)
[...]

PR #122, opened 2023-05-15
Replace unpack_params with dispatch

trontrytel:
Closing https://github.com/CliMA/CloudMicrophysics.jl/issues/46

This is very very slightly faster than the previous version. So I think it's ok to merge in. Not sure if it's more or less elegant.


charleskawczynski (approved):
Nice that we have benchmarks to confirm this now!

PR #123, opened 2023-05-15
Add intro for new SURFs

trontrytel:
This is a quick cleanup in our README and the introduction page in our documentation. I'm hoping that both will come useful this summer during our SURF projects.


test/performance_tests.jl
claresinger, line 1: Seems like some test is timing out. But it doesn't seem like this code should have changed anything...

claresinger (approved):

PR #128, opened 2023-06-14
Add ABIFM to IceNucleation

amylu00:
## Purpose 

Adding immersion freezing for sulphuric acid droplets.

## To-do

- in parcel, remove growth to compare with Knopf&Alpert2013
- in parcel, plot J in cm^-2 s^-1 
- in parcel, plot ice production rate, P, in min^-1

## Content

- added ABIFM to IceNucleation.md
- added liquid water to supersaturation equation in parcel.jl and IceNucleationBox.md
- parameterized J from ABIFM in IceNucleation.jl
- added ABIFM parcel 

----

### DIFF
--- a/docs/bibliography.bib
+++ b/docs/bibliography.bib
@@ -191,6 +191,17 @@ url  ="http://dx.doi.org/10.1039/C3FD00035D"
   year = {2016}
 }
 
+@article{Koop2000,
+author = {Koop, Thomas and al, et},
+title = {Water activity as the determinant for homogeneous ice nucleation in aqueous solutions},
+journal = {Nature},
+volume = {406},
+pages = {611-614},
+year = {2000},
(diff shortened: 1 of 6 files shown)


trontrytel (commented):
Could we split this PR into 2 PRs:
1st - adding the ABIFM parameterization. With the goal being to reproduce the figure we talked about from the paper
2nd - changes to the parcel model?

Then for the 1st PR I would:
- change the output to be in base SI units
- add some unit tests, performance tests and GPU tests. (Please ask Nat for help with those in case of doubt)
- add the plot to the docs

For the 2nd PR it would be better to add different RHS (ABIFM, deposition, etc...) as options to the parcel model. This way we won't have to copy/paste the whole model. As the simplest one we could just pass additional arguments to the parcel model that would switch on/off different tendencies.

PR #139, opened 2023-06-30
Add options to parcel model

amylu00:
## Purpose 

Add feature that lets the user switch between different freezing modes/ turn different tendencies on or off. Parcel also able to run without freezing and defaults to no freezing with invalid freezing mode input.

## To-do

- take out conditional if statements for ABIFM and homo so user gets error if T or Delta_a_w out of valid range
- add evaporation/sublimation to dT/dt and mixing ratios (maybe different PR)
- add condensation for ABIFM and hom modes of freezing? (should ignore aerosols that have already nucleated into ice, they are considered in dqi_dt already)
- add option for different dust types?

## Content

- add liquid to equations
- option for deposition freezing
- option for no growth of ice crystals
- option for ABIFM
- option for homogeneous freezing (assuming aerosols are completely soluble)
- updated documentation for parcel

----


amylu00:
This is what ABIFM option looks like when I force `Delta_a_w` to be evaluated at T = 200K (so that T wouldn't be outside the valid T range) and comment out `@assert isapprox(e, e_check; rtol = 0.01)` because `e` and `e_check` were not in fact equal. On the bright side, it runs

cirrus_box

amylu00:
At this point: 
- deposition plots still work as before
- ABIFM and homogeneous freezing works but very little ice is forming (probably because we're multiplying J by either r^2 or r^3 and r is much smaller than 1)
- The ABIFM plots
ABIFM

- The homogeneous plots
homoge

parcel/parcel.jl
```diff
@@ -56,32 +62,68 @@ function parcel_model(dY, Y, p, t)
...
+
+        dN_act_dt = max(FT(0), AF * N_aerosol - N_act) / τ_relax
+    elseif freeze_mode == "ABIFM"
+
+        Delta_a_w =
+            T > FT(185) && T < FT(235) ? CMO.Delta_a_w(prs, x_sulph, T) : FT(0)
+        J_immer =
+            T > FT(185) && T < FT(235) ?
+            CMI_het.ABIFM_J(CMT.DesertDustType(), Delta_a_w) : FT(0)
+        P_ice = J_immer * 4 * π * r_nuc^2 * N_aerosol
```
amylu00, line 86: `N_aerosol` should be replaced with `(N_aerosol - N_act)`

parcel/parcel.jl
```diff
@@ -56,32 +62,68 @@ function parcel_model(dY, Y, p, t)
...
+
+        dN_act_dt = max(FT(0), P_ice)
+    elseif freeze_mode == "homogeneous"
+
+        Delta_a_w =
+            T > FT(185) && T < FT(235) ? CMO.Delta_a_w(prs, x_sulph, T) : FT(0)
+        J_homogeneous =
+            Delta_a_w > 0.26 && Delta_a_w < 0.34 ?
+            CMI_hom.homogeneous_J(Delta_a_w) : FT(0)
+        P_ice = J_homogeneous * 4 / 3 * π * r_nuc^3 * N_aerosol
```
amylu00, line 96: `N_aerosol` should be replaced with `(N_aerosol - N_act)`

PR #144, opened 2023-07-05
Implementing Chen 2022 Terminal Velocity Formulas into 2M scheme

apoorva-thanvantri:
#### made unit tests

----
- [X] I have read and checked the items on the review checklist.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -115,6 +115,7 @@ CommonTypes.B1994Type
 CommonTypes.TC1980Type
 CommonTypes.LD2004Type
 CommonTypes.SB2006Type
+CommonTypes.Chen2022Type
 CommonTypes.AbstractAerosolType
 CommonTypes.ArizonaTestDustType
 CommonTypes.DesertDustType
--- a/docs/src/TerminalVelocityComparisons.jl
+++ b/docs/src/TerminalVelocityComparisons.jl
@@ -1,0 +1,172 @@
+import Plots as PL
+
+FT = Float64
+
+import CloudMicrophysics as CM
+import CLIMAParameters as CP
+
+const CMT = CM.CommonTypes
+const CM1 = CM.Microphysics1M
+const CM2 = CM.Microphysics2M
+const CMP = CM.Parameters
+
+include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))
+toml_dict = CP.create_toml_dict(FT; dict_type = "alias")
+const param_set = cloud_microphysics_parameters(toml_dict)
+
+const rain = CMT.RainType()
+const SB2006 = CMT.SB2006Type()
+const Chen2022 = CMT.Chen2022Type()
+
+"""
+    rain_terminal_velocity_individual_C(ρ, scheme, D_r)
+
+ - `ρ`` - air density
+ - `scheme` - type for parameterization
+ - `D_r` - diameter of the raindrops
+
+ Returns the fall velocity of a raindrop using multiple gamma-function terms
+ to increase accuracy for `scheme == Chen2022Type`
+"""
+function rain_terminal_velocity_individual_Chen(
+    param_set,
+    ρ::FT,
+    scheme::CMT.Chen2022Type,
+    D_r::FT, #in m
+) where {FT <: Real}
+
+    # coefficients from Table B1 from Chen et. al. 2022
+    ρ0::FT = CMP.q_coeff_rain_Ch2022(param_set)
+    a1::FT = CMP.a1_coeff_rain_Ch2022(param_set)
+    a2::FT = CMP.a2_coeff_rain_Ch2022(param_set)
+    a3::FT = CMP.a3_coeff_rain_Ch2022(param_set)
+    a3_pow::FT = CMP.a3_pow_coeff_rain_Ch2022(param_set)
+    b1::FT = CMP.b1_coeff_rain_Ch2022(param_set)
+    b2::FT = CMP.b2_coeff_rain_Ch2022(param_set)
+    b3::FT = CMP.b3_coeff_rain_Ch2022(param_set)
+    b_ρ::FT = CMP.b_rho_coeff_rain_Ch2022(param_set)
+    c1::FT = CMP.c1_coeff_rain_Ch2022(param_set)
+    c2::FT = CMP.c2_coeff_rain_Ch2022(param_set)
+    c3::FT = CMP.c3_coeff_rain_Ch2022(param_set)
+
+    q = exp(ρ0 * ρ)
+    ai = (a1 * q, a2 * q, a3 * q * ρ^a3_pow)
+    bi = (b1 - b_ρ * ρ, b2 - b_ρ * ρ, b3 - b_ρ * ρ)
+    ci = (c1, c2, c3)
+
+    D_r = D_r * 1000 #D_r is in mm in the paper --> multiply D_r by 1000
+
+    v = 0
+    for i in 1:3
+        v += (ai[i] * (D_r)^(bi[i]) * exp(-D_r * ci[i]))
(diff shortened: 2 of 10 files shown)


docs/src/ReproduceWillis.jl
```diff
@@ -0,0 +1,120 @@
+import Plots
+
```
trontrytel, line 2: We can use those plots for testing things, but we probably wont need them in the final documentation.

docs/src/TerminalVelocityComparisons.jl
```diff
@@ -0,0 +1,229 @@
...
+    ],
+    linewidth = 3,
+    xlabel = "D [m]",
+    ylabel = "terminal velocity [m/s]",
+    label = "Rain-Chen2022",
+)
+PL.title!("Terminal Velocity vs. Diameter", titlefontsize = 13)
+PL.savefig("docs/src/terminal_velocity_individual_raindrop.svg") # hide
+
+#old work
```
trontrytel, line 124: Could we delete the commented code?

docs/src/TerminalVelocityComparisons.jl
```diff
@@ -0,0 +1,229 @@
...
+const param_set = cloud_microphysics_parameters(toml_dict)
+thermo_params = CMP.thermodynamics_params(param_set)
+
+const liquid = CMT.LiquidType()
+const ice = CMT.IceType()
+const rain = CMT.RainType()
+const snow = CMT.SnowType()
+const SB2006 = CMT.SB2006Type()
+const Chen2022 = CMT.Chen2022Type()
+
```
trontrytel, line 26: We use the velocities of individual drops only for visualisation/documentation. Therefore, could we define the Chen2022 individual velocity here instead of in the source code? Also, could we also add the 1-moment terminal velocity for comparison (I think it will be quite different than the other two)

docs/src/TerminalVelocityComparisons.jl
```diff
@@ -0,0 +1,229 @@
...
+const CMP = CM.Parameters
+const CH = CM.TerminalVelocityChen2022
+
+const APS = CMP.AbstractCloudMicrophysicsParameters
+include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))
+FT = Float64
+toml_dict = CP.create_toml_dict(FT; dict_type = "alias")
+const param_set = cloud_microphysics_parameters(toml_dict)
+thermo_params = CMP.thermodynamics_params(param_set)
+
```
trontrytel, line 19: You are not using all the types defined below. Could we delete the unused ones?

src/TerminalVelocityChen2022.jl
```diff
@@ -0,0 +1,111 @@
...
+    rain_terminal_velocity_individual(ρ, scheme, D_r)
+
+ - `ρ`` - air density
+ - `scheme` - type for parameterization
+ - `D_r` - diameter of the raindrops
+
+ Returns the fall velocity of a raindrop using multiple gamma-function terms
+ to increase accuracy for `scheme == Chen2022Type`
+"""
+function rain_terminal_velocity_individual(
```
trontrytel, line 25: I would suggest defining that one in the docs example. It's not going to be used in the actual source code. We just need the parameters to compute the bulk group velocity

src/TerminalVelocityChen2022.jl
```diff
@@ -0,0 +1,111 @@
...
+
+"""
+    rain_terminal_velocity(param_set, scheme, q_rai, ρ, N_rai)
+
+ - `param_set` - abstract set with Earth parameters
+ - `scheme` - type for parameterization
+ - `q_rai` - rain water specific humidity
+ - `ρ` - air density
+ - `N_rai` - raindrops number density
+
```
trontrytel, line 51: What are the units of input/output? Could you also add the link to the Chen2022 paper here?

test/Project.toml
```diff
@@ -3,7 +3,10 @@ BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 CUDAKernels = "72cfdca4-0801-4ab0-bf6a-d52aa10adc57"
+CloudMicrophysics = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
```
trontrytel, line 6: We don't want to add this dependency to the tests, because it will mess up with the CI setup. It's ok to leave it for now, so that things work for us locally. But we should remove it before merging

docs/src/TerminalVelocityChen2022.md
```diff
@@ -0,0 +1,52 @@
...
+```
+where $\Gamma$ is the standard gamma function integral and $\delta = \mu + k + 1$.
+
+
+```@example
+include("TerminalVelocityComparisons.jl")
+```
+![](terminal_velocity_bulk_comparisons.svg)
+![](terminal_velocity_individual_raindrop.svg)
+
```
trontrytel, line 52: Lets also remove `willis_figure2.svg` and `willis_figure3b.svg` from this PR

trontrytel:
I left some small comments. I'm wondering about something. If we move the individual velocity to documentation, then mabe the bulk velocity from Chen2022 could just be defined in the 2-moment microphysics module? I think that would make more sense, sonce it can only be used with it, as we are assuming the same size distribution

docs/src/TerminalVelocityComparisons.jl
apoorva-thanvantri: Yes, I'll do that!

src/TerminalVelocityChen2022.jl
apoorva-thanvantri: Sounds good, will do!

docs/src/API.md
```diff
@@ -126,3 +127,11 @@ CommonTypes.DesertDustType
 PrecipitationSusceptibility.precipitation_susceptibility_autoconversion
 PrecipitationSusceptibility.precipitation_susceptibility_accretion
 ```
+
+# Terminal Velocity (Chen 2022)
+
+```@docs
+TerminalVelocityChen2022
```
trontrytel, line 134: Now that we moved the velocities to `Microphysics2M` module, we should delete all of that.

test/chen_unit_tests.jl
```diff
@@ -0,0 +1,122 @@
...
+    TT.@testset "Chen 2022 rain terminal velocity individual" begin
+        #setup
+        ρ = FT(1.1)
+        q_rai = FT(1e-6)
+        N_rai = FT(1e4)
+        D_r = FT(2e-4)
+
+        ϕ = 1
+        q = exp(0.115231 * ρ)
+        ai = [0.044612 * q, -0.263166 * q, 4.7178 * q * ρ^(-0.47335)]
```
trontrytel, line 36: Thank you for adding the tests! Maybe instead of coding the same thing twice (in the source code and tests) we could instead take the values from the Chen2022 paper. For example for individual rain drop, you could take the values from their figure 3 (you can use tools like https://automeris.io/WebPlotDigitizer/ to read the data from the plot.

For the bulk velocity, I think it's ok to just take some values from our output (instead of computing them again in the test) and compare against them. This is not a test that checks for correctness, but only that things won't change in the future for example when the API changes. Also, if you do that you won't need special functions in your test anymore.

apoorva-thanvantri: Hi! I was thinking of deleting the tests for the individual raindrop since we're not adding that function into the code and only using it for plotting comparisons (let me know if you think that's ok). I will change the bulk velocity though!

trontrytel (approved):

PR #149, opened 2023-07-28
Implementing Chen 2022 Formulas (rain, ice, snow) into 1M scheme

apoorva-thanvantri:
## Purpose 
Implementing formulas from Chen 2022 into the 1M scheme

## Content
- formulas
- documentation
- unit tests

----

### DIFF
--- a/docs/src/TerminalVelocityComparisons.jl
+++ b/docs/src/TerminalVelocityComparisons.jl
@@ -1,37 +1,41 @@
 import Plots as PL
 
-FT = Float64
-
 import CloudMicrophysics as CM
 import CLIMAParameters as CP
+
+FT = Float64
 
 const CMT = CM.CommonTypes
 const CM1 = CM.Microphysics1M
 const CM2 = CM.Microphysics2M
 const CMP = CM.Parameters
 
+const rain = CMT.RainType()
+const liquid = CMT.LiquidType()
+const ice = CMT.IceType()
+const snow = CMT.SnowType()
+const SB2006 = CMT.SB2006Type()
+const Chen2022 = CMT.Chen2022Type()
+const SB2006Vel = CMT.SB2006VelType()
+const Blk1MVel = CMT.Blk1MVelType()
+const APS = CMP.AbstractCloudMicrophysicsParameters
+
 include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))
 toml_dict = CP.create_toml_dict(FT; dict_type = "alias")
 const param_set = cloud_microphysics_parameters(toml_dict)
 
-const rain = CMT.RainType()
-const SB2006 = CMT.SB2006Type()
-const Chen2022 = CMT.Chen2022Type()
-
 """
-    rain_terminal_velocity_individual_C(ρ, scheme, D_r)
+    rain_terminal_velocity_individual_Chen(param_set, ρ, D_r)
 
- - `ρ`` - air density
- - `scheme` - type for parameterization
+ - `param_set` - set with free parameters
+ - `ρ` - air density
  - `D_r` - diameter of the raindrops
 
- Returns the fall velocity of a raindrop using multiple gamma-function terms
- to increase accuracy for `scheme == Chen2022Type`
+Returns the fall velocity of a raindrop from Chen et al 2022
 """
 function rain_terminal_velocity_individual_Chen(
     param_set,
     ρ::FT,
-    scheme::CMT.Chen2022Type,
     D_r::FT, #in m
 ) where {FT <: Real}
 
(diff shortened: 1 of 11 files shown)


trontrytel:
@apoorva-thanvantri  - Thank you for adding all the parameterizations to the 1-moment scheme. It's really cool that you added the aspect ratio computation for snow!

I squashed and rebased everything and modified the velocity comparison plots and the documentation. I think that to avoid conflicts the easiest thing  would be to delete or rename your old branch. And then checkout to a new branch from main and pull the changes. Otherwise you will be resolving many conflicts.

A couple of things to do before we can merge this:
 - [ ] Would be great to add a plot of how the snow aspect ratio depends on the size
 - [ ] Would be great to also add a plot of individual terminal velocity for ice particles
 - [ ] We should move the free parameters to CLIMAParameters. Similar to what I did for the 2-moment rain PR.
 Let me know if you could take those on.
 
 I'll think some more on how to refactor the 1-moment terminal velocity functions such that we could keep the function signature the same and just dispatch based on the particle and scheme type.

trontrytel:
See also issue #154 To fully update the bulk 1-moment scheme to Chet et al 2022 velocities we would also have to re-do the integrals such as accretion, evaporation and such. Maybe we should meet for some integral party at some point. Those are not hard, but it will be annoying (especially for snow...)

docs/src/TerminalVelocityComparisons.jl
```diff
@@ -78,95 +89,134 @@ function rain_terminal_velocity_individual_SB(
...
+"""
+    snow_terminal_velocity_individual_Chen(prs, ρ, D_r)
+
+ - `prs` - set with free parameters
+ - `ρ` - air density
+ - `D_r` - particle diameter
+
+Returns the fall velocity of snow from Chen et al 2022
+"""
+function snow_terminal_velocity_individual_Chen(
```
amylu00, line 128: It would be nice to have the function throw a warning or error if D_r >= 0.625mm. If you do end up doing that, make sure to add a gpu/unit test for it

docs/src/TerminalVelocityComparisons.jl
```diff
@@ -1,37 +1,39 @@
...
- - `scheme` - type for parameterization
+ - `param_set` - set with free parameters
+ - `ρ` - air density
  - `D_r` - diameter of the raindrops
 
- Returns the fall velocity of a raindrop using multiple gamma-function terms
- to increase accuracy for `scheme == Chen2022Type`
+Returns the fall velocity of a raindrop from Chen et al 2022
 """
 function rain_terminal_velocity_individual_Chen(
```
amylu00, line 34: It would be nice to have the function throw a warning or error if D_r >= 0.1mm. If you do end up doing that, make sure to add a gpu/unit test for it

src/Microphysics1M.jl
```diff
@@ -205,6 +205,149 @@ function terminal_velocity(
...
+            ϕ = 1
+            κ = 0
+            q = exp(0.115231 * ρ)
+            ai = [0.044612 * q, -0.263166 * q, 4.7178 * q * ρ^(-0.47335)]
+            bi = [
+                2.2955 - 0.038465 * ρ,
+                2.2955 - 0.038465 * ρ,
+                1.1451 - 0.038465 * ρ,
+            ]
+            ci = [0, 0.184325, 0.184325]
```
amylu00, line 274:
```suggestion
            ci = [FT(0), 0.184325, 0.184325]
```

amylu00 (commented):
It looks beautiful, you did great Apoorva! A lot of my suggestions are just changing some whole numbers to `FT(number)` just to keep things consistent, not sure if there's any value in that computationally, but just in case? My other suggestions are related to the valid range of `D_r`. I think these changes would make it easier on future users/developers because it's so easy to be unaware that you're out of the valid range.

docs/src/TerminalVelocityComparisons.jl
trontrytel: This is just for plotting. We don't use it in the actual code. - In the code we use the mass weighted mean. And the integral is done from zero to infinity, so I guess we are using the function outside of it's range. I think it's fine tho

src/Microphysics1M.jl
```diff
@@ -205,6 +205,149 @@ function terminal_velocity(
...
+            0.00419647 * sqrt(ρ_i)
+        Es = -0.156593 - 0.0189334 * (log(ρ_i))^2 + 0.1377817 * sqrt(ρ_i)
+        Fs = -exp(-3.35641 - 0.0156199 * (log(ρ_i))^2 + 0.765337 * log(ρ_i))
+        Gs =
+            (-0.0309715 + 1.55054 / (log(ρ_i)) - 0.518349 * log(ρ_i) / ρ_i)^(-1)
+
+        k = 3 #for mass weighted terminal velocity
+
+        if precip isa CT.RainType && spherical
+            ϕ = 1
```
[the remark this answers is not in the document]
trontrytel: Definitely! Otherwise it's not type stable. I'll try to refactor the code a little to have the same function signature for all 1-moment terminal velocities. I can change all of those

apoorva-thanvantri:
Hi! Sounds good - I'm running a few more simulations on KiD and then I will
get to these issues (I can add the plots and move the parameters). I'm also
free for an integral party whenever :)

trontrytel:
Added some work in progress on refactoring things. Need to fix 3 tests to make it work

trontrytel:
Closing in favour of https://github.com/CliMA/CloudMicrophysics.jl/pull/163

PR #150, opened 2023-07-29
P3 - look-up table

rorlija1:
## Purpose 

Rather than use NonlinearSolve.jl at each time step, for the P3 scheme, we have decided to create a look-up table that can be used to quickly generate threshold values. This PR aims to add functions which create and navigate this table.

## To-do

- add lookup table functions

## Content

----


charleskawczynski:
> Rather than use NonlinearSolve.jl at each time step, for the P3 scheme, we have decided to create a look-up table that can be used to quickly generate threshold values.

just a heads up, this may not be GPU-compatible

rorlija1:
> > Rather than use NonlinearSolve.jl at each time step, for the P3 scheme, we have decided to create a look-up table that can be used to quickly generate threshold values.
> 
> just a heads up, this may not be GPU-compatible



> > Rather than use NonlinearSolve.jl at each time step, for the P3 scheme, we have decided to create a look-up table that can be used to quickly generate threshold values.
> 
> just a heads up, this may not be GPU-compatible

Ah, thanks for letting me know! I'll bring this up with Anna, Sajjad, and others

trontrytel:
Closing it for now. Lets see how slow/fast things are when we run it on the GPU

PR #172, opened 2023-08-21
Add ML aerosol activation emulators

mikhailmints:
## Purpose 

This PR adds support for ML aerosol activation emulators to be used instead of the ARG scheme.

## To-do

Fix the tests and documentation for these changes, possibly improve the abstractions used in the models to avoid hardcoding hyperparameters.

## Content

- Add support for using data-driven emulators of aerosol activation, which can be loaded through MLJ.
- Add utilities for data preprocessing and model construction.
- Add the free parameters of the ARG scheme to the parameter set.
- Add PySDM data to ARG plots in documentation.

----


trontrytel:
Done in https://github.com/CliMA/CloudMicrophysics.jl/pull/349 and https://github.com/CliMA/CloudMicrophysics.jl/pull/318

PR #193, opened 2023-08-28
Add modal nucleation parameters to struct, update functions

### LINKED ISSUE #192: Integrate modal nucleation parameters into struct
The modal nucleation parameters are stored in climaparameters but are just extracted from the `toml_dict` during tests.
We need to update the parameter struct, functions, and tests to add the parameters to the full cloudmicrophysics struct.
This can serve as an example for breaking up the parameters struct as well.

nefrathenrici:
## Content
- Updates CloudMicrophysicsParameters to include a ModalNucleationParameters struct, update machinery to generate functions for parameters
- Updates nucleation functions to match the struct. Changes a few function names where values are explicitly given (mostly for testing)
- Update unit and GPU perf tests
- Update docs
- Update from `const CMP = CM.Parameters` to using `import CloudMicrophysics.Parameters as CMP` for all submodules

closes #192

### DIFF
--- a/docs/src/nucleation_plotting/CLOUD_Nucleation_Plots.jl
+++ b/docs/src/nucleation_plotting/CLOUD_Nucleation_Plots.jl
@@ -1,39 +1,13 @@
 using Plots
-using CLIMAParameters
 
-include("../../src/Nucleation.jl")
-using .Nucleation
-
-# Testing for CLOUD-experiment based nucleation rates.
+import CLIMAParameters as CP
+import CloudMicrophysics as CM
+import CloudMicrophysics.Nucleation as Nucleation
+include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))
 
 FT = Float64
-toml_dict = CLIMAParameters.create_toml_dict(FT)
-param_names = [
-    "u_b_n",
-    "v_b_n",
-    "w_b_n",
-    "u_b_i",
-    "v_b_i",
-    "w_b_i",
-    "u_t_n",
-    "v_t_n",
-    "w_t_n",
-    "u_t_i",
-    "v_t_i",
-    "w_t_i",
-    "p_t_n",
-    "p_A_n",
-    "a_n",
-    "p_t_i",
-    "p_A_i",
-    "a_i",
-    "p_b_n",
-    "p_b_i",
-    "p_t_n",
-    "p_t_i",
-]
-params = CLIMAParameters.get_parameter_values!(toml_dict, param_names)
-params = (; params...)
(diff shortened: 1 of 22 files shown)


docs/src/nucleation_plotting/CLOUD_Nucleation_Plots.jl
```diff
@@ -1,39 +1,13 @@
...
-using CLIMAParameters
 
-include("../../src/Nucleation.jl")
-using .Nucleation
-
-# Testing for CLOUD-experiment based nucleation rates.
+import CLIMAParameters as CP
+import CloudMicrophysics.Parameters as CMP
+import CloudMicrophysics.Nucleation as Nucleation
+include("../../test/create_parameters.jl")
```
trontrytel, line 6: Is this way better? Or it doesn't really matter?

include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))

test/runtests.jl
```diff
@@ -1,4 +1,4 @@
-include("performance_tests.jl")
+# include("performance_tests.jl")
```
trontrytel, line 1: We should include that back, rught?

test/create_parameters.jl
```diff
@@ -21,50 +22,17 @@ function cloud_microphysics_parameters(toml_dict)
...
+    modal_nucleation_params = CMP.ModalNucleationParameters{FT}(; pairs...)
+    MNP = typeof(modal_nucleation_params)
+
+    aliases = string.(fieldnames(CMP.CloudMicrophysicsParameters))
+    pairs = CP.get_parameter_values!(toml_dict, aliases, "CloudMicrophysics")
+
+    return CMP.CloudMicrophysicsParameters{FT, TP, MNP}(;
         pairs...,
         thermo_params,
+        modal_nucleation_params,
```
trontrytel, line 36: This still puts all parameters into one struct right?

I was wondering if it would be better for GPU performance to have smaller separate structs? Or it doesn't make a difference?

test/runtests.jl
nefrathenrici: Yes! An embarassing typo

test/create_parameters.jl
nefrathenrici: I'm not sure, I can ask Sriharsha.

trontrytel: So far we have this: https://github.com/CliMA/CloudMicrophysics.jl/pull/194

So at least for now we pass in the whole struct, but we interact with it differently inside the functions. Such that we don't have to unpack the parameters by hand inside the functions.

trontrytel (approved):
Thank you!

PR #203, opened 2023-08-31
Water Activity Documentation/Plots

amylu00:
## Purpose 

Add detailed documentation for water activity.

## To-do

- add plot of various vapor pressure parameterizations
- add plot of various T vs a_w's parameterizations
- explain the different parameterizations
- explain what happens when droplets are equilibrium with environment
- explain why sometimes dew point is needed
- add note about needing a parameterization for solution droplets at T > 235K

## Content

----


trontrytel (approved):
LGTM! Thank you for persisting and figuring it all out.

PR #206, opened 2023-09-02
Cleanup in the P3 scheme solver

trontrytel:
This PR:
- updates the P3 scheme nonlinear solver to new way of handling parameters,
- cleans up the code by defining some helper functions to be used in the nonlinear problem definition and tests
- simplifies the way things are plotted for the docs

### DIFF
--- a/src/parameters/cmp_p3.jl
+++ b/src/parameters/cmp_p3.jl
@@ -1,0 +1,31 @@
+"""
+    CloudMicrophysicsParametersP3{FT}
+
+Parameters for P3 bulk microphysics scheme
+
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+struct CloudMicrophysicsParametersP3{FT} <: AbstractCloudMicrophysicsParameters
+    "Coefficient in mass(size) relation [g/μm^β_va]"
+    α_va::FT
+    "Coefficient in mass(size) relation [-]"
+    β_va::FT
+    "Cloud ice density [kg/m3]"
+    ρ_i::FT
+    "Cloud liquid water density [kg/m3]"
+    ρ_l::FT
+    "Coefficient in area(size) for ice side plane, column, bullet, and planar polycrystal aggregates from Mitchell 1996 [μm^(2-σ)]"
+    γ::FT
+    "Coefficient in area(size) for ice side plane, column, bullet, and planar polycrystal aggregates from Mitchell 1996 [-]"
+    σ::FT
+end
+
+CloudMicrophysicsParametersP3(param_set) = CloudMicrophysicsParametersP3(
+    param_set.α_va_BF1995,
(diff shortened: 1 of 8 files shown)


src/parameters/cmp_p3.jl
```diff
@@ -0,0 +1,31 @@
...
+    CloudMicrophysicsParametersP3{FT}
+
+Parameters for P3 bulk microphysics scheme
+
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+struct CloudMicrophysicsParametersP3{FT} <: AbstractCloudMicrophysicsParameters
+    "Coefficient in mass(size) relation [g/μm^β_va]"
+    α_va::FT
```
sriharshakandala, line 11: Can we use `αᵥₐ::FT` , etc. for variable names!

src/parameters/cmp_p3.jl
```diff
@@ -0,0 +1,31 @@
...
+    ρ_i::FT
+    "Cloud liquid water density [kg/m3]"
+    ρ_l::FT
+    "Coefficient in area(size) for ice side plane, column, bullet, and planar polycrystal aggregates from Mitchell 1996 [μm^(2-σ)]"
+    γ::FT
+    "Coefficient in area(size) for ice side plane, column, bullet, and planar polycrystal aggregates from Mitchell 1996 [-]"
+    σ::FT
+end
+
+CloudMicrophysicsParametersP3(param_set) = CloudMicrophysicsParametersP3(
```
sriharshakandala, line 24: This looks good. But it is best to contruct this directly from the `ClimaParameters` dictionary/ toml file! This way, we can also keep deleting fields from the bigger `CloudMicrophysicsParameters` struct as we keep moving fields to smaller structs!

test/p3_tests.jl
```diff
@@ -3,166 +3,88 @@ import CloudMicrophysics as CM
...
     prs = cloud_microphysics_parameters(toml_dict)
-
-    # bulk density of ice:
-    ρ_i::FT = CMP.ρ_cloud_ice(prs)
-    # mass power law coefficient and exponent:
-    α_va::FT = P3.α_va(prs)
-    β_va = CMP.β_va_BF1995(prs)
-    # threshold particle dimension
-    D_th::FT = P3.D_th(prs, FT)
+    p3 = CMP.CloudMicrophysicsParametersP3(prs)
```
sriharshakandala, line 14: Please see the comment above!

sriharshakandala:
bors try

src/parameters/cmp_p3.jl
trontrytel: Lets leave it for now, so that we don't do any breaking changes for ClimaAtmos and other upstream packages. We will delete the big struct once we are done with reorganizing things and make another breaking release then.

PR #208, opened 2023-09-05
Add aqua tests, small fixes for tests

nefrathenrici:
Content:
- Add aqua and tests. There currently four function signatures with unbound args due to a known issue with aqua.
- Remove exports for variables that no longer exist: `ABIFM_Delta_a_w` and `Chen_snow_ice_coeffs`


charleskawczynski (approved):

nefrathenrici:
bors try

nefrathenrici:
bors try

nefrathenrici:
bors try-

nefrathenrici:
bors try

test/aqua.jl
```diff
@@ -38,7 +39,7 @@ using Aqua
 
     @testset "Aqua tests (additional)" begin
         Aqua.test_undefined_exports(CloudMicrophysics)
-        Aqua.test_stale_deps(CloudMicrophysics)
+        Aqua.test_stale_deps(CloudMicrophysics; ignore = [:CUDA])
```
charleskawczynski: Why do we need to ignore CUDA? It doesn't look like it's a dependency

test/aqua.jl
```diff
@@ -0,0 +1,50 @@
...
+    end
+    ignore = [:CUDA]
+    @testset "Aqua tests (additional)" begin
+        Aqua.test_undefined_exports(CloudMicrophysics)
+        Aqua.test_stale_deps(CloudMicrophysics; ignore)
+        Aqua.test_deps_compat(CloudMicrophysics; ignore)
+        Aqua.test_project_extras(CloudMicrophysics)
+        Aqua.test_project_toml_formatting(CloudMicrophysics)
+        Aqua.test_piracy(CloudMicrophysics)
+    end
```
charleskawczynski, line 47:
```suggestion
    @testset "Aqua tests (additional)" begin
        Aqua.test_undefined_exports(CloudMicrophysics)
        Aqua.test_stale_deps(CloudMicrophysics)
        Aqua.test_deps_compat(CloudMicrophysics)
        Aqua.test_project_extras(CloudMicrophysics)
        Aqua.test_project_toml_formatting(CloudMicrophysics)
        Aqua.test_piracy(CloudMicrophysics)
    end
```

charleskawczynski: CUDA is not a direct dependency, so we should not need to ignore it.

test/aqua.jl
nefrathenrici: It gets added for GPU tests during the [pipeline](https://github.com/CliMA/CloudMicrophysics.jl/blob/main/.buildkite/pipeline.yml#L25)- I don't think it's necessary to add the package into CloudMicrophysics but I didn't want to mess with the pipeline.

nefrathenrici:
bors try-

nefrathenrici:
bors try

PR #226, opened 2023-09-10
Use RootSolvers in P3

trontrytel:
Would this be GPU compatible? And can I make allocations go all the way to zero?

It is definitely faster! Showing benchmarks for Float32:

```
NonlinearSolver version (current main):

BenchmarkTools.Trial: 4726 samples with 1 evaluation.
 Range (min … max):  1.004 ms …   2.797 ms  ┊ GC (min … max): 0.00% … 62.61%
 Time  (median):     1.022 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.057 ms ± 204.686 μs  ┊ GC (mean ± σ):  2.91% ±  8.38%
 Memory estimate: 1.02 MiB, allocs estimate: 11294.

RootSolvers version (PR):

BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.321 μs … 105.704 μs  ┊ GC (min … max): 0.00% … 97.07%
 Time  (median):     1.350 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.426 μs ±   1.765 μs  ┊ GC (mean ± σ):  2.11% ±  1.69%
 Memory estimate: 1.45 KiB, allocs estimate: 71.
```

Also, are [Integrals.jl](https://docs.sciml.ai/Integrals/stable/) ok to use on GPU?


trontrytel:
bors try

charleskawczynski (approved):
Nice improvement!

I think we should be able to get everything stack allocated (and therefore show 0 allocations)

> Also, are [Integrals.jl](https://docs.sciml.ai/Integrals/stable/) ok to use on GPU?

My guess is probably not-- maybe `solve` is, but setting up the problem is probably not. What is it needed for?

docs/src/P3SchemePlots.jl
```diff
@@ -44,53 +44,53 @@ A_ns(p3::PSP3, D::FT) = p3.γ * D^p3.σ
...
+ - th - P3Scheme nonlinear solve output tuple (D_cr, D_gr, ρ_g, ρ_d)
 
 Returns mass(D) regime, used to create figures for the docs page.
 """
 function mass(
     p3::PSP3,
     D::FT,
     F_r::FT,
-    D_th::Vector{FT} = [0.0, 0.0, 0.0, 0.0],
+    th = (; D_cr = 0.0, D_gr = 0.0, ρ_g = 0.0, ρ_d = 0.0),
```
charleskawczynski, line 60:
```suggestion
    th = (; D_cr = FT(0), D_gr = FT(0), ρ_g = FT(0), ρ_d = FT(0)),
```

docs/src/P3SchemePlots.jl
```diff
@@ -44,53 +44,53 @@ A_ns(p3::PSP3, D::FT) = p3.γ * D^p3.σ
...
+ - th - P3Scheme nonlinear solve output tuple (D_cr, D_gr, ρ_g, ρ_d)
 
 Returns area(D), used to create figures for the documentation.
 """
 function area(
     p3::PSP3,
     D::FT,
     F_r::FT,
-    D_th::Vector{FT} = [0.0, 0.0, 0.0, 0.0],
+    th = (; D_cr = 0.0, D_gr = 0.0, ρ_g = 0.0, ρ_d = 0.0),
```
charleskawczynski, line 93:
```suggestion
    th = (; D_cr = FT(0), D_gr = FT(0), ρ_g = FT(0), ρ_d = FT(0)),
```

trontrytel:
> Nice improvement!
> 
> I think we should be able to get everything stack allocated (and therefore show 0 allocations)
> 
> > Also, are [Integrals.jl](https://docs.sciml.ai/Integrals/stable/) ok to use on GPU?
> 
> My guess is probably not-- maybe `solve` is, but setting up the problem is probably not. What is it needed for?

The next thing I have to add to P3 scheme is computing some integrals over the size distribution of particles. The integrals don't have analytical solutions because they are not from zero to infinity but instead over some limited range of particle sizes. I saw that the `Integrals.jl` package can handle such things

Maybe I can prototype with it anyway, and then we will see how to replace it

PR #229, opened 2023-09-14
Remove nucleation nested parameter struct

nefrathenrici:
Content
- Remove nucleation nested parameter struct, place in top-level
- Extend the new parameter scheme (0 moment, P3) to the nucleation parameters

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.13.3"
+version = "0.14.0"
 
 [deps]
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
--- a/src/parameters/modal_nucleation.jl
+++ b/src/parameters/modal_nucleation.jl
@@ -1,0 +1,84 @@
+"""
+    H2S04NucleationParameters{FT}
+
+Parameters for pure sulfuric acid nucleation
+
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+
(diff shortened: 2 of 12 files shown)


trontrytel:
This should be a breaking change? Should we bump the version to 0.14?

nefrathenrici:
> This should be a breaking change? Should we bump the version to 0.14?

@trontrytel This will be a breaking change, I will up the version when the PR is ready

src/parameters/modal_nucleation.jl
```diff
@@ -0,0 +1,85 @@
...
+    param_set.a_4,
+    param_set.a_5,
+    param_set.Y_MTO3,
+    param_set.Y_MTOH,
+    param_set.k_MTO3,
+    param_set.k_MTOH,
+    param_set.exp_MTO3,
+    param_set.exp_MTOH,
+    param_set.k_H2SO4org,
+)
```
charleskawczynski, line 78:
```suggestion
function ModalNucleationParameters(param_set)
    fns = fieldnames(ModalNucleationParameters)
    vals = getproperty.(Ref(param_set), fns)
    nt = (;zip(fns, vals)...)
    ModalNucleationParameters(;nt...)
end
```
I think this should work (safely) too, if all of the names are consistently stored in `param_set`

trontrytel:
The docs error should be fixed now on main

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.13.3"
+version = "0.14.0"
```
trontrytel, line 4: Could I ask you to update ClimaAtmos and Kinematic1D models to the new release?

trontrytel (approved):
Thank you!

PR #235, opened 2023-09-22
parcel with gamma size distribution of droplets

trontrytel:
Adds a liquid only parcel example with a gamma (`(r) = A r exp(-\lambada r)`) size distribution of droplets.

@amylu00 - lmk if you agree that this makes sense


trontrytel:
bors try

amylu00 (approved):

PR #236, opened 2023-09-23
switch to using RootSolvers in docs

trontrytel:
We are now using RootSolvers in the P3 scheme. It can do the same thing as Roots, and we will have one less dependency to worry about. 

Closing https://github.com/CliMA/CloudMicrophysics.jl/issues/228

Also updating the documenter. I think the main thing is `makedocs(..., bib) -> makedocs(..., plugins=[bib])`


charleskawczynski (approved):
RootSolvers part looks fine to me. Note that the default tolerance is very small, so you may want to adjust it.

amylu00 (approved):

PR #237, opened 2023-09-23
Clean up liquid parcel initial condition

### LINKED ISSUE #230: fix air density in liquid_only
Currently, moist air density is calculated using `q_l = 0`. This air density is then used in the calculation of `q_l` to be inputted into the parcel model as an initial condition. It is a non-zero quantity since the initial condition has liquid droplets. Need to find a way to make air density consistent with the fact that `q_l ` is non-zero.

trontrytel:
I think this fixes the parcel initial condition for the liquid only example. Doesn't really change the results, but it consistently treats the vapor and liquid

Closes #230

### DIFF
--- a/parcel/Liquid_only.jl
+++ b/parcel/Liquid_only.jl
@@ -29,21 +29,23 @@ r_0 = FT(8e-6)
 p = FT(800 * 1e2)
 T = FT(273.15 + 7.0)
 x_sulph = FT(0)
-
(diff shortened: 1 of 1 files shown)


trontrytel:
bors try

sriharshakandala:
bors try

sriharshakandala:
bors try

PR #241, opened 2023-09-25
RH based water activity

amylu00:
## Content

- break  `Delta_a_w` function up into 2 solution droplet water activity parameterizations (one based off Luo, the other on RH) and a water activity over ice parameterization

----


amylu00:
bors try

trontrytel (approved):

PR #257, opened 2023-10-12
adding box model to test immersion freeze

amylu00:
## To-do

- add to documentation (where?)

## Content

- added a box model to test immersion freezing

----


amylu00:
Final plots: 

Screenshot 2023-10-12 at 2 40 16 PM

trontrytel:
Looks great! I would maybe suggest to:
- create a separate folder for the box model (so that we don't mix our zero-dimesnional models)
- add a (very short) page to the docs to describe the model and include the plots?

trontrytel:
Also, do we know why our frozen fraction is more steep than AK2016? Does it depend in model timestep? Or is something not specified about the initial conditions?

It's strange that we don't have to model the amount of water here. But I guess if you prescribe the temperature, then it doesn't matter how much latent heat gets released

amylu00:
> Also, do we know why our frozen fraction is more steep than AK2016? Does it depend in model timestep? Or is something not specified about the initial conditions?
> 
> It's strange that we don't have to model the amount of water here. But I guess if you prescribe the temperature, then it doesn't matter how much latent heat gets released

It's just that at the start (warmer T) our J is negligibly small (55 cm^-2 s^-1) and at the cooler end, most of the droplets have already froze. The main difference is where we calculate `a_{ice}` so I guess our `a_{ice}` is more sensitive to temperature changes than the paper's.

trontrytel:
All problems solved. We can reproduce Fig 4 as done in the AK2016 paper, i.e. computing the apparent and actual immersion J based on prescribed frozen fraction.

Alpert_Knopf_2016_backward

We can also reproduce the frozen fraction line running the forward model and sampling from the distribution of surface areas. If we assume constant A we first underestimate and then overestimate the freezing rate, leading to much sharper transition.

Alpert_Knopf_2016_forward

trontrytel (approved):

PR #264, opened 2023-10-20
Split microphysics tests into 0, 1, 2 -moment and nonequilibrium

trontrytel:


test/microphysics0M_tests.jl
```diff
@@ -0,0 +1,61 @@
...
+import CloudMicrophysics.Microphysics0M as CM0
+
+include(joinpath(pkgdir(CM), "test", "create_parameters.jl"))
+
+@info "0-moment microphysics tests"
+
+function test_microphysics0M(FT)
+
+    toml_dict = CP.create_toml_dict(FT; dict_type = "alias")
+    prs = cloud_microphysics_parameters(toml_dict)
```
sriharshakandala, line 17: Do we still need this?

test/microphysics0M_tests.jl
```diff
@@ -0,0 +1,61 @@
...
+        TT.@test CM0.remove_precipitation(params_0m, q, q_vap_sat) ≈ FT(0)
+
+        # rain based on qc threshold
+        for lf in frac
+            q_liq = qc * lf
+            q_ice = (1 - lf) * qc
+
+            q = TD.PhasePartition(q_tot, q_liq, q_ice)
+
+            TT.@test CM0.remove_precipitation(prs, q) ≈
```
sriharshakandala, line 40: Can we use 'params_0m` here?

test/microphysics0M_tests.jl
trontrytel: Will be fixed in the next PR

test/microphysics0M_tests.jl
trontrytel: Will be fixed in the next PR

PR #268, opened 2023-10-25
more cleanup

trontrytel:
This PR:
- removes `kwdef` from modal aerosol nucleation parameters
- removes inline from aerosol model
- replaces inner constructor with outer constructor for aerosol model


sriharshakandala:
bors try

sriharshakandala (approved):

PR #271, opened 2023-10-31
No kwargs in parameters constructors

trontrytel:


nefrathenrici (commented):
Is this resolving the same ambiguous dispatch issue as before?

trontrytel:
> Is this resolving the same ambiguous dispatch issue as before?

I just noticed that this is the only constructor that has `;` instead of `,`. So I figured they should all be the same. Dont have a preference which way

nefrathenrici (approved):
Sounds good to me!

trontrytel:
I think the CI is broken right now. I saw the same package error in ClimaAtmos CI. So probably something with the cluster

PR #274, opened 2023-11-07
fix docs builds

trontrytel:


trontrytel:
@Sbozzolo - should we mention keeping the main branch for docs builds in the wiki?

Sbozzolo:
I didn't think it was needed because it is run as part of the PR, but maybe it is.

trontrytel:
I just noticed that the documentation page was not updated. I'll wait for it to build and see now

Sbozzolo (approved):

PR #279, opened 2023-12-03
Adding a Mode_INP aerosol distribution

amylu00:
## Purpose 

- Add an INP aerosol distribution to eventually have aerosol activation in parcel

## To-do

- unit test
- performance test
- gpu test
- add aerosol activation function(?)

## Content

----


trontrytel:
@amylu00  Could you remind me what we were trying to do here? Do we want to use it to store the aerosol size distribution parameters for the parcel simulations? I guess we would need a different type for delta, gamma, etc size distributions?

PR #282, opened 2024-01-04
lat heat of fus in parcel dT/dt and dSl/dt

amylu00:
Accounts for liquid-ice transitions in temperature and liquid supersaturation tendencies in the parcel. Documentation also updated to reflect the change. Updated aqua tests.

### DIFF
--- a/docs/src/IceNucleationParcel0D.md
+++ b/docs/src/IceNucleationParcel0D.md
@@ -55,13 +55,15 @@ where:
 From the moist adiabatic assumption
 ```math
 \begin{equation}
-\frac{dT}{dt} = \frac{R_a T}{c_p p} \frac{dp}{dt} + \frac{L_v}{c_p} \frac{dq_l}{dt} + \frac{L_s}{c_p} \frac{d q_i}{dt}
+\frac{dT}{dt} = \frac{R_a T}{c_p p} \frac{dp}{dt} + \frac{L_v}{c_p} \frac{dq_{l,vap}}{dt} + \frac{L_s}{c_p} \frac{d q_{i,subl}}{dt} + \frac{L_f}{c_p} \frac{d q_{i,fus}}{dt}
 \end{equation}
 ```
 where:
-- ``q_l`` is the cloud liquid water specific humidity,
-- ``q_i`` is the cloud ice specific humidity,
-- ``L_s`` is the latent heat of sublimation.
+- ``q_{l,vap}`` is the cloud liquid water specific humidity from vaporization/condensation,
+- ``q_{i,subl}`` is the cloud ice specific humidity from sublimation/deposition,
+- ``q_{i,fus}`` is the cloud ice specific humidity from melting/freezing,
+- ``L_s`` is the latent heat of sublimation,
+- ``L_f`` is the latent heat of fusion.
 
 From hydrostatic balance and assuming constant vertical velocity:
 ```math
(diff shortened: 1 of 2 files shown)


parcel/parcel.jl
```diff
@@ -53,15 +53,17 @@ function parcel_model(dY, Y, p, t)
     R_a = TD.gas_constant_air(tps, q)
     cp_a = TD.cp_m(tps, q)
     L_subl = TD.latent_heat_sublim(tps, T)
+    L_fus = TD.latent_heat_sublim(tps, T)
```
trontrytel, line 56: Should be `latent_heat_fusion`

test/aqua.jl
```diff
@@ -41,8 +41,7 @@ using Aqua
         Aqua.test_stale_deps(CloudMicrophysics)
         Aqua.test_deps_compat(CloudMicrophysics)
         Aqua.test_project_extras(CloudMicrophysics)
-        Aqua.test_project_toml_formatting(CloudMicrophysics)
-        Aqua.test_piracy(CloudMicrophysics)
+        Aqua.test_piracies(CloudMicrophysics)
```
trontrytel, line 44: Thank you!

docs/src/IceNucleationParcel0D.md
```diff
@@ -88,7 +90,7 @@ a_1 = \frac{L_v g}{c_p T^2 R_v} - \frac{g}{R_a T}
 ```
 ```math
 \begin{equation}
-a_2 = \frac{1}{q_v}
+a_2 = \frac{R_v P}{e_{sl}R_a}
```
trontrytel, line 93: `\frac{R_v p}{e_{sl}R_a} = \frac{S_l}{q_v}`? I think the `S/q_v` is cleaner

parcel/parcel.jl
```diff
@@ -135,12 +137,17 @@ function parcel_model(dY, Y, p, t)
         end
     end
 
+    dq_liq_dt_lv = dql_dt_cond        # from liq-vap transitions
```
trontrytel, line 140: Maybe we should be more verbose now about the direction of the phase change (since we now have more and more phase changes and this is becoming more error prone). 

Would it make sense to change `dq_liq_dt_lv` into `dq_liq_dt_vapor_to_liquid` or something similar? So that we dont make error somewhere in the signs. Same for the other ones

docs/src/IceNucleationParcel0D.md
amylu00: Oh yeah I'll fix that. I think it's `1/(S*q_v)` though?

docs/src/IceNucleationParcel0D.md
```diff
@@ -77,7 +79,7 @@ Accounting for conservation of water, i.e. ``\frac{dq_v}{dt} + \frac{dq_l}{dt} +
 and rearranging the terms
 ```math
 \begin{equation}
-\frac{dS_l}{dt} = a_1 w S_l - \left(a_2 + a_3 \right) S_l \frac{dq_l}{dt} - \left(a_2 + a_4\right) S_l \frac{dq_i}{dt}
+\frac{dS_l}{dt} = a_1 w S_l - \left(a_2 + a_3 S_l \right) \frac{dq_{l,vap}}{dt} - \left(a_2 + a_4  S_l\right) \frac{dq_{i,subl}}{dt} - a_5 S_l \frac{dq_{i,fus}}{dt}
```
trontrytel, line 82: Shouldn't it be ``(a_2 + a_5) S_l \frac{dq_{i,fus}}{dt}`` ?

Also, since ``a_2 = S_l / q_v`` we can actually move ``S_l`` outside of everything. I.e., lets define ``a_2 = `1 / q_v``.

trontrytel: If you agree then we should also change the parcel equations...

PR #287, opened 2024-01-18
Update to Julia 1.10

nefrathenrici:
- Update buildkite and GH actions to 1.10
- Set KernelAbstractions compat
- Bump patch version

The required CI actions should be set to 1.10, I don't have permissions to do this.


trontrytel (approved):
Thank you

PR #288, opened 2024-01-18
Update to ClimaParameters 0.8

nefrathenrici:
## Content
- Removes `src/parameters/Thermodynamics.jl` and replaces the CM thermo params constructor with the new constructor from the Thermodynamics package. 
- Propagates above change to all tests and docs
- Fixes broken `create_toml_dict` calls
- DOES NOT change CloudMicrophysics parameter constructors, since the internals of the toml dict are unchanged except for alias removal. I will add a later PR that uses ClimaParameters properly, this PR only sets compatibility with the breaking changes.


trontrytel (approved):
Thank you!

PR #293, opened 2024-01-26
Move P3 mass functions to src, add gamma functions to docs

trontrytel:
This PR adds the documentation about finding shape parameters to docs.

It also moves the P3 mass functions to source.


trontrytel:
Lets add tests for the two functions we are adding int source code

PR #298, opened 2024-01-29
F_r = 0 exception fixed, mass tests updated

anastasia-popova:
## Content

----

### DIFF
--- a/src/P3Scheme.jl
+++ b/src/P3Scheme.jl
@@ -120,32 +120,39 @@ Returns a named tuple containing:
 """
 function thresholds(p3::PSP3{FT}, ρ_r::FT, F_r::FT) where {FT}
 
-    @assert ρ_r > FT(0)   # rime density must be positive ...
-    @assert ρ_r <= p3.ρ_l # ... and as a bulk ice density can't exceed the density of water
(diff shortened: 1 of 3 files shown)


src/P3Scheme.jl
```diff
@@ -120,32 +120,39 @@ Returns a named tuple containing:
...
     @assert F_r < FT(1)   # ... and there must always be some unrimed part
 
-    P3_problem(ρ_d) =
-        ρ_d - ρ_d_helper(
-            p3,
-            D_cr_helper(p3, F_r, ρ_g_helper(ρ_r, F_r, ρ_d)),
-            D_gr_helper(p3, ρ_g_helper(ρ_r, F_r, ρ_d)),
+    if F_r == 0
+        
+        return()
```
trontrytel, line 128: We might have to return zeros to make the GPU test run

src/P3Scheme.jl
```diff
@@ -199,4 +206,20 @@ function p3_mass(
     end
 end
 
+#= """
```
trontrytel, line 209: Lets remove it if its not needed

trontrytel (approved):
Thank you

PR #299, opened 2024-01-29
Add shape parameters solver for P3

trontrytel:
This PR adds shape parameters solver to the P3 scheme. 

Right now it only works for Float64. We need to switch to different units inside the solver. We should also test the range of input parameters for which the solver converges.


anastasia-popova (approved):
Looks good.

PR #300, opened 2024-01-30
adding activity based dep to parcel

amylu00:
## To-do

- compare water based dep with mohler's method

## Content

- sanity check for correct parameterization of J in docs
- add water based deposition nucleation to parcel

----


docs/src/IceNucleationParcel0D.md
```diff
@@ -198,14 +198,28 @@ where:
...
+  from [Mohler2006](@cite). `"MohlerRate_Deposition"` will trigger a
+  nucleation rate approach from [Mohler2006](@cite). For both approaches,
+  there is no nucleation if saturation over ice exceeds 1.35 as conditions
+  above this value will result in nucleation in a different mode.
+  `"ActivityBasedDeposition"` will trigger a water activity based approach
+  from [Alpert2022](@cite). In this approach, ice production rate ``P_{ice, depo}``
+  is calculated from 
+```math
+\begin{equation}
+  P_{ice} = \left[ \frac{dN_i}{dt} \right]_{dep0} = J_{depo}A_{aero}N_{aero}
```
trontrytel, line 211:
```suggestion
  P_{ice} = \left[ \frac{dN_i}{dt} \right]_{depo} = J_{depo}\;A_{aero}\;N_{aero}
```

parcel/parcel.jl
```diff
@@ -99,6 +99,15 @@ function parcel_model(dY, Y, p, t)
         end
         dqi_dt_new_depo = dN_act_dt_depo * 4 / 3 * FT(π) * r_nuc^3 * ρ_ice / ρ_air
     end
+    if "ActivityBasedDeposition" in ice_nucleation_modes
+        Δa_w = CMO.a_w_eT(tps, e, T) - CMO.a_w_ice(tps, T)
+        J_deposition = CMI_het.deposition_J(aerosol, Δa_w)
+        if "Monodisperse" in droplet_size_distribution
```
trontrytel, line 105: Maybe we should add some error or at least warning that we don't support any other distribution type here?

parcel/parcel.jl
```diff
@@ -99,6 +99,15 @@ function parcel_model(dY, Y, p, t)
         end
         dqi_dt_new_depo = dN_act_dt_depo * 4 / 3 * FT(π) * r_nuc^3 * ρ_ice / ρ_air
     end
+    if "ActivityBasedDeposition" in ice_nucleation_modes
```
trontrytel, line 102: I think we are assuming here that someone will either run with `MohlerAF` `MohlerRate` or `ActivityBased` because each time we are doing `dqi_dt_new_depo = ...` and `dN_act_dt_depo = ...` instead of `+=`. I think thats fine, but maybe we could add some error warning to highlight that.

I think because the aerosol types don't overlap between `Mohler` and `Activity` based, one could try to run both at the same time

parcel/parcel.jl
amylu00: Yeah that's a good idea. Do you think it's ok to have a printed statement or should we have a hard `@assert`? I can also open an issue to add more size distributions to some of the freezing modes

docs/src/plots/activity_based_deposition.jl
```diff
@@ -0,0 +1,87 @@
+import Plots as PL
+
+import Thermodynamics as TD
+import CloudMicrophysics as CM
+import CloudMicrophysics.Common as CO
+import CloudMicrophysics.HetIceNucleation as IN
+import CloudMicrophysics.Parameters as CMP
+
+FT = Float64
```
trontrytel, line 9: Could we switch to Float32

docs/src/plots/activity_based_deposition.jl
```diff
@@ -0,0 +1,87 @@
...
+const tps = TD.Parameters.ThermodynamicsParameters(FT)
+const feldspar = CMP.Feldspar(FT)
+const ferrihydrite = CMP.Ferrihydrite(FT)
+const kaolinite = CMP.Kaolinite(FT)
+
+# Initializing
+Δa_w = range(0, stop = 0.32, length = 50)    # difference in solution and ice water activity
```
trontrytel, line 16:
```suggestion
Δa_w = range(FT(0), stop = 0.32, length = 50)    # difference in solution and ice water activity
```
You might need something like this to make `Δa_w` `Float32`

parcel/parcel.jl
trontrytel: I think its ok to just add warnings, or whatever is the easiest. We will refactor parcel soon anyway.

trontrytel (approved):
Looks great! Thank you for adding the plots with comparisons. 

Apologies for merging the parameters update before this. Lmk if there is any friction with it from the package manager

PR #303, opened 2024-02-01
unitless N_hat added, Float32 tests passing

anastasia-popova:
## Content

----


trontrytel (approved):
LGTM - could you also fix the documentation? I think we are incorrectly stating the units of `N_0` 

I can squash/rebase and merge once that is fixed. Thank you!

PR #304, opened 2024-02-02
replacing Mohler AF with nucleation rate

amylu00:
## Purpose 

- replacing the AF parameterization from Mohler with the nucleation rate parameterization from the same paper. This requires one less empirical parameter and would be more numerically stable in the parcel model.

## To-do

- clean up? delete AF function or replace with a function that calculates the rate. also delete unnecessary parameters (see CLIMAParameters.jl)
- if replacing AF function with new function for rate, revise unit/gpu/performance tests, api, etc.
- fix Tully; the jump between GCM time steps should cause a jump in ice formation

### DIFF
--- a/docs/src/IceNucleation.md
+++ b/docs/src/IceNucleation.md
@@ -22,28 +22,41 @@ The parameterization for deposition on dust particles is an implementation of
     freezing modes.
 
 ## Activated fraction for deposition freezing on dust
-The parameterization models the activated fraction
-  as an empirical function of ice saturation ratio,
-  see eq. (3) in [Mohler2006](@cite).
+There are 2 parameterizations from [Mohler2006](@cite) available: one
+  which calculates the activated fraction and one which calculates nucleation
+  rate. 
+The activated fraction parameterization follows eq. (3) in the paper.
 ```math
 \begin{equation}
 f_i(S_i) = exp[a(S_i - S_0)] - 1
 \end{equation}
 ```
-where:
+where 
   - ``f_i`` is the activated fraction
-      (the ratio of aerosol particles acting as ice nuclei to the total number of aerosol particles),
+     (the ratio of aerosol particles acting as ice nuclei to the total number of aerosol particles),
+  - ``a`` is a scaling parameter dependent on aerosol properties and temperature,
+  - ``S_i`` is the ice saturation ratio,
+  - ``S_0`` is an empirically derived threshold ice saturation ratio.
+The other parameterization models the nucleation rate of ice
+  as an empirical function of ice saturation ratio,
+  see eq. (5) in [Mohler2006](@cite).
+```math
+\begin{equation}
+\frac{dn_{ice}}{dt} = N_{aer} a \frac{dS_i}{dt}
+\end{equation}
+```
+where:
+  - ``N_{aer}`` is the number of available aerosol/ice nuclei,
+  - ``a`` is a scaling parameter dependent on aerosol properties and temperature,
   - ``S_i`` is the ice saturation ratio
-      (the ratio of water vapor partial pressure and the water vapor partial pressure at saturation over ice),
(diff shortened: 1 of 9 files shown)


docs/src/IceNucleation.md
```diff
@@ -22,28 +22,41 @@ The parameterization for deposition on dust particles is an implementation of
...
+    (the ratio of water vapor partial pressure and the water vapor partial pressure at saturation over ice).
 
-Limited experimental values for the free parameters ``S_0`` and ``a`` can be found in [Mohler2006](@cite).
-Both parameters are dependent on aerosol properties and temperature.
+Limited experimental values for the free parameters ``a`` and ``S_0`` can be found in [Mohler2006](@cite). These
+  free parameters are strongly dependent on aerosol properties and temperature.
 
 !!! note
 
-    For a ``f_i`` values above 0.08 or ``S_i`` between 1.35 and 1.5,
```
trontrytel, line 46: Is the `f_i` `0.08` limit no longer needed?

docs/src/IceNucleationParcel0D.md
```diff
@@ -241,13 +238,20 @@ Here we show various example simulation results from the adiabatic parcel
...
+  for 30 minutes simulation time, (shown by three different colors on the plot).
 Between each run the water vapor specific humidity is changed,
   while keeping all other state variables the same as at the last time step
   of the previous run.
 The prescribed vertical velocity is equal to 3.5 cm/s.
 Supersaturation is plotted for both liquid (solid lines) and ice (dashed lines).
+The pale blue line uses the ``"MohlerRate_Deposition"``approach. 
+  We only run it for the first GCM timestep because the rate approach requires
+  the change in ice saturation over time. With the discontinuous jump in saturation,
+  the parameterization is unablet to determine a proper nucleation rate. When we force
```
trontrytel, line 251:
```suggestion
  the parameterization is unable to determine a proper nucleation rate. When we force
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -196,15 +196,12 @@ It follows that
...
-```math
-\begin{equation}
-  N_{act} = N_{aer} f_i
-\end{equation}
-```
-where:
-- ``N_{aer}`` is the number of available dust aerosol particles.
+## Deposition Nucleation on dust particles
+There are multiple ways of running deposition nucleation in the parcel.
+  ``"MohlerAF_Deposition"`` will trigger an activated fraction approach from [Mohler2006](@cite).
```
trontrytel, line 201: MohlerAF_Deposition here and throughout the added text doesn't render properly. The last line has an example that works

src/IceNucleation.jl
```diff
@@ -41,6 +42,37 @@ function dust_activated_number_fraction(
...
 end
 
+"""
+    MohlerDepositionRate(dust, ip, S_i, dSi_dt)
+
+ - `dust` - a struct with dust parameters
+ - `ip` - a struct with ice nucleation parameters
+ - `Si` - ice saturation
+ - `dSi_dt` - change in ice saturation over time
+
```
trontrytel, line 52: `T` and `N_aer` should be added to the docstring

test/heterogeneous_ice_nucleation_tests.jl
```diff
@@ -69,8 +60,24 @@ function test_heterogeneous_ice_nucleation(FT)
...
+                dSi_dt,
+                N_aer,
+            ) > CMI_het.MohlerDepositionRate(
+                dust,
+                ip.deposition,
+                Si_med,
+                T_warm,
+                dSi_dt,
+                N_aer,
+            )
```
trontrytel, line 77: Would it be possible to have a test that compares the rate and AF Mohler parameterizations? Are there any conditions where they should give similar output (for a non-zero Si case)

trontrytel (approved):
Thank you!

PR #305, opened 2024-02-02
P3 ice nucleation parameterizations

amylu00:
## To-do

- compare to our ice nucleation parameterizations in separate PR?

## Content

- added P3 heterogeneous ice nucleation parameterizations in IceNucleation.jl
- add p3 options to parcel
- make pretty plots to test it out
- add implementation in parcel and plots to documentation

### DIFF
--- a/test/gpu_tests.jl
+++ b/test/gpu_tests.jl
@@ -568,6 +568,35 @@ end
     end
 end
 
+@kernel function IceNucleation_P3_deposition_N_i_kernel!(
+    output::AbstractArray{FT},
+    ip,
+    T,
+) where {FT}
+
+    i = @index(Group, Linear)
+
+    @inbounds begin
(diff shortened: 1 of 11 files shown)


test/gpu_tests.jl
```diff
@@ -30,8 +30,8 @@ CUDA.allowscalar(false)
 const ArrayType = CuArray
 
 # For debugging on the CPU
-#const backend = CPU()
-#const ArrayType = Array
+# const backend = CPU()
```
trontrytel, line 33: should uncomment

trontrytel (approved):
Thank you so much for doing this! I left some minor comments.

It would be nice to move the free parameters of those schemes to CLIMAParameters, but I can do it later

test/gpu_tests.jl
amylu00: that's for debugging, I think it shows up as a change because of the space. I'll get rid of it tho

PR #313, opened 2024-02-06
box tracks droplet radius instead of area

amylu00:
## Purpose 

Generalize box model with individual droplet radius in state vector instead of total surface area

## Content

----


trontrytel:
@amylu00 - should we close this one for now?

amylu00:
> @amylu00 - should we close this one for now?

I'll keep the changes for the constant area box problem and just use that one for testing P3/CM.jl ice nucleation functions

PR #318, opened 2024-02-09
Aerosol activation NN

trontrytel:
This PR brings in the first part of the work of @mikhailmints from last year SURF. It:

- Creates an extension for the CM function `N_activated_per_mode` that will be available if the user has the following packages: `DataFrames`, `Flux`, `MLJ`, `MLJFlux`, `CSV`, `DataFramesMeta`, `MLJModels`
- Adds a tests that trains a NN model based on PySDM training data and compares the results with the default function.
- It would be nice to have fewer packages here (maybe we could write it without `DataFrames`? But lets refactor that once we have all models working first.)

If the `jls` file with the NN model already exists the training is skipped.


ml_test/train_emulator.jl
```diff
@@ -0,0 +1,155 @@
...
+                    init = init,
+                ),
+            )
+        end
+    end
+    return Flux.Chain(layers...)
+end
+
+# Load aerosol data reading and preprocessing functions
+include(joinpath(fpath, "Common.jl"))
```
nefrathenrici, line 62: If Common.jl will be widely used, perhaps it could live in the extension?

ml_test/train_emulator.jl
```diff
@@ -0,0 +1,155 @@
...
+N2 = FT(1e6)          # 1/m3
+acc = AM.Mode_κ(r1, σ1, N1, (FT(1.0),), (FT(1.0),), (salt.M,), (salt.κ,), 1)
+crs = AM.Mode_κ(r2, σ2, N2, (FT(1.0),), (FT(1.0),), (salt.M,), (salt.κ,), 1)
+ad = AM.AerosolDistribution((crs, acc))
+
+# Get the ML model
+mach = get_ML_model("2modal_nn_machine_naive.jls")
+
+@info("ML version: ", AA.N_activated_per_mode(mach, ap, ad, aip, tps, T, p, w, q))
+@info("CM version: ", AA.N_activated_per_mode(ap, ad, aip, tps, T, p, w, q))
```
nefrathenrici, line 155: Do we want some tests to compare the ML and CM output?

ml_test/train_emulator.jl
```diff
@@ -0,0 +1,155 @@
...
+# We will be precise
+const FT = Float64
+
+# Container for the NN we are about to build
+struct NNBuilder <: MLJFlux.Builder
+    layer_sizes::Vector{Integer}
+    dropout::Vector{Float64}
+end
+
+# TODO - what does it do exactly?
```
nefrathenrici, line 22: I agree with this TODO - this could probably be simplified.

ml_test/train_emulator.jl
```diff
@@ -0,0 +1,155 @@
...
+        end
+    end
+    return Flux.Chain(layers...)
+end
+
+# Load aerosol data reading and preprocessing functions
+include(joinpath(fpath, "Common.jl"))
+
+# Get the ML model
+function get_ML_model(machine_name)
```
nefrathenrici, line 65: The `machine_name` could be hardcoded, I think it is a bit confusing as the input argument.

nefrathenrici (commented):
I'm not sure about the MLJ/Flux specifics, but this looks like a good working start.

ml_test/train_emulator.jl
trontrytel: Agreed. I don't have the time to do it now, but at would be nice to try to get rid of some of those packages

trontrytel:
I have applied all the comments. Thank you @nefrathenrici for all the help with the extensions.

Every now and then I get this warning when training the NN. Do you see where the `Float64` is coming in?

```
[ Info: Training a new ML model. This may take a minute or two.
┌ Warning: Layer with Float32 parameters got Float64 input.
│   The input will be converted, but any earlier layers may be very slow.
│   layer = Dense(11 => 250, sigmoid_fast)  # 3_000 parameters
│   summary(x) = "11×1000 Matrix{Float64}"
```

trontrytel:
Seems like codecov gets confused about the code inside the `ext` folder. I guess there is no easy way to fix it?

test/emulator_NN.jl
```diff
@@ -0,0 +1,169 @@
...
+                    init = init,
+                ),
+            )
+        end
+    end
+    return Flux.Chain(layers...)
+end
+
+# Load aerosol data reading and preprocessing functions
+include(joinpath(pkgdir(CM), "ext", "Common.jl"))
```
nefrathenrici, line 66: I didn't realize that Common would add a lot of dependencies to the extension. Maybe it is best to have it manually included (as you are doing here) and keep its deps in the test Project.toml

nefrathenrici (approved):
This looks good! I think DataFrames really shouldn't be needed here, I can try and remove it when I find the time next week.

test/emulator_NN.jl
trontrytel: Sounds good. I'll do that, thanks!

trontrytel:
> This looks good! I think DataFrames really shouldn't be needed here, I can try and remove it when I find the time next week.

Thanks! That would be great, if you have the time.

> I think the Codecov just ignores extensions, I have not been able to find any way around it.

So sad. I'm worried my codecov stats will suffer and I will not be able to brag about test coverage anymore. Could we open an issue somewhere on the intercept of codecov and julia extensions communities (assuming it's not empty)...

PR #324, opened 2024-02-15
Bring the docs up to speed

trontrytel:
This PR brings the docs compile time back to reasonable. `112.712372 seconds (1.54 G allocations: 72.581 GiB, 5.14% gc time, 36.17% compilation time: 8% of which was recompilation)` on my laptop.  The reason for the slowdown was setting a very demanding tolerances for the ODE solver in our parcel model (especially if running with `Float64` precision). I think by default we should run everything in `Float32`

Btw, when I was running individual simulations I noticed some of them were throwing warnings like:

`WARNING: Multiple deposition nucleation parameterizations chosen to run in one simulation. Parcel will only run MohlerAF_Deposition for now.`

Is that expected behavior @amylu00 ?

### DIFF
--- a/parcel/Immersion_Freezing.jl
+++ b/parcel/Immersion_Freezing.jl
@@ -6,7 +6,7 @@ import CLIMAParameters as CP
 
 # definition of the ODE problem for parcel model
 include(joinpath(pkgdir(CM), "parcel", "parcel.jl"))
-FT = Float64
+FT = Float32
 # get free parameters
 tps = TD.Parameters.ThermodynamicsParameters(FT)
 aps = CMP.AirProperties(FT)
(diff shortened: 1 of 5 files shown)


trontrytel:
It takes about 11 minutes to get all the dependencies and less than 4 minutes to generate the docs. We could make it shorter by for example deciding on using only one package for plotting. And by making the solver tolerances in parcel even less restrictive. 

But maybe it's good enough. When building things locally it should be faster (about 2 minutes for me, not counting instantiating)

amylu00:
Oh that might be wrong. I'll try fixing it sometime, it shouldn't be showing up for the examples we currently have. I think I mistakenly used `or` conditional statements when it should've been `&&`. Since there's so many modes being added, it'll have to be done another way anyways though. Thanks for the heads up!

amylu00 (approved):
Looks good to me

trontrytel:
> Oh that might be wrong. I'll try fixing it sometime, it shouldn't be showing up for the examples we currently have. I think I mistakenly used `or` conditional statements when it should've been `&&`. Since there's so many modes being added, it'll have to be done another way anyways though. Thanks for the heads up!

Sounds good. I'll open an issue so that we don't forget

PR #325, opened 2024-02-15
Add ML calibrated ARG2000 parameters

trontrytel:
This PR adds the PySDM calibrated parameters to a toml file that we can use if we want to overwrite the default values. I add the overwritten option to some of the tests.


src/parameters/toml/ARG2000.toml
```diff
@@ -0,0 +1,29 @@
+[ARG2000_f_coeff_1]
+value = 0.26583888195264627
+type = "float"
+description = "Scaling coefficient for an empirically determined functional dependence in Abdul-Razzak and Ghan, 2000. Unitless. DOI: https://doi.org/10.1029/1999JD901161"
```
nefrathenrici, line 4: Now only the new value is required for overrides. Up to you!
```suggestion
[ARG2000_f_coeff_1]
value = 0.26583888195264627
```

nefrathenrici (approved):

src/parameters/toml/ARG2000.toml
trontrytel: Thats great. I like the short version. We shouldn't duplicate places where we define things.

PR #315, opened 2024-02-07
Refactor parameters to use ClimaParameters API

nefrathenrici:
This PR changes all existing parameter constructors to use `get_parameter_values` instead of accessing the toml dict directly. This requires a lot of lines changed, but most of it is boilerplate.

### Example for Kaolinite
Original:
```julia
function Kaolinite(
    ::Type{FT},
    toml_dict::CP.AbstractTOMLDict = CP.create_toml_dict(FT),
) where {FT}
    (; data) = toml_dict
    return Kaolinite(
        FT(data["China2017_J_deposition_m_Kaolinite"]["value"]),
        FT(data["China2017_J_deposition_c_Kaolinite"]["value"]),
        FT(data["KnopfAlpert2013_J_ABIFM_m_Kaolinite"]["value"]),
        FT(data["KnopfAlpert2013_J_ABIFM_c_Kaolinite"]["value"]),
    )
end
```
New constructor:
```julia
Kaolinite(::Type{FT}) where {FT <: AbstractFloat} =
    Kaolinite(CP.create_toml_dict(FT))

function Kaolinite(td::CP.AbstractTOMLDict)
    name_map = (;
        :China2017_J_deposition_m_Kaolinite => :deposition_m,
        :China2017_J_deposition_c_Kaolinite => :deposition_c,
        :KnopfAlpert2013_J_ABIFM_m_Kaolinite => :ABIFM_m,
        :KnopfAlpert2013_J_ABIFM_c_Kaolinite => :ABIFM_c,
    )
    parameters = CP.get_parameter_values(td, name_map, "CloudMicrophysics")
    FT = CP.float_type(td)
    return Kaolinite{FT}(; parameters...)
end
```

### Key Changes
Constructors with the form: `ParamStruct(FT, toml_dict = CP.create_toml_dict(FT))` have been changed to 
`ParamStruct(FT) = ParamStruct(CP.create_toml_dict(FT))` and `ParamStruct(toml_dict)`. 
This preserves the behavior of `ParamStruct(FT)`, but is technically an interface change since a call like `ParamStruct(FT, toml_dict)` is no longer possible.

Add a kwarg constructor for most parameter structs - this is needed for nicer integration with climaparameters.

The biggest changes are in the 1 moment Rain, Snow, and CloudIce constructors:
- Add a few new TOML-based constructors for small structs (AcnvB1994, AccrKK2000, AccrB1994, etc)
- Dispatch `ParticleMass` and `ParticleArea` constructors for Rain, Snow, and CloudIce

PR #327, opened 2024-02-16
Clean up parameter changes

nefrathenrici:
I missed changing a few constructors in #315. 

This PR changes the following structs to have separate FT and toml dict based constructors:
- IceNucleationParameters
- KK2000
- Blk1MVelType
- Chen2022VelType

This is not strictly needed since no functionality is changed but I want to standardize the parameter constructors across repos.

### DIFF
--- a/src/parameters/IceNucleation.jl
+++ b/src/parameters/IceNucleation.jl
@@ -78,14 +78,15 @@ struct IceNucleationParameters{FT, DEP, HOM} <: ParametersType{FT}
     homogeneous::HOM
 end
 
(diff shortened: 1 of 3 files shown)


trontrytel (approved):
Do we need to make a release for it?

nefrathenrici:
@trontrytel We could make a patch release, but it's not urgent since the functionality is the same.

trontrytel:
Maybe I'll do a patch release before I update the kinematic model dependencies. Thanks

PR #328, opened 2024-02-18
Parcel model with "rainbow" ice nucleation from Frostenberg et al 2023

AgnieszkaMakulska:
## Purpose 

Add a parcel model example that uses the "rainbow" parameterization the way it was described in Frostenberg et al 2023 . Test the dependency on dt, drawing frequency, etc.

## Content

----

### DIFF
--- a/docs/src/IceNucleationParcel0D.md
+++ b/docs/src/IceNucleationParcel0D.md
@@ -334,3 +334,83 @@ Shown below are three separate parcel simulations for deposition nucleation,
 include("../../parcel/Example_P3_ice_nuc.jl")
 ```
 ![](P3_ice_nuc.svg)
+
+
+## Immersion Freezing Parametrization based on Frostenberg et al. 2023
+
+The parcel model also includes the parametrization of immersion freezing based on [Frostenberg2023](@cite).
+The concentration of ice nucleating particles (INPs) is randomly selected based on a lognormal relative frequency distribution, depending only on the temperature. If the random INP concentration exceeds the existing concentration of ice crystals, new ice crystals are created, such that their total concentration is equal to the new INP concentration.
+
+Three different implementations of this parametrization are used in the parcel model:
+- **Frostenberg_mean**, in which the concentration of INPs is equal to its mean value defined in [Frostenberg2023](@cite) for a given temperature,
+- **Frostenberg_random**, in which the concentration of INPs is drawn randomly from the distribution defined in [Frostenberg2023](@cite). The time between draws is set by the *drawing_interval* parameter,
+- **Frostenberg_stochastic**, in which the freezing is treated as a stochastic process, with mean defined in [Frostenberg2023](@cite) and Wiener noise. The timescale of the process is set by the `\gamma` parameter. 
+
+
+The stochastic implementation is based on the equation for a stochastic process $x$ with Gaussian random noise:
+
+```math
+\begin{equation}
+  dx = - \gamma \; x \; dt + g dW
+\end{equation}
+```
+
+```math
+\begin{equation}
+  dW = {\bf{N}}(0, dt^2)
+\end{equation}
+```
+where ``\bf{N}`` is the lognormal distribution.
+
+```math
+\begin{equation}
+  x(t) = x_0 \; e^{-\gamma t} + g \; \int_0^t e^{\gamma(s-t)} dW
+\end{equation}
+```
+where ``1/\gamma`` is the assumed timescale of the process.
+```math
+\begin{equation}
+  {\bf{V}}(x) = g^2 \int_0^t e^{2\gamma(s-t)} ds = \frac{g^2}{2\gamma}(1 - e^{-2\gamma t})
+\end{equation}
+```
+where ``\bf{V}`` is the variance.
+We map this notation onto our problem:
+
+```math
+\begin{equation}
+  x = ln(INPC)
+\end{equation}
+```
+
+```math
+\begin{equation}
+  dx = -\gamma (x - \mu(T)) dt + g dW
+\end{equation}
+```
+
+```math
+\begin{equation}
+  x(t) = x e^{-\gamma t} + \mu(T) (1 - e^{-\gamma t}) + g \int_0^t e^{-\gamma (t-s)} dW
+\end{equation}
+```
+
+```math
+\begin{equation}
+  x(t) = {\bf{N}}(x_0 e^{-\gamma t} + (1-e^{-\gamma t}) \mu(T), \frac{g^2}{2\gamma} (1-e^{-2\gamma t}))
+\end{equation}
+```
+If ``\gamma >> 1``, then
+```math
+\begin{equation}
+  x(t) = {\bf{N}}(\mu(T), \frac{g^2}{2\gamma})
+\end{equation}
+```
+Hence ``g = \sigma * \sqrt{2 \gamma}``.
+
+
+The following plot shows resuls of the parcel model with Frostenberg_mean parametrization, Frostenberg_random parametrization with different drawing intervals `t_d` and Frostenberg_stochastic parametrization with different time scales `\gamma`. The timestep of the model is 1 s.
+
+```@example
+include("../../parcel/Example_Frostenberg_Immersion_Freezing.jl")
+```
+![](frostenberg_immersion_freezing.svg)
--- a/docs/src/IceNucleationParcel0D.md
+++ b/docs/src/IceNucleationParcel0D.md
@@ -296,7 +296,7 @@ include("../../parcel/Example_Liquid_only.jl")
 The plots below are the results of the adiabatic parcel model
   with immersion freezing, condensation growth, and deposition growth for
   both a monodisperse and gamma size distribution. Note that this has not
-  yet been validated against literature. Results are very sensitive to 
+  yet been validated against literature. Results are very sensitive to
   initial conditions.
 
 ```@example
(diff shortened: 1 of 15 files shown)


parcel/Frostenberg_Immersion_Freezing.jl
```diff
@@ -0,0 +1,97 @@
...
+const_dt = FT(1)                           # model timestep
+t_max = FT(1000)
+aerosol = CMP.Illite(FT)
+ice_nucleation_modes = ["Frostenberg"]
+growth_modes = ["Condensation", "Deposition"]
+droplet_size_distribution= ["Monodisperse"]
+
+# Plotting
+fig = MK.Figure(resolution = (800, 600))
+ax1 = MK.Axis(fig[1, 1], ylabel = "Ice Supersaturation [-]")
```
trontrytel, line 58: Can we also plot liquid water S here (on the same plot)

parcel/parcel.jl
```diff
@@ -127,6 +129,19 @@ function parcel_model(dY, Y, p, t)
         dqi_dt_new_immers = dN_act_dt_immersion * 4 / 3 * FT(π) * r_l^3 * ρ_ice / ρ_air
     end
 
+    dN_act_dt_frostenberg = FT(0)
+    dqi_dt_new_frostenberg = FT(0)
+    if "Frostenberg" in ice_nucleation_modes
+
+        T_Celsius = T - 273.15
+        μ = log(-(ip_frostenberg.b * T_Celsius)^9 * 10^(-9))
```
trontrytel, line 137: Should we add this as a separate function in the source code?

parcel/Example_Frostenberg_Immersion_Freezing.jl
```diff
@@ -0,0 +1,175 @@
...
+        deposition_growth = deposition_growth,
+        size_distribution = DSD,
+        γ = γ,
+    )
+    # solve ODE
+    sol = run_parcel(IC, FT(0), t_max, params)
+    plot_results(sol, γ, "γ=", " s", :solid, color)
+
+end
+
```
trontrytel, line 174: We don't need to include the actual `svg` file in the PR. It should be automatically generated when the documentation is build. Once you add a short description of the example to the `docs/src/IceNucleationParcel0D.md` file you should do something like what we do for the other plots there at the end of the file. (Sorry, I tried to paste the code snippet, but github tries to render it and fails)

parcel/ParcelTendencies.jl
```diff
@@ -3,6 +3,7 @@ import CloudMicrophysics.Common as CMO
 import CloudMicrophysics.HetIceNucleation as CMI_het
 import CloudMicrophysics.HomIceNucleation as CMI_hom
 import CloudMicrophysics.Parameters as CMP
+using Distributions
```
trontrytel, line 6: Could you replace `using Distributions` with `import Distributions as DS`. And then replace all the function calls to functions from the `Distributions` module with `DS.foo`. It helps to keep track of which function is defined where.

(here are some docs on it: https://docs.julialang.org/en/v1/manual/modules/#Renaming-with-as)

parcel/ParcelTendencies.jl
```diff
@@ -106,6 +107,39 @@ function immersion_freezing(params::P3_het, PSD, state)
     return max(FT(0), (Nᵢ_het - Nᵢ) / const_dt)
 end
 
+function immersion_freezing(params::Frostenberg_random, PSD, state)
+    FT = eltype(state)
+    (; ip, drawing_interval) = params
+    (; T, Nₗ, Nᵢ, t) = state
+    if mod(t, drawing_interval) == 0
+        μ = CMI_het.INP_concentration_mean(T)
+        INPC = exp(rand(Normal(μ, ip.σ)))
```
trontrytel, line 116: Would it be possible to define the seed of the random number generator used here? We could even make it one of the parameters of the `_random` and `_stochastic` functions. This way our results will be reproducible, even when random

parcel/ParcelTendencies.jl
```diff
@@ -106,6 +107,39 @@ function immersion_freezing(params::P3_het, PSD, state)
...
+    else
+        INPC = 0
+    end
+    return min(Nₗ, max(FT(0), INPC - Nᵢ))
```
trontrytel, line 120: All the `immersion_freezing` functions should be returning the freezing rate. Thats the expectation of the ODE solver  - we can't overwrite the sate, we can only define the tendencies for the solver. So you should divide what you are returning here by `dt` (It shouldn't change the results much since I think by default you were running with `dt=1s`)

parcel/Example_Frostenberg_Immersion_Freezing.jl
```diff
@@ -0,0 +1,175 @@
...
+# solve ODE
+sol = run_parcel(IC, FT(0), t_max, params)
+plot_results(sol, "mean")
+
+
+# Frostenberg_random with different drawing frequencies
+drawing_interval_range = range(FT(1), stop = FT(5), length = 3)
+
+for (drawing_interval, color) in zip(drawing_interval_range, colors)
+
```
trontrytel, line 136: Once we have random seed as one of the parameters, could you loop over a couple of different random seed realizations here and plot the avarage of them. I imagine there is some (maybe a lot) of variability here and plotting an ensemble average should clarify what is happening

trontrytel:
I left some comments on the draft already. Aside from them, could you also add a small entry to the documentation page with parcel model examples? Including the notes I have in PR 336 and including the plot with the parcel model results. Thank you, this is a really cool example!

parcel/ParcelTendencies.jl
AgnieszkaMakulska: I have some problems with passing the random seed as a parameter. I'm probably doing something wrong here, but when I try to use the random seed as a parameter to our functions, all of the results are equal. Instead, I tried to set different seeds before calling functions in the Example file. I'm attaching results here. It looks like the for the stochastic process, results average to the mean.
frostenberg_immersion_freezing

parcel/Example_Frostenberg_Immersion_Freezing.jl
```diff
@@ -0,0 +1,194 @@
...
+            deposition_growth = deposition_growth,
+            size_distribution = DSD,
+            γ = γ,
+        )
+        # solve ODE
+        sol = run_parcel(IC, FT(0), t_max, params)
+        push!(solutions, sol)
+    end
+    mean_sol = sum(solutions) / length(solutions)
+    plot_results(sol.t, sol, γ, "γ=", " s", :solid, color)
```
trontrytel, line 191:
```suggestion
    plot_results(sol.t, mean_sol, γ, "γ=", " s", :solid, color)
```

trontrytel: I think this is why the plots collapsed

parcel/ParcelTendencies.jl
trontrytel: I think you are right in how to use the seed and I think I found the typo in plotting. So maybe this will fix it?

AgnieszkaMakulska: Thank you for noticing the typo! 
With the typo fixed, I think there is still something wrong here. I'm attaching two plots: in the first one, I'm averaging different realizations with selected seeds (as in this pull request). In the second one, I'm averaging different realizations, but without specifying the seed (deleting the line `RAND.seed!(trunc(Int, random_seed))`). There is something wrong with the first plot, in which I'm specifying the seed.
frostenberg_immersion_freezing_seed
frostenberg_immersion_freezing

trontrytel:
Here is the final version (I think). Initially I wrote the equations wrong, thinking `g = sigma sqrt(2 gamma)` But that actually gives us the wrong limit and the noise in the system does not increase as we change the gamma scale. It should be `g = sigma` (duh...) 

Now we see that for large gammas (i.e. short assumed timescales) the solution converges to the mean solution. 

I'm not sure if I understand why we converge to the mean when we draw sell frequently in the random option.

Screenshot 2024-02-29 at 4 30 36 PM

trontrytel:
I cleaned up some of the names in the example. Not sure what is the least confusing way of naming/specifying the sampling interval in the `random` scheme. But at least now on the labels we have the same units (i.e. either the time between sampling in the `random` case or the assumed process time scale in the `stochastic` case). The problem is that even if the units are the same the interpretation is the opposite (I think). The larger sampling interval the more close we are to the mean. But the shorter the time scale the closer we are to the mean.

Below I show two plots with the same assumed sampling intervals and timescales, but different model `dt` of either 0.1s or 1s. It would be better to plot everything on one plot, but I'm thinking how to do it neatly. 

Screenshot 2024-02-29 at 8 10 07 PM

Screenshot 2024-02-29 at 8 10 20 PM

trontrytel:
Yet another attempt at squeezing in more plots:
Screenshot 2024-02-29 at 10 12 56 PM

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+Three different implementations of this parametrization are used in the parcel model:
+- `mean` - in which INPC is equal to its mean value defined in [Frostenberg2023](@cite) for a given temperature.
+- `random` - in which INPC is sampled randomly from the distribution defined in [Frostenberg2023](@cite).
+  The number of model time steps between sampling is set by the `sampling_interval` parameter.
+- `stochastic` - in which INPC is modelled as a stochastic process,
+  with the mean and starndard deviation defined in [Frostenberg2023](@cite).
+  The timescale of the process is set by the ``\gamma`` parameter.
+
+The stochastic implementation is based on the equation for a stochastic process ``x``
+  with Gaussian random noise:
```
haakon-e, line 357:
```suggestion
  with Gaussian random noise (i.e. Wiener process):
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+```math
+\begin{equation}
+  dx = - \gamma \; x \; dt + g dW; \;\;\;\;\;\;\;   dW = {\bf{N}}(0, dt^2)
+\end{equation}
+```
+where ``\bf{N}`` is the lognormal distribution.
```
haakon-e, line 363:
```suggestion
where ``\bf{N}`` is the normal distribution. For constant ``\gamma`` and ``g``, ``x`` has solution
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+```math
+\begin{equation}
+  x(t) = x_0 \; e^{-\gamma t} + g \; \int_0^t e^{\gamma(s-t)} dW
+\end{equation}
+```
+where ``1/\gamma`` is the assumed timescale of the process.
+```math
+\begin{equation}
+  {\bf{V}}(x) = g^2 \int_0^t e^{2\gamma(s-t)} ds = \frac{g^2}{2\gamma}(1 - e^{-2\gamma t})
```
haakon-e, line 372:
```suggestion
  {\bf{V}}(t) = g^2 \int_0^t e^{2\gamma(s-t)} ds = \frac{g^2}{2\gamma}(1 - e^{-2\gamma t})
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+
+```math
+\begin{equation}
+  x(t) = x e^{-\gamma t} + \mu(T) (1 - e^{-\gamma t}) + g \int_0^t e^{-\gamma (t-s)} dW
+\end{equation}
+```
+
+```math
+\begin{equation}
+  x(t) = {\bf{N}}(x_0 e^{-\gamma t} + (1-e^{-\gamma t}) \mu(T), \frac{g^2}{2\gamma} (1-e^{-2\gamma t}))
```
haakon-e, line 398:
```suggestion
  \log(\text{INPC})(t) = {\bf{N}} \left(\log(\text{INPC})_0 e^{- t / \tau} + (1-e^{- t / \tau}) \mu(T),\ \frac{\sigma^2\tau}{2} (1 - e^{- 2 t / \tau}) \right).
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+
+Three different implementations of this parametrization are used in the parcel model:
+- `mean` - in which INPC is equal to its mean value defined in [Frostenberg2023](@cite) for a given temperature.
+- `random` - in which INPC is sampled randomly from the distribution defined in [Frostenberg2023](@cite).
+  The number of model time steps between sampling is set by the `sampling_interval` parameter.
+- `stochastic` - in which INPC is modelled as a stochastic process,
+  with the mean and starndard deviation defined in [Frostenberg2023](@cite).
+  The timescale of the process is set by the ``\gamma`` parameter.
+
+The stochastic implementation is based on the equation for a stochastic process ``x``
```
haakon-e, line 356:
```suggestion
The stochastic implementation is based on the equation for a generic stochastic process ``x``
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+  with Gaussian random noise:
+```math
+\begin{equation}
+  dx = - \gamma \; x \; dt + g dW; \;\;\;\;\;\;\;   dW = {\bf{N}}(0, dt^2)
```
haakon-e, line 360:
```suggestion
  dx = - \gamma \; x \; dt + g dW; \;\;\;\;\;\;\;   dW = {\bf{N}}(0, dt^2),
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+\end{equation}
+```
+where ``\bf{N}`` is the lognormal distribution.
+```math
+\begin{equation}
+  x(t) = x_0 \; e^{-\gamma t} + g \; \int_0^t e^{\gamma(s-t)} dW
+\end{equation}
+```
+where ``1/\gamma`` is the assumed timescale of the process.
```
haakon-e, line 369:
```suggestion
where ``1/\gamma`` is the assumed timescale of the process. We can calculate the variance ``\bf{V}`` as,
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+```math
+\begin{equation}
+  {\bf{V}}(x) = g^2 \int_0^t e^{2\gamma(s-t)} ds = \frac{g^2}{2\gamma}(1 - e^{-2\gamma t})
+\end{equation}
+```
+where ``\bf{V}`` is the variance.
```
haakon-e, line 375:
```suggestion

```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+  and is based on a lognormal relative frequency distribution.
+New ice crystals are created if the INPC exceeds the existing concentration of ice crystals,
+  provided there are sufficient numbers of cloud liquid droplets to freeze.
+
+Three different implementations of this parametrization are used in the parcel model:
+- `mean` - in which INPC is equal to its mean value defined in [Frostenberg2023](@cite) for a given temperature.
+- `random` - in which INPC is sampled randomly from the distribution defined in [Frostenberg2023](@cite).
+  The number of model time steps between sampling is set by the `sampling_interval` parameter.
+- `stochastic` - in which INPC is modelled as a stochastic process,
+  with the mean and starndard deviation defined in [Frostenberg2023](@cite).
```
haakon-e, line 353:
```suggestion
  with the mean and standard deviation defined in [Frostenberg2023](@cite).
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+```
+where ``1/\gamma`` is the assumed timescale of the process.
+```math
+\begin{equation}
+  {\bf{V}}(x) = g^2 \int_0^t e^{2\gamma(s-t)} ds = \frac{g^2}{2\gamma}(1 - e^{-2\gamma t})
+\end{equation}
+```
+where ``\bf{V}`` is the variance.
+
+We map this notation onto our problem:
```
haakon-e, line 377:
```suggestion
We use this process to model the logarithm of INPC (``x=\log(\text{INPC})``), which tends toward the mean ``\mu(T)``. If we denote ``\tau=1 / \gamma`` as the process timescale and ``\sigma=g`` as the process uncertainty, then the solution of the stochastic process reads:
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+```math
+\begin{equation}
+  x = ln(INPC)
+\end{equation}
```
haakon-e, line 381:
```suggestion

```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+```
+
+```math
+\begin{equation}
+  dx = -\gamma (x - \mu(T)) dt + g dW
+\end{equation}
+```
```
haakon-e, line 388:
```suggestion

```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+
+```math
+\begin{equation}
+  x(t) = x e^{-\gamma t} + \mu(T) (1 - e^{-\gamma t}) + g \int_0^t e^{-\gamma (t-s)} dW
+\end{equation}
+```
```
haakon-e, line 394:
```suggestion

```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+
+```math
+\begin{equation}
+  x(t) = {\bf{N}}(x_0 e^{-\gamma t} + (1-e^{-\gamma t}) \mu(T), \frac{g^2}{2\gamma} (1-e^{-2\gamma t}))
+\end{equation}
+```
+In the limit of ``\gamma >> 1``, we want ``x(t) = \mu(T)``.
+Hence ``g = \sigma``.
```
haakon-e, line 402:
```suggestion
With this formulation, notice that ``\log(\text{INPC}) \approx \mu(T)`` whenever ``\tau \ll 1``.
```

docs/src/IceNucleationParcel0D.md
```diff
@@ -334,3 +334,80 @@ Shown below are three separate parcel simulations for deposition nucleation,
...
+
+The following plot shows resuls of the parcel model with the `mean` (black line),
+ `random` (dotted lines) and `stochastic` (solid lines) parameterization options.
+The `random` option uses different sampling intervals `t_s` and
+ the `stochastic` option uses different process time scales ``/\gamma``.
```
haakon-e, line 407:
```suggestion
 the `stochastic` option uses different process time scales ``\tau``.
```

haakon-e (commented):
https://en.wikipedia.org/wiki/Euler–Maruyama_method

haakon-e:
for future implementation: https://github.com/SciML/StochasticDiffEq.jl

PR #329, opened 2024-02-19
fixing immersion freezing production rate

amylu00:
## Content

- immersion freezing production rate proportional to INP surface area, _not_ the droplet's surface area

----

### DIFF
--- a/box/Alpert_Knopf_2016_forward.jl
+++ b/box/Alpert_Knopf_2016_forward.jl
@@ -20,7 +20,7 @@ end
 include(joinpath(pkgdir(CM), "box", "box.jl"))
 
 # initial conditions following KA16 figure 4 (Cr1 case)
-A_droplet = FT(1e-5) * 1e-4 # INP surface area, m^2
+A_aero = FT(1e-5) * 1e-4 # INP surface area, m^2
 σg = 10
 N₀ = 1000
 N_ice = 0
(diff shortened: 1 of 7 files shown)


amylu00:
Before the fix:
Screenshot 2024-02-19 at 12 55 12 PM

After the fix:
Screenshot 2024-02-19 at 12 55 44 PM

`q_liq` is stabilized for mono disperse size distribution now. Ice supersaturation and `q_ice` match up better, but not sure if that is an expected behavior or not.

amylu00:
Also fixing the deposition warning issue here since it's only a few lines of print code. If you think it's better to have it in a separate PR, I can do that

trontrytel (approved):
LGTM!

We could also have a separate parameter for the assumed aeorsol area and not have to re-use the r_nuc which is used in the deposition freezing. I think this will be helpful once you start running simulations with more than just one mode of freezing.

Also, why are you changing the initial condition for the immersion freezing example?

Lmk if you need help adapting it to the new parcel setup. Tbh it might be easier to "manually" add your changes rather than try to rebase this

amylu00:
> Also, why are you changing the initial condition for the immersion freezing example?

I was getting situations where there was very little to no ice nucleation for the monodisperse case. If that doesn't seem right, we can discuss it tomorrow?

PR #331, opened 2024-02-21
Refactor parcel

trontrytel:
## Purpose 

Refactor the adiabatic parcel code to get rid of convoluted if/else logic that frew on us

## Content

The parcel model code is now structured into
- `Parcel.jl` - includes all the files. Needs to be included in all examples
- `ParcelCommon.jl` - defines some functions that are reused when computing initial conditions, etc.
- `ParcelDistributions.jl` - computes the mean radius, area and volume for particles for monodisperse, Gamma and Lambda distributions
- `ParcelModel.jl` - defines the parcel problem for ODE solver
- `ParcelParameters.jl` - creates structs with parameters that will be used to dispatch over different parameterizations
- `ParcelTendencies.jl` - defines different ice nucleation and growth tendencies. Note that for each of them the first argument is a parameter struct that serves as type for dispatching.

## To-do

Couple of problems that I noticed while coding

- I think there is something wrong with the `Lognormal` distribution. It seems like it is not connected at all with the predicted `ql` and `Nl`. Is it meant to be used only for background aerosol?
- In a couple of places we were using <r>^3 instead of <r^3>. For example [here]( https://github.com/CliMA/CloudMicrophysics.jl/blob/7fbe3443c5cd4f5afec594601dfdf8e86e433667/parcel/parcel.jl#L132). This should be fixed now thanks to the `ParcelDistributions`
- The `Jensen` example gives different results than before. I think it is because of the issues with the `Lognormal` size distribution.


parcel/ParcelDistributions.jl
```diff
@@ -0,0 +1,79 @@
...
+        rᵢ = FT(0)
+    else
+        λᵢ = cbrt(32 * FT(π) * Nᵢ / qᵢ * ρᵢ / ρ_air)
+        rᵢ = 2 / λᵢ
+    end
+    return (; rₗ, Aₗ, Vₗ, rᵢ)
+end
+
+# Lognormal r = R_mode_liq, σ
+# Jensen r = r_nuc, σ
```
amylu00, line 64: Jensen makes the aerosol have a lognormal distribution -- not the droplets

amylu00 (changes requested):
Thank you so much Anna! It looks beautiful! 

- I think you may have used the deposition water activity based parameterization for Jensen instead of the homogeneous one. Maybe those can be renamed to avoid confusion in the future. This might fix the problem of the plot not looking the same but let me know if otherwise. Jensen also assumes the _aerosols_ are lognormally distributed and not the droplets so that may also be affecting the results.
- I think lognormal droplets account for q_liq and N_liq through R_mode_liq and σ. I'll see if I can find out how they relate and fix it. For now, none of the files actually use it so it's fine but maybe we should have it throw a warning? I'll open an issue rn

AgnieszkaMakulska (approved):

parcel/ParcelTendencies.jl
```diff
@@ -0,0 +1,172 @@
...
+    e = eᵥ(qᵥ, p_air, R_air, Rᵥ)
+
+    # TODO - use CLIMAParameters
+    Δa_w =
+        T > FT(185) && T < FT(235) ?
+        CMO.a_w_xT(H₂SO₄ps, tps, xS, T) - CMO.a_w_ice(tps, T) :
+        CMO.a_w_eT(tps, e, T) - CMO.a_w_ice(tps, T)
+
+    if T < FT(185) || T > FT(235)
+        @warn("Clipping Δa_w for ABIFM immersion freezing")
```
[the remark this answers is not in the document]
trontrytel: 👍 Sounds good to me. We should then remove the `x_sulph` altogether from the model, right? 

I'l remove the comment, and lets remove the sulphur for now. We won't be modeling that any time soon in Atmos model anyway

parcel/ParcelDistributions.jl
trontrytel: Then I think we should just have the `Monodisperse` and `Gamma` options for liquid and ice. And `Lognormal` for background aerosol, if needed. And then it makes sense to just prescribe the mode and variance. I'll see if the examples we have right now agree with me

trontrytel: Fixed the size distribution, but it still looks different than before.

amylu00 (approved):
I love it

PR #333, opened 2024-02-22
Add "Flexible Microphysics" option, based on Cloudy.jl

edejong-caltech:
## Purpose 
Creates a flexible many-moment microphysics option based on the computational framework of Cloudy.jl.

## To-do
- [x] Make Cloudy dependency an extension
- [x] Add physical default parameters (see [Cloudy.jl issue ](https://github.com/CliMA/Cloudy.jl/issues/139))
- [x] Test condensation in parcel model
- [x] Refactor the sedimentation to return a moment_weighted terminal velocity

## Next steps:
- [ ] Test coalescence in box model (maybe)
- [ ] Test condensation + coalescence + sedimentation in (KiD or CliMA Atmos single column)

## Content
- [x] Link existing dynamics in a new module
- [x] Write unit tests for new Cloudy dynamics 

Note: GPU and performance tests are not and will not be included in the initial integration.

----


docs/src/MicrophysicsFlexible.md
```diff
@@ -0,0 +1,39 @@
+# Microphysics Flexible
+
```
trontrytel, line 2: You should also add the docs page you are creating to the docs make file. Maybe somewhere here:

https://github.com/CliMA/CloudMicrophysics.jl/blob/e2e1a5450c14c6bd8b1c985aa6ff1d7937f1b698/docs/make.jl#L41

docs/src/MicrophysicsFlexible.md
```diff
@@ -0,0 +1,39 @@
...
+Options for dynamics and size distributions are under continuous development in the `Cloudy.jl` package, thus only the default and suggested use cases are described in detail here.
```
trontrytel, line 5: Maybe it would be good to add a link to Cloudy documentation or repo somewhere?

docs/src/MicrophysicsFlexible.md
```diff
@@ -0,0 +1,39 @@
...
+from the Julia REPL. Upon recognizing that `Cloudy.jl` is being loaded, the extension `CloudyExt.jl` will then be loaded and overwrite the function stubs defined in `src/MicrophysicsFlexible.jl`.
+
+## Setting up a system
+All the details from the number of moments and type of subdistributions, to the parameterizations of coalescence, condensation, and sedimentation are defined through the `CLSetup` (CLoudySetup) mutable struct. This struct is mutable specifically because certain of its components, such as backend-computed coalescence tendencies, are updated prior to being passed to the timestepper. The components of a `CLSetup` object and their defaults are further described below.
+
+|   component         |   description                            |   default                  |
+|---------------------|------------------------------------------|----------------------------|
+| ``pdists``          | Vector of subdistributions corresponding | ``[Exponential, Gamma]``   |
+|                     | to the moments                           |                            |
+| ``mom``             | Prognostic moments, in the same order as | ``[0, 0, 0, 0, 0]``        |
```
trontrytel, line 32: Does it mean that by default I have two distributions - one `Exponential` and one `Gamma`? Which sizes do they refer to? And why by default do I have the `0` moment 5 times? (It probably just shows I should read Cloudy docs...

src/MicrophysicsFlexible.jl
```diff
@@ -0,0 +1,72 @@
...
+ - `T` - air temperature
+Returns a vector of moment tendencies due to condensation/evaporation
+"""
+function condensation end
+
+"""
+    weighted_vt(clinfo)
+
+ - `clinfo` - kwarg structure containing pdists, moments, and coalescence parameters
+Returns the integrated fall speeds corresponding to the rate of change of prognostic moments
```
trontrytel, line 68: Would it be correct to say moment-weighted terminal velocities?

docs/src/MicrophysicsFlexible.md
```diff
@@ -0,0 +1,39 @@
...
+|                     | the corresponding subdistributions       |                            |
+| ``KernelFunc``      | Form of the coalescence kernel function  | ``LongKernelFunction``     |
+| ``mass_thresholds`` | Particle size thresholds for coalescence | ``[10.0, Inf]``            |
+|                     | integration                              |                            |
+| ``kernel order``    | Polynomial order for the approx. kernel  | ``1``                      |
```
trontrytel, line 37: The `CLSetup` has more fields than what the table shows? Are they not meant to be changed by the user?

ext/CloudyExt.jl
```diff
@@ -0,0 +1,158 @@
...
+        CL.get_moments_normalizing_factors(clinfo.NProgMoms, clinfo.norms)
+    mom_normalized = clinfo.mom ./ mom_norms
+    # first, update the particle distributions
+    for (i, dist) in enumerate(clinfo.pdists)
+        ind_rng = CL.get_dist_moments_ind_range(clinfo.NProgMoms, i)
+        CPD.update_dist_from_moments!(dist, mom_normalized[ind_rng])
+    end
+    return CL.Condensation.get_cond_evap(
+        S - 1,
+        (; ξ = ξ, pdists = clinfo.pdists),
```
trontrytel, line 137: This will probably not work on the GPUs. But lets not worry about it now

test/microphysicsflexible_tests.jl
```diff
@@ -0,0 +1,83 @@
...
+            kernel_limit = FT(100),
+        )
+
+        # Create Cloudy struct with defaults
+        clinfo = CMF.CLSetup{FT}()
+        TT.@test length(clinfo.pdists) == 2
+        TT.@test length(clinfo.mom) == 5
+        TT.@test length(clinfo.NProgMoms) == 2
+
+        # Test coalescence
```
trontrytel, line 46: Should the mass be conserved in this approach when doing collisions? Could we test that?

parcel/Example_NMoment_condensation.jl
```diff
@@ -0,0 +1,178 @@
...
+import CloudMicrophysics.Parameters as CMP
+import CloudMicrophysics.MicrophysicsFlexible as CMF
+import ClimaParams as CP
+
+FT = Float64
+
+"""
+    ODE problem definitions
+"""
+function parcel_model_cloudy(dY, Y, p, t)
```
trontrytel, line 16: I'll think about how to unify the OG parcel model with the Cloudy parcel model. But it's ok to have two separate ones for now. My only concern is that we won't see anywhere if this example stops working. So you could add the plot to the docs somewhere (then the docs will break if things break). Or move it to tests.

trontrytel (approved):

sajjadazimi (approved):

docs/src/MicrophysicsFlexible.md
edejong-caltech: 👍

docs/src/MicrophysicsFlexible.md
edejong-caltech: 👍

docs/src/MicrophysicsFlexible.md
edejong-caltech: Actually these have changed, and the Cloudy docs aren't helpful... I"ll try to make it more self explanatory

src/MicrophysicsFlexible.jl
edejong-caltech: Yes!

docs/src/MicrophysicsFlexible.md
edejong-caltech: Correct, unless the user really knows what they're doing ;)

test/microphysicsflexible_tests.jl
edejong-caltech: Yes, but these tests are already covered in Cloudy.jl so I think it would be redundant here

edejong-caltech: These tests are only to check that the new features/structs that are created in CM.jl work. But, if you think it would be useful to create some redundant tests, I can absolutely do so! Let's open an issue if so.

trontrytel: It's fine. Lets add as many tests as is necessary to test that Cloudy can be used in CM.

PR #338, opened 2024-02-26
Move GPU CI to the new expansion

trontrytel:


.buildkite/pipeline.yml
```diff
@@ -24,6 +22,7 @@ steps:
       - julia --project=test -e 'using CUDA; CUDA.precompile_runtime()'
       - julia --project=test -e 'using Pkg; Pkg.status()'
     agents:
+      partition: gpu
```
Sbozzolo, line 25: I think you can remove this if you want. It should pick up the gpu queue anyway.

Sbozzolo (commented):
You also have to remove the [JuliaProject.toml](https://github.com/CliMA/CloudMicrophysics.jl/blob/main/.buildkite/JuliaProject.toml) file

Sbozzolo (approved):

.buildkite/pipeline.yml
```diff
@@ -1,15 +1,13 @@
 agents:
   queue: central
```
Sbozzolo, line 2:
```suggestion
  queue: new-central
```

I approved but we need this.

trontrytel: argh! Thank you

PR #339, opened 2024-02-27
Switch solver to q/n

anastasia-popova:
## Content

----

### DIFF
--- a/docs/src/P3Scheme.md
+++ b/docs/src/P3Scheme.md
@@ -119,7 +119,7 @@ N_{ice} = \int_{0}^{\infty} \! N'(D) \mathrm{d}D = \int_{0}^{\infty} \! N_{0} D^
 
 ``q_{ice}`` depends on the variable mass-size relation ``m(D)`` defined above.
 We solve for ``q_{ice}`` in a piece-wise fashion defined by the same thresholds as ``m(D)``.
-As a result ``q\_{ice}`` can be expressed as a sum of inclomplete gamma functions.
+As a result ``q_{ice}`` can be expressed as a sum of inclomplete gamma functions.
   and the shape parameters are found using iterative solver.
 
 |      condition(s)                            |    ``q_{ice} = \int \! m(D) N'(D) \mathrm{d}D``                                          |         gamma representation          |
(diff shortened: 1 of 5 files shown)


docs/src/P3Scheme.md
```diff
@@ -133,18 +133,29 @@ As a result ``q\_{ice}`` can be expressed as a sum of inclomplete gamma function
 where ``\Gamma \,(a, z) = \int_{z}^{\infty} \! t^{a - 1} e^{-t} \mathrm{d}D``
   and ``\Gamma \,(a) = \Gamma \,(a, 0)`` for simplicity.
 
+Within our solver, we approximate ``\mu`` from q/N and keep it constant throughout the solving step. We approximate ``\mu`` by an exponential function given by the q/N points corresponding to ``\mu = 6`` and ``\mu = 0``. This is shown below as well as how this affects the solvers ``\lambda`` solutions.
```
trontrytel, line 136: Is it exponential or linear approximation?

src/P3Scheme.jl
```diff
@@ -219,24 +219,79 @@ end
...
+function μ_approx(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
+    # Get thresholds for given F_r, ρ_r
+    th = thresholds(p3, ρ_r, F_r)
+
+    # Get min and max lambda values
+    λ_0 = μ_to_λ(p3, FT(0))
+    λ_6 = μ_to_λ(p3, p3.μ_max)
+
+    # Get corresponding q/N values at given F_r
+    q_min = log(q_gamma(p3, F_r, log(λ_0), FT(0), th))
```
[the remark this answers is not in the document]
trontrytel: And I the local variable names here should be `q_over_N_min` and `q_over_N_max` ?

src/P3Scheme.jl
```diff
@@ -219,24 +219,79 @@ end
...
+    q_max = log(q_gamma(p3, F_r, log(λ_6), p3.μ_max, th))
+
+    # Return approximation between them
+    return (p3.μ_max / (q_max - q_min)) * (log(q / N) - q_min)
+end
+
+"""
+    DSD_μ(p3, q, N, ρ_r, F_r)
```
trontrytel, line 260: Do we need both `μ_approx` and `DSD_μ(p3, q, N, ρ_r, F_r)` ? Would it be enough to combine them and have something like `DSD_μ_approx`?

src/P3Scheme.jl
```diff
@@ -246,12 +301,12 @@ function DSD_N₀(p3::PSP3, N::FT, λ::FT) where {FT}
 end
 
 """
-    q_(p3, ρ, F_r, N_0, λ, μ, D_min, D_max)
+    q_(p3, ρ, F_r, λ, μ, D_min, D_max)
 
  - p3 - a struct with P3 scheme parameters
  - ρ - bulk ice density (ρ_i for small ice, ρ_g for graupel) [kg/m^3]
  - F_r - rime mass fraction [q_rim/q_i]
- - N_0 - intercept parameter of N′ gamma distribution
```
trontrytel, line 254: This now also returns q/N and not mass density? We should update the docstring below

src/P3Scheme.jl
```diff
@@ -262,69 +317,70 @@ end
...
-    return α_va_si(p3) * N_0 / (1 - F_r) / λ^x *
+function q_r(p3::PSP3, F_r::FT, μ::FT, λ::FT, D_min::FT) where {FT}
+    x = μ + p3.β_va + 1
+    return α_va_si(p3) / (1 - F_r) / λ^x *
            (Γ(x) + Γ(x, λ * D_min) - (x - 1) * Γ(x - 1))
 end
 
 """
-    q_gamma(p3, F_r, N, λ, th)
+    q_gamma(p3, F_r, λ, th)
```
trontrytel, line 343: Lets change the name of the function accordingly

docs/src/P3Scheme.md
```diff
@@ -133,18 +133,29 @@ As a result ``q\_{ice}`` can be expressed as a sum of inclomplete gamma function
...
+
+```@example
+include("plots/P3LambdaErrorPlots.jl")
+```
+![](μApprox.svg)
```
trontrytel, line 141: Just in case, lets avoid unicode characters like `μ` in the file names

docs/src/P3Scheme.md
anastasia-popova: It is linear on a log scale of q/N so it is exponential but appears linear on our plots. I did also try a purely linear approximation and I believe it is slightly worse (though still works)

docs/src/P3Scheme.md
```diff
@@ -133,18 +133,29 @@ As a result ``q\_{ice}`` can be expressed as a sum of inclomplete gamma function
...
+
 An initial guess for the non-linear solver is found by approximating the gamma functions as a simple power function. 
 
 ```@example
 include("plots/P3ShapeSolverPlots.jl")
 ```
 ![](SolverInitialGuess.svg)
 
-This equation is given by ``(log(q_{approx}) - log(q1)) = slope (log(\lambda) - log(p1))``. Solving for ``q_{approx}`` we get `` q_{approx} = q_1 \frac{\lambda}{p_1} ^{slope}`` where `` slope = \frac{log(q1) - log(q2)}{log(p1) - log(p2)}``, ``p1`` and ``p2`` are defining ``\lambda`` values of the estimated line (we use p1 = 1e2, p2 = 1e6), ``q1 = q(p1)`` and ``q2 = q(p2)`` are the corresponding calculated q values for the given ``F_r`` and ``\rho_r`` values. 
+This equation is given by ``(log(q_{approx}) - log(q1)) = slope (log(\lambda) - log(p1))``. Solving for ``q_{approx}`` we get `` q_{approx} = q_1 \frac{\lambda}{p_1} ^{slope}`` where `` slope = \frac{log(q1) - log(q2)}{log(p1) - log(p2)}``, ``p1`` and ``p2`` are defining ``\lambda`` values of the estimated line decided off of the ``q/N`` value (described below), ``q1 = q(p1)/N`` and ``q2 = q(p2)/N`` are the corresponding calculated q values for the given ``F_r`` and ``\rho_r`` values.
```
trontrytel, line 150: I'm wondering if it would make it simpler to explain the initial guess by plotting log of lambda as a function of log of q/N. Instead of q/N as a function of lambda on a log scale. Also, for simpler notation we can use y=log(lambda) x=log(q/N) etc. Then I think it would be almost immediate to introduce the equation you use for the initial guess?

trontrytel: This would also simplify the `p1`, `p2`, `q1`, `q2` notation. I think sometimes those `q` are actually `q/N`?

test/p3_tests.jl
```diff
@@ -177,20 +177,22 @@ function test_p3_shape_solver(FT)
...
-                            ρ_r,
-                            F_r,
-                        )
-
-                        # Compare solved values with the input expected values
-                        TT.@test λ ≈ λ_ex rtol = ep
-                        TT.@test N₀ ≈ N₀_ex rtol = ep
+                        q_calc = N * P3.q_gamma(p3, F_r, x, μ_ex, th)
+
+                        if q_calc < FT(1)
```
trontrytel, line 182: `q = 1 kg/kg` is still too large to be physical. Lets cut it off at about `0.05`. (50 grams per kilogram is still larger than what we would normally expect)

Do we gen enough testing points like that. And if yes, is the error on them smaller than eps=1?

docs/src/plots/P3ShapeSolverPlots.jl
```diff
@@ -14,32 +14,42 @@ function guess_value(λ::FT, p1::FT, p2::FT, q1::FT, q2::FT)
 end
 
 function lambda_guess_plot(F_r::FT, ρ_r::FT) where {FT}
-    N = FT(1e8)
+    N = FT(1e6)
```
trontrytel, line 17: Could you plot it for a couple of different `N`, `F_r` and `rho_r` combinations?

trontrytel (approved):
Left some small comments mostly trying to see if we can make the documentation of the solver easier. Looks great otherwise. It's great to see that the solver is starting to work more reliably.

PR #340, opened 2024-02-28
Add a sanity check for GPU pipeline

trontrytel:
Add a tests agains this issue: https://github.com/CliMA/SurfaceFluxes.jl/issues/142

### DIFF
--- a/test/gpu_tests.jl
+++ b/test/gpu_tests.jl
@@ -121,6 +121,24 @@ end
     end
 end
 
+@kernel function test_1_moment_micro_smooth_transition_kernel!(
+    acnv1M,
+    output::AbstractArray{FT},
+    ql,
+) where {FT}
+
+    i = @index(Group, Linear)
+
+    output[1, 1] = FT(-99999.99)
+    output[1, 2] = FT(-99999.99)
+    output[2, 1] = FT(-99999.99)
+    output[2, 2] = FT(-99999.99)
+    @inbounds begin
+        output[1, i] = CM1.conv_q_liq_to_q_rai(acnv1M, ql[i], false)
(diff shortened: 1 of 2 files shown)


test/gpu_tests.jl
```diff
@@ -121,6 +121,28 @@ end
     end
 end
 
+@kernel function test_1_moment_micro_smooth_transition_kernel!(
+    acnv1M,
+    output::AbstractArray{FT},
+    ql
+) where {FT}
+
+    i = @index(Group, Linear)
```
charleskawczynski, line 130: I think we need to use `Global` here? xref: https://github.com/CliMA/SurfaceFluxes.jl/issues/142

trontrytel: Thank you! I wonder how the tests are passing with that. 

I'll fix it following what you did in SurfaceFluxes

test/gpu_tests.jl
```diff
@@ -747,6 +761,19 @@ function test_gpu(FT)
...
+        (; output, ndrange) = setup_output(dims, FT)
+        ql = ArrayType([FT(1e-3), FT(5e-4)])
+
+        kernel! =
+            test_1_moment_micro_smooth_transition_kernel!(backend, work_groups)
+        kernel!(rain.acnv1M, output, ql; ndrange)
+
+        # Sanity checks for the GPU KernelAbstractions workflow
+        # See https://github.com/CliMA/SurfaceFluxes.jl/issues/142
+        @test !any(isequal(Array(output), -99999.99))
```
charleskawczynski, line 774: Is `output` filled with `-99999.99` when its initialized?

charleskawczynski: If so, looks good!

PR #343, opened 2024-02-29
Perfect model calibration of ABDINM & ABIFM

amylu00:
## Purpose 

Add perfect model calibrations for ABDINM and ABIFM. Will add ABHOM calibrations in separate PR after adding a linear parameterization for ABHOM.

## Content

 - calibration for kaolinite ABDINM and ABIFM in perfect model setting
 - test for calibrations mentioned above
 - small fixes to parcel (some functions didn't pull all required variables from argument)

----


parcel/ParcelTendencies.jl
```diff
@@ -57,7 +57,7 @@ end
 function deposition_nucleation(params::ABDINM, state, dY)
     FT = eltype(state)
     (; tps, aerosol, r_nuc, const_dt) = params
-    (; T, p_air, qᵥ, qₗ, qᵢ) = state
+    (; T, p_air, qᵥ, qₗ, qᵢ, Nₐ) = state
```
trontrytel, line 60: Could I ask you to check the other functions here? I think it's not the only one that doesn't grab `Na` from `state`?

test/ice_nucleation_calibration.jl
```diff
@@ -0,0 +1,278 @@
+import EnsembleKalmanProcesses as EKP
```
trontrytel, line 1: We should add this file to `runtestst.jl` . Otherwise it will not be executed in the GHA tests.

test/ice_nucleation_calibration.jl
```diff
@@ -0,0 +1,278 @@
...
+
+    return [m_coeff_ekp, c_coeff_ekp, coeff_true[1], coeff_true[2]]
+end
+
+function test_J_calibration(FT, IN_mode)
+    output = calibrate_J_parameters(FT, IN_mode)
+    calibrated_parameters = [output[1], output[2]]
+    coeff_true = [output[3], output[4]]
+
+    TT.@test calibrated_parameters[1] ≈ coeff_true[1] rtol = FT(3.0)
```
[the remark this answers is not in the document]
amylu00: I think it's more related to the noise I added to the model output (ICNC) is our pseudo data. I could try using smaller noise or making the relative tolerance a percent of the true coefficients. It's big because the order of magnitude for the m coefficient is around 50.

test/ice_nucleation_calibration.jl
```diff
@@ -0,0 +1,278 @@
...
+import Thermodynamics as TD
+
+import Test as TT
+
+# definition of the ODE problem for parcel model
+include(joinpath(pkgdir(CM), "parcel", "Parcel.jl"))
+
+# Define model which wraps around parcel and overwrites calibrated parameters
+function run_model(p, coefficients, IN_mode, FT, IC)
+    # grabbing parameters
```
trontrytel, line 21: Is the calibration setup you are using in the tests the same as in the papers examples? This seems like a lot of code duplication. 

Would it make more sense to define the calibration once somewhere. And then use it in tests and also use the output of it to generate plots?

test/ice_nucleation_calibration.jl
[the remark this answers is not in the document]
trontrytel: Relative tolerance is by definition the percentage?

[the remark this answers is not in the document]
amylu00: You're so right, my bad. I just fixed it to 15% and adjusted the prior for ABIFM so that it would fall in that range. Thanks Anna!

papers/ice_nucleation_2024/perfect_model_J.jl
```diff
@@ -0,0 +1,270 @@
...
+
+    dim_output = length(G_truth)
+    Γ = 0.005 * LinearAlgebra.I * (maximum(G_truth) - minimum(G_truth))
+    noise_dist = Distributions.MvNormal(zeros(dim_output), Γ)
+    y_truth = G_truth .+ rand(noise_dist)
+
+    # Generate initial ensemble and set up EKI
+    N_ensemble = 10     # runs N_ensemble trials per iteration
+    N_iterations = 15    # number of iterations the inverse problem goes through
+    initial_ensemble = EKP.construct_initial_ensemble(rng, prior, N_ensemble)
```
trontrytel, line 219: todo - why Float32 doesn't work?

amylu00: `initial_ensemble`, `y_truth`, and `Γ` are Matrix{Float64}, Matrix{Float64}, and UniformScaling{64} respectively (`y_truth` being dependent on `Γ`). `Γ` is some noise multiplied by the identity matrix from the LinearAlgebra package, `I`, which is of the type UniformScaling. I can't use a Matrix{Float64} for the identity matrix because then `EKP.EnsembleKalmanProcess` gets confused and breaks

papers/ice_nucleation_2024/plot_ensemble_mean.jl
```diff
@@ -0,0 +1,116 @@
...
+iterations = collect(1:size(ABDINM_output[5])[1])
+
+ABDINM_calibrated_parameters = [ABDINM_output[1], ABDINM_output[2]]
+ABDINM_coeff_true = [ABDINM_output[3], ABDINM_output[4]]
+ABIFM_calibrated_parameters = [ABIFM_output[1], ABIFM_output[2]]
+ABIFM_coeff_true = [ABIFM_output[3], ABIFM_output[4]]
+
+# ensemble_means(ϕ_n_values, N_iterations, ensembles)
+ABDINM_ensemble_means = ensemble_means(
+    ABDINM_output[5],
```
trontrytel, line 21: If you want, I think it's fine to switch off the formatter for plotting

trontrytel (approved):
LGTM!

papers/ice_nucleation_2024/perfect_model_J.jl
odunbar: You should be able to use FT32 or FT64 as the FT is inferred from the inputs consistently. 

could you say more about this error with `I` vs `Matrix`? I haven't come across this.

amylu00: Yeah but when I use `FT32`, I get the following error:

Screenshot 2024-04-05 at 1 39 36 PM

So it seems like` Matrix{Float32}` doesn't count as an `AbstractMatrix{FT}`? And if I further use `identity_matrix = Matrix{FT}(LinearAlgebra.I, dim_output, dim_output)` instead of using the built in identity matrix `I`, I get an error that looks like this:

Screenshot 2024-04-05 at 1 51 48 PM

odunbar: These errors are both coming from a mismatch where the `update_ensemble!` or `EnsembleKalmanProcess(...)` is being applied to a function with a bunch of FT64 arguments and one FT32 argument. 
The way it's set up is that either everything needs to be one type or the other. for example you would need the following to use FT32 
```
 EKI_obj = EKP.EnsembleKalmanProcess(
        initial_ensemble, # must contain FT32 data
        y_truth, # must be a FT32 vector
        Γ, # must have FT32 entries
        EKP.Inversion();
        rng = rng,
    )
```

odunbar: Happy to chat further about this if the FloatType is important to preserve!

amylu00: Yeah I just realized that even though I defined everything as `FT`, calling `length(some_array)` returns a `Float64` specifically. Fixing it now. Thanks Ollie!

PR #349, opened 2024-03-10
Aerosol activation emulators: GP, ET, and ARG-informed

sajjadazimi:
This PR follows #318 with adding the remaining parts of the work of @mikhailmints. Here we add Gaussian processes and Evotrees. Also, informing data with ARG for training is enabled.


ext/Common.jl
```diff
@@ -68,6 +71,67 @@ function preprocess_aerosol_data(X::DataFrame)
     return X
 end
 
+function get_ARG_act_frac(data_row::NamedTuple)
+
+    FT = Float32
```
trontrytel, line 76: Would it be possible to pass in the FT and the parameter structs as arguments, instead of defining them here?

test/aerosol_activation_emulators.jl
```diff
@@ -0,0 +1,362 @@
...
+# Get the ML model
+function get_2modal_model_FT32(;
+    ml_model = "NN",
+    with_ARG_act_frac = false,
+    machine_name = "2modal_nn_machine_naive.jls",
+)
+    FT = Float32
+
+    # If the ML model already exists load it in.
+    # If it does not exist, train a NN
```
trontrytel, line 168:
```suggestion
    # If it does not exist, train
```

test/aerosol_activation_emulators.jl
```diff
@@ -0,0 +1,362 @@
...
+    crs = AM.Mode_κ(r2, σ2, N2, (FT(1.0),), (FT(1.0),), (salt.M,), (salt.κ,), 1)
+    ad = AM.AerosolDistribution((crs, acc))
+
+    # NN
+    # Get the ML model
+    mach_NN = get_2modal_model_FT32(
+        ml_model = "NN",
+        machine_name = "2modal_nn_machine_naive.jls",
+    )
+
```
trontrytel, line 309: Could we encapsulate individual tests with 

```
TT.@testset "0M_microphysics" begin

end
```
blocks? I think this way they display nicer in the test summary in the CI

trontrytel (approved):
Thank you for doing this! My two asks would be to:
- not hardcode the `FT` and parameter structs in `Common.jl`
- add some small documentation page about the available extensions. Apologies I didn't do it with the initial NN extension. I was planning to add it when we have the other guys. I was thinking to add a short section in the Aerosol Activation part of our docs and mention there that we have those additional models. (We can add plots, once we have them for the paper)

trontrytel:
Actually, one more thought. Is there a way to add the `N_activated_per_mode` that we define in `ext` to the `API` docs?

sajjadazimi:
I tried loading the extension in docs to make Julia understand we have this extra function, but it doesn't add it to API. Maybe documenter goes over the specific files of each module to get the documentations and does not do it with the extensions.

PR #363, opened 2024-03-20
Add EKP calibration of ARG to emulators ext

edejong-caltech:
## Purpose 
Adds calibration of the original ARG parameterization using ensemble kalman processes to the existing Emulators extension and testing,

----


ext/Common.jl
```diff
@@ -156,3 +157,18 @@ end
 function inverse_target_transform(transformed_act_frac)
     return @. (1.0 / (2.0 * 0.99)) * tanh(transformed_act_frac) + 0.5
 end
+
+function calibration_error_metrics(X, Y, ensemble, aip, tps, FT)
+    N_ensemble = length(ensemble[1, :])
+    rmse = zeros(N_ensemble)
+    for i in 1:N_ensemble
+        pred = calibrated_prediction(X, ensemble[:, i], aip, tps, FT)
+        rmse[i] = StatsBase.mean(sqrt.((pred .- Y) .^ 2))
```
trontrytel, line 166: If we only need the mean, would it be ok to compute it by ourselves? Instead of adding another dependency?

Project.toml
```diff
@@ -14,17 +14,19 @@ Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
...
 CloudyExt = "Cloudy"
-EmulatorModelsExt = ["DataFrames", "MLJ"]
+EmulatorModelsExt = ["DataFrames", "MLJ", "EnsembleKalmanProcesses"]
 
 [compat]
 ClimaParams = "0.10"
 Cloudy = "0.4"
 DataFrames = "1.6"
 DocStringExtensions = "0.8, 0.9"
+EnsembleKalmanProcesses = "1.1.5"
```
trontrytel, line 29: Would it be enough to set it to "1.1"?

ext/EmulatorModelsExt.jl
```diff
@@ -60,4 +62,25 @@ function AA.N_activated_per_mode(
     end
 end
 
+"""
+    CalibratedAerosolActivationParameters(ekp_params)
+
+    - `ekp_params` - parameters from the trained Ensemble Kalman Process
+Returns a calibrated set of aerosol activation parameters
+"""
+function AA.CalibratedAerosolActivationParameters(
```
trontrytel, line 71: Instead of creating a new function, could we extend this one: `AerosolActivationParameters` ? It's defined in [here](https://github.com/CliMA/CloudMicrophysics.jl/blob/26e29d49ae397b6d1039c78db7a8d88b98f0658c/src/parameters/AerosolActivation.jl#L40C10-L40C37)

test/aerosol_activation_emulators.jl
```diff
@@ -5,10 +5,13 @@ import MLJModels
...
 import GaussianProcesses
 import StatsBase
+import Random
+import EnsembleKalmanProcesses as EKP
+using EnsembleKalmanProcesses.ParameterDistributions
 
-Standardizer = MLJ.@load Standardizer pkg = MLJModels
-NeuralNetworkRegressor = MLJ.@load NeuralNetworkRegressor pkg = MLJFlux
-EvoTreeRegressor = MLJ.@load EvoTreeRegressor pkg = EvoTrees
+# Standardizer = MLJ.@load Standardizer pkg = MLJModels
```
trontrytel, line 12: We should uncomment those

test/aerosol_activation_emulators.jl
```diff
@@ -280,12 +283,116 @@ function get_2modal_model_FT32(;
     end
 end
 
+function calibrate_ARG(
```
trontrytel, line 286: Lets add it as a separate file and try running in the normal CI (so add it to `runtests.jl`). We moved the emulator tests to longruns, because alltogether they were taking more than 30minutes

trontrytel: Or if we want to keep all ML tests together, then you have to manually trigger a buildkite-longruns job. Lmk if you don't know how to do that

Project.toml
edejong-caltech: Yes, thanks!

test/aerosol_activation_emulators.jl
edejong-caltech: Good catch, thank you!

test/aerosol_activation_emulators.jl
edejong-caltech: I was thinking this should stay living with the emulators, but actually I've changed my mind on that now... thanks

ext/EmulatorModelsExt.jl
edejong-caltech: Just kidding, it does work! Yay for multiple dispatch.

Project.toml
trontrytel: Sounds good. Do we need Optim? Or is it a dependence of EKP or something else?

test/aerosol_activation_calibration.jl
```diff
@@ -0,0 +1,179 @@
...
+        push!(all_mean_error_metrics, mean_error_metrics)
+    end
+    final_params_means = StatsBase.mean(all_params_means)
+
+    return final_params_means, all_mean_error_metrics
+end
+
+function test_emulator(
+    FT;
+    ml_model = "NN",
```
trontrytel, line 123: We only test the `calibrated` `ml_model` here. So maybe we don't need this argument and the following `if` statement?

trontrytel (approved):

PR #366, opened 2024-04-08
Linear HOM J option

amylu00:
## Content

- create linear parameterization for homogeneous ice nucleation rate coefficient J dependent on water activity criterion

----


trontrytel (approved):
Thank you!

PR #374, opened 2024-04-17
add JET test_opt to performance tests

trontrytel:
Any other tests like this I should be adding?


charleskawczynski (approved):
Checking that `@test 0 == @allocated foo(args...)` would be good!

trontrytel:
> Checking that `@test 0 == @allocated foo(args...)` would be good!

Is that different than:
```

    TT.@test trail.memory <= min_memory
    TT.@test trail.allocs <= min_allocs
```
(I already have those in the `bench_press` function

charleskawczynski:
I think that's similar/equivalent, I didn't see it. Looks good

PR #378, opened 2024-04-18
new r_eff parametrization and update 2-moment formulas radarrefl and …

crocicc:
…r_eff

## Content

----

### DIFF
--- a/docs/src/Microphysics2M.md
+++ b/docs/src/Microphysics2M.md
@@ -43,14 +43,14 @@ The default value of ``x^*=2.6\times 10^{-10} kg`` corresponds to the drop radiu
 The cloud droplets Gamma distribution function is described by
 ```math
 \begin{align}
-    f_c(x)=A_cx^\nu_c e^{-B_cx},\quad \nu_c=\text{2.0}
+    f_c(x)=A_cx^\nu_c e^{-B_cx},\quad \nu_c=\text{2.0}.
 \end{align}
 ```
 Here,
  - ``A_c=\frac{N_{liq} \, B_c^{\nu_c+1}}{\Gamma(\nu_c+1)}``,
  - ``B_c=\left(\frac{\Gamma(\nu_c+1)}{\Gamma(\nu_c+2)}x_c\right)^{-1}``,
  - ``x_c=\frac{L_c}{N_{liq}}`` - averaged mass of cloud particles,
- - ``L_c = \rho_w q_{liq}`` - cloud liquid water content.
+ - ``L_c = \rho q_{liq}`` - cloud liquid water content.
 
 The raindrops exponential distribution is expressed as
 ```math
@@ -70,7 +70,7 @@ In this case,
  - ``A_r=\frac{\mu_r \, N_{rai}}{\Gamma(\frac{\nu_r+1}{\mu_r})}B_r^{\frac{\nu_r+1}{\mu_r}}``,
  - ``B_r=\left(\frac{\Gamma(\frac{\nu_r+1}{\mu_r})}{\Gamma(\frac{\nu_r+2}{\mu_r})}x_r\right)^{(-\mu_r)}``,
  - ``x_r=\frac{L_r}{N_{rai}}`` - averaged mass of rain particles,
- - ``L_r = \rho_w q_{rai}`` - rain water content.
+ - ``L_r = \rho q_{rai}`` - rain water content.
 
 !!! note
     In the derivation of the parametrization, it is assumed that the cloud droplet distribution ``f_c(x)`` does not contain a significant number of droplets with masses almost equal or larger than ``x^*``. This is reffered to as the undeveloped cloud droplet spectrum assumption. Similarly the raindrop distribution does not contain a significant number of rain drops with masses almost equal or smaller than ``x^*``. These assumptions allow us to simplify the calculation of moments of the distributions by integrating from zero to infinity.
@@ -392,7 +392,7 @@ Z = {\int_0^\infty r^{6} \, n(r) \, dr}.
 \label{eq:Z}
 \end{equation}
 ```
-To take into consideration the effect of both cloud and rain droplets, we integrate separately over the two dstributions defined in equations (2) and (3).
+To take into consideration the effect of both cloud and rain droplets, we integrate separately over the two distributions defined in equations (2) and (3).
 
 For cloud droplets, integrating over the assumed Gamma distribution (eq. 6) leads to
 ```math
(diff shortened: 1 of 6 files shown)


docs/src/Microphysics2M.md
```diff
@@ -408,53 +408,38 @@ Z_r = 3 \, \frac{6! \, A_r}{(B_r)^{7} \, (\frac{4}{3} \, \pi \, \rho_w)^{2}}.
...
-```math
-\begin{equation}
-L_z = \frac{L_{Z_c} N_{liq} + L_{Z_r} N_{rai}}{N_{liq} + N_{rai}}.
-\end{equation}
-```
-
 ### Effective radius
 
-The effective radius of hydrometeors (``r_{eff}``) is the weighted average of their size distribution, and it is defined as the ratio of the third to the second moment of a droplet size distribution (``n(r)``) :
+The effective radius of hydrometeors (``r_{eff}``) is the weighted average of their size distribution, and it is defined as the ratio of the third  to the second moment of a droplet size distribution.
```
trontrytel, line 422:
```suggestion
The effective radius of hydrometeors (``r_{eff}``) is the area weighted radius of the population of particles.  It can be computed as the ratio of the third  to the second moment of the size distribution.
```

docs/src/Microphysics2M.md
```diff
@@ -408,53 +408,38 @@ Z_r = 3 \, \frac{6! \, A_r}{(B_r)^{7} \, (\frac{4}{3} \, \pi \, \rho_w)^{2}}.
...
 
-To obtain the logarithmic radar reflectivity ``L_z``, which is commonly used to refer to the radar reflectivity values, we divide both ``Z_c`` and ``Z_r`` with the equivalent return of a ``1 mm`` drop in a volume of a meter cube (``Z_0``), apply the decimal logarithm to the result, and multiply the result by ``10``. 
-
-For example, for the cloud droplets radar reflectivity we have:
+To obtain the logarithmic radar reflectivity ``L_z``, which is commonly used to refer to the radar reflectivity values, we sum ``Z_c`` and ``Z_r``, and we normalize the result with the equivalent return of a ``1 mm`` drop in a volume of a meter cube (``Z_0``). 
+Then, we apply the decimal logarithm, and multiply the result by ``10``:
 ```math
 \begin{equation}
-L_{Z_c} = 10 \log_{10}(\frac{Z_c}{Z_0}).
+L_z = 10 \log_{10}(\frac{Z_c + Z_r}{Z_0}).
```
trontrytel, line 415:
```suggestion
L_z = 10 \log_{10}\left(\frac{Z_c + Z_r}{Z_0}\right).
```

docs/src/Microphysics2M.md
```diff
@@ -408,53 +408,38 @@ Z_r = 3 \, \frac{6! \, A_r}{(B_r)^{7} \, (\frac{4}{3} \, \pi \, \rho_w)^{2}}.
...
-
 ### Effective radius
 
-The effective radius of hydrometeors (``r_{eff}``) is the weighted average of their size distribution, and it is defined as the ratio of the third to the second moment of a droplet size distribution (``n(r)``) :
+The effective radius of hydrometeors (``r_{eff}``) is the weighted average of their size distribution, and it is defined as the ratio of the third  to the second moment of a droplet size distribution.
+In our case, since we have two separate distributions for cloud and rain, after applying the appropriate transformation to express the distributions as functions of the radius, we obtain:
 ```math
 \begin{equation}
-r_{eff} = \frac{{\int_0^\infty r^{3} \, n(r) \, dr}}{{\int_0^\infty r^{2} \, n(r) \, dr}}.
+r_{eff} = \frac{(M_{c}^3 + M_{r}^3)}{M_{c}^2 + M_{r}^2} = \frac{{\int_0^\infty r^{5} \, (f_c(r) + f_r(r)) \, dr}}{{\int_0^\infty r^{4} \, (f_c(r) + f_r(r)) \, dr}}.
```
trontrytel, line 426: Why r^5 and r^4 and not r^3 and r^2?

trontrytel: Also, I think we were using `n(r)` for the particle size distribution. Could we replace `f(r)` by that?

docs/src/Microphysics2M.md
```diff
@@ -722,6 +707,28 @@ and the default free parameter values are:
 |------------|---------------------------|
 |``A``       | ``4.7``                   |
 
+### Effective radius - Liu and Hallett, 1997
```
trontrytel, line 710: Could we move that to the same section as the effective radius computed from the size distribution?

docs/src/Microphysics2M.md
```diff
@@ -722,6 +707,28 @@ and the default free parameter values are:
...
+  - ``k = 0.8``. 
+
+Where the volume-averaged radius is computed using
+```math
+\begin{equation}
+  r_{vol} = \left(\frac{3}{4 \pi \rho_w}\right)^{\frac{1}{3}} \, \left(\frac{L}{N}\right)^{\frac{1}{3}} = \left(\frac{3  \rho (q_{liq} + q_{rai})}{4 \pi \rho_w (N_{liq} + N_{rai})}\right)^{\frac{1}{3}},
+\end{equation}
+```
+where:
+  - ``L = \rho q_{liq}``, is the liquid water content,
```
trontrytel, line 729:
```suggestion
  - ``L = \rho q``, is the water content,
```
Since we are taking both cloud liquid and rain contributions?

test/microphysics2M_tests.jl
```diff
@@ -487,17 +487,19 @@ function test_microphysics2M(FT)
...
-    TT.@testset "2M_microphysics - Seifert and Beheng 2006 effective radius" begin
+    TT.@testset "2M_microphysics - Seifert and Beheng 2006 radar reflectivity" begin
         #setup 
-        ρ_air = FT(1)
-        ρ_w = FT(1000)
-        q_liq = FT(3.654e-4)
-        N_liq = FT(22886304)
-        q_rai = FT(1.563e-4)
-        N_rai = FT(6813825)
+        # need to pass densities (ρ) in g/cm^3,
```
trontrytel, line 492: Could we stick to passing in variables in base SI units? So meters and kilograms, etc. If we need to convert to different units for better accuracy, lets do it inside the reflectivity and effective radius functions

src/Microphysics2M.jl
```diff
@@ -494,28 +494,27 @@ function radar_reflectivity(
...
-    return ((N_liq + N_rai) == 0) ? -200 :
-           ((Zc * N_liq + Zr * N_rai) / (N_liq + N_rai))
+    return ((Zc + Zr) == 0) ? -200 : 10 * (log10((Zc + Zr) / Z₀) + log10(1e-9))
 end
 
 """ 
     effective_radius(struct, q_liq, q_rai, N_liq, N_rai, ρ_air, ρ_w)
 
     - `struct` - type for 2-moment rain autoconversion parameterization
+                ONLY for the computation using the bulk scheme
```
trontrytel, line 507:
```suggestion
```

src/Microphysics2M.jl
```diff
@@ -545,12 +544,37 @@ function effective_radius(
...
+        (Bc == 0) ? 0 : (Ac * SF.gamma(11 / 3) / (Bc^(11 / 3) * C^(2 / 3)))
+    rain_3moment = (Br == 0) ? 0 : (24 * Ar / (Br^4 * C))
+    rain_2moment = (Br == 0) ? 0 : (6 * Ar / (Br^3 * C^(2 / 3)))
+
+    return ((cloud_2moment + rain_2moment) == 0) ? 0 :
+           ((cloud_3moment + rain_3moment) / (cloud_2moment + rain_2moment)) *
+           1e-3
+end
+
+function effective_radius(
```
trontrytel, line 561: Maybe it would be better to name this function `effective_radius_Liu_Hallet_97` (or something among those lines. This way no-one will use it by accident, thinking they are using the size distribution one.

trontrytel: Then it would also need a docstring. I think just moving the docstring you have in lines 515 is ok

src/Microphysics2M.jl
```diff
@@ -545,12 +544,37 @@ function effective_radius(
     Br =
         (xr == 0) ? 0 :
         ((SF.gamma((νr + 1) / μr) * xr / SF.gamma((νr + 2) / μr))^(-μr))
+    Ac = (N_liq * Bc^(νc + 1)) / SF.gamma(νc + 1)
+    Ar = (μr * N_rai * Br^((νr + 1) / μr)) / SF.gamma((νr + 1) / μr)
+
+    cloud_3moment = (Bc == 0) ? 0 : (6 * Ac * C^3 / (Bc * C)^4)
+    cloud_2moment =
+        (Bc == 0) ? 0 : (Ac * SF.gamma(11 / 3) / (Bc^(11 / 3) * C^(2 / 3)))
```
trontrytel, line 552: The type of `(1/3)` etc will be `Float64`, so we should wrap all of them in `FT()`. 

It's not super important here, because both `Z` and `r_eff` will be used as diagnostics. But just in case

src/Microphysics2M.jl
```diff
@@ -545,12 +544,37 @@ function effective_radius(
     Br =
         (xr == 0) ? 0 :
         ((SF.gamma((νr + 1) / μr) * xr / SF.gamma((νr + 2) / μr))^(-μr))
+    Ac = (N_liq * Bc^(νc + 1)) / SF.gamma(νc + 1)
+    Ar = (μr * N_rai * Br^((νr + 1) / μr)) / SF.gamma((νr + 1) / μr)
+
+    cloud_3moment = (Bc == 0) ? 0 : (6 * Ac * C^3 / (Bc * C)^4)
```
trontrytel, line 550: Maybe the logic would be cleaner to just have one `if/else` per function. 

You could define another helper function that computes the effective radius for cloud and rain. Then that one would return a zero in case no particles are present. And then it would be called twice to compute the whole thing. I think it would be more readable. You could maybe even use dispatch to have one function defined twice for cloud and rain.

trontrytel: In case I'm not explaining properly, I can add that after you are done with the other review comments.

trontrytel (commented):
Thank you for adding all of that! I left some comments. The main ones were:
- to stick to using the base si units in input vars, and doing the needed conversions inside the functions
- I was confused about the `r^5` and `r^4` in the docs

trontrytel:
Also, you will need to squash and rebase. We did merge a couple of PRs today. But I don't expect any conflicts :)

docs/src/Microphysics2M.md
crocicc: It's r^4, and r^5 because when going from the mass distribution (proportional to r^3) to the radius one, I have to multiply by the derivative of the mass wrt r (= 4 * pi * r^2 * \rho)

src/Microphysics2M.jl
crocicc: I am not sure I understood what you'd want to compute exactly in that helper function (since the computation of the moments is different for cloud and rain). And should I do it for both Radar reflectivity and effective radius?

For now I updated the code following the other comments, but feel free to tell me how to create those functions (or to do it by yourself if you prefer)!

trontrytel (approved):

PR #379, opened 2024-04-19
Clean dependencies, reorder runtests

nefrathenrici:
Content
- Clean dependencies
- Reorder runtests so that slow tests run last. It's quicker to test when developing if performance tests run last


trontrytel (approved):
Thank you!

PR #381, opened 2024-04-23
Add P3 Sink Terms

anastasia-popova:
## Purpose 

This PR aims to add the necessary collision sink terms to the P3 Scheme

## Content

----

### DIFF
--- a/docs/src/P3Scheme.md
+++ b/docs/src/P3Scheme.md
@@ -13,10 +13,15 @@ This simplification aids in attempts to constrain the scheme's free parameters.
 
 The prognostic variables are:
  - ``N_{ice}`` - number concentration 1/m3
- - ``q_{ice}`` - ice mass density kg/m3
- - ``q_{rim}`` - rime mass density kg/m3
+ - ``Q_{ice}`` - ice mass density kg/m3
+ - ``Q_{rim}`` - rime mass density kg/m3
  - ``B_{rim}`` - rime volume - (volume of rime per total air volume: dimensionless)
 
+where 
+ - ``Q_{ice} = \rho_a * q_{ice}`` 
+ - ``Q_{rim} = \rho_a * q_{rim}``
+ - ``\rho_a`` - density of air 
+ - ``q_{ice}``, ``q_{rim}`` are respective specific humidities
 !!! note
     TODO - At some point we should switch to specific humidities...
 
@@ -30,10 +35,10 @@ The mass ``m`` and projected area ``A`` of particles
 | particle properties                  |      condition(s)                            |    m(D) relation                             |    A(D) relation                                           |
 |:-------------------------------------|:---------------------------------------------|:---------------------------------------------|:-----------------------------------------------------------|
 |small, spherical ice                  | ``D < D_{th}``                               | ``\frac{\pi}{6} \rho_i \ D^3``               | ``\frac{\pi}{4} D^2``                                      |
-|large, unrimed ice                    | ``q_{rim} = 0`` and ``D > D_{th}``           | ``\alpha_{va} \ D^{\beta_{va}}``             | ``\gamma \ D^{\sigma}``                                    |
-|dense nonspherical ice                | ``q_{rim} > 0`` and ``D_{gr} > D > D_{th}``  | ``\alpha_{va} \ D^{\beta_{va}}``             | ``\gamma \ D^{\sigma}``                                    |
-|graupel (completely rimed, spherical) | ``q_{rim} > 0`` and ``D_{cr} > D > D_{gr}``  | ``\frac{\pi}{6} \rho_g \ D^3``               | ``\frac{\pi}{4} D^2``                                      |
-|partially rimed ice                   | ``q_{rim} > 0`` and ``D > D_{cr}``           | ``\frac{\alpha_{va}}{1-F_r} D^{\beta_{va}}`` | ``F_{r} \frac{\pi}{4} D^2 + (1-F_{r})\gamma \ D^{\sigma}`` |
+|large, unrimed ice                    | ``Q_{rim} = 0`` and ``D > D_{th}``           | ``\alpha_{va} \ D^{\beta_{va}}``             | ``\gamma \ D^{\sigma}``                                    |
+|dense nonspherical ice                | ``Q_{rim} > 0`` and ``D_{gr} > D > D_{th}``  | ``\alpha_{va} \ D^{\beta_{va}}``             | ``\gamma \ D^{\sigma}``                                    |
+|graupel (completely rimed, spherical) | ``Q_{rim} > 0`` and ``D_{cr} > D > D_{gr}``  | ``\frac{\pi}{6} \rho_g \ D^3``               | ``\frac{\pi}{4} D^2``                                      |
+|partially rimed ice                   | ``Q_{rim} > 0`` and ``D > D_{cr}``           | ``\frac{\alpha_{va}}{1-F_r} D^{\beta_{va}}`` | ``F_{r} \frac{\pi}{4} D^2 + (1-F_{r})\gamma \ D^{\sigma}`` |
 
 where:
  - ``D_{th}``, ``D_{gr}``, ``D_{cr}`` are particle size thresholds in ``m``,
@@ -55,8 +60,8 @@ The remaining thresholds: ``D_{gr}``, ``D_{cr}``, as well as the
  - ``\rho_g = \rho_r F_r + (1 - F_r) \rho_d``
  - ``\rho_d = \frac{6\alpha_{va}(D_{cr}^{\beta{va} \ - 2} - D_{gr}^{\beta{va} \ - 2})}{\pi \ (\beta_{va} \ - 2)(D_{cr}-D_{gr})}``
 where
- - ``F_r = \frac{q_{rim}}{q_{ice}}`` is the rime mass fraction,
- - ``\rho_{r} = \frac{q_{rim}}{B_{rim}}`` is the predicted rime density.
+ - ``F_r = \frac{Q_{rim}}{Q_{ice}}`` is the rime mass fraction,
+ - ``\rho_{r} = \frac{Q_{rim}}{B_{rim}}`` is the predicted rime density.
 
 !!! note
     TODO - Check units, see in [issue #151](https://github.com/CliMA/CloudMicrophysics.jl/issues/151)
@@ -94,7 +99,7 @@ The model predicted ice number concentration and ice mass density are defined as
 N_{ice} = \int_{0}^{\infty} \! N'(D) \mathrm{d}D
 ```
 ```math
-q_{ice} = \int_{0}^{\infty} \! m(D) N'(D) \mathrm{d}D
+Q_{ice} = \int_{0}^{\infty} \! m(D) N'(D) \mathrm{d}D
 ```
 
 ## Calculating shape parameters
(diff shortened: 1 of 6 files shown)


trontrytel:
I think we need to make the integrals more complicated. This is because we want to use the 2M Seifert and Beheng scheme for the cloud and rain part. That means we should integrate overt the assumed 2M distributions for cloud and rain (instead of assuming continuous q). And for precipitation we need to consider the fact that rain has a terminal velocity of its own (see for example eq 25 in 1M docs. 

I wrote down some notes, trying to keep the notation the same as in SB and P3 scheme docs pages:

PXL_20240424_230008608

trontrytel:
In the 1M scheme we take the velocity difference out of the integral by approximating it by the mass weighted velocity difference. Thats one way of making this simpler. The alternative is to swicth to numerical integrals, and keep the velocities in the integral

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
 
+"""
+    get_rain_parameters(q, N) 
+
+ - q - mass mixing ratio of rain
+ - N - number mixing ratio of rain 
+
+ returns the parameters λ and N_0 where N' = N_0 * exp(-λ  * D)
+"""
+function get_rain_parameters(q::FT, N::FT) where {FT}
```
trontrytel, line 778: As discussed, could we re-use the functions (or add them if needed) in the 2-moment microphysics scheme? The same for the `get_cloud_parameters`

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+    N′_rain(D, q, N)
+
+ - D - diameter of particle 
+ - q - mass mixing ratio
+ - N - number mixing ratio 
+
+ Returns the distribution of rain particles (assumed to be of the form 
+ N'(D) = N0 * exp(-λD)) at given D 
+"""
+function N′rain(D::FT, q::FT, N::FT) where {FT}
```
trontrytel, line 811: Could we define this and `N_cloud` in the 2-moment microphysics scheme?

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+
+"""
+    ϕ_ice(p3, D, F_r, th) 
+
+ - p3 - a struct containing P3 Scheme parameters 
+ - D - maximum particle dimension
+ - F_r - rime mass fraction (q_rim/q_i)
+ - th - P3Scheme nonlinear solve output tuple (D_cr, D_gr, ρ_g, ρ_d)
+
+ Returns the ϕ calculation for an ice particle in the P3 Scheme with given
```
trontrytel, line 854: What is `ϕ` ?

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+
+    v = 0
+    for i in 1:3
+        v += ai[i] * D^(bi[i]) * exp(-ci[i] * D)
+    end
+
+    return v
+end
+
+function get_ice_bound(p3, λ::FT, tolerance::FT) where {FT}
```
trontrytel, line 928: Could you add some docstring explaining the logic how you are getting the bounds? I'm wondering if that could be also tied to the limited vars assumptions from the 2-moment scheme (or something along those lines)

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+    Nᵢ::FT,
+    q::FT,
+    N::FT,
+) where {FT}
+    if type == "rain"
+        rain_λ, = get_rain_parameters(q, N)
+        colliding_x = -1 / rain_λ * log(tolerance)
+    elseif type == "cloud"
+        cloud_λ, = get_cloud_parameters(q, N)
+        cloud_problem(x) =
```
trontrytel, line 973: The same for th bounds for the `cloud_probelm`. Could you explain how you get those bounds? Could the logic for them be combined between the cloud rain and ice?

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+    a(D1, D2) 
+
+ - D1 - maximum dimension of first particle 
+ - D2 - maximum dimension of second particle 
+
+ Returns the collision kernel (assumed to be of the form π(r1 + r2)^2) for the 
+ two colliding particles
+"""
+function a(D1::FT, D2::FT) where {FT}
+    # TODO make this more accurate for non-spherical particles
```
trontrytel, line 1018: Do we have any ideas what to assume for the other particle regimes?

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+    T::FT,
+    E_ci = FT(1),
+) where {FT}
+    ρ_w = FT(1000)      # TO DO: add this as density_liquid_water from ClimaParams
+    th = thresholds(p3, ρ_r, F_r)
+    (λ, N_0) = distribution_parameter_solver(p3, qᵢ, Nᵢ, ρ_r, F_r)
+    (colliding_bound, ice_bound) =
+        integration_bounds(type, p3, eps(FT), λ, N_0, Nᵢ, q_c, N_c)
+
+    if T > 273.15
```
trontrytel, line 1115: This should be `T_freeze` from ClimaParams. And agreed that `rho_w` should be taken from ClimaParams also. We can add them to the `p3` parameters and then just grab them from the struct here. Since we are already passing in the `p3` parameters struct. Lmk if you need help with that

src/P3Scheme.jl
```diff
@@ -667,4 +767,464 @@ function D_m(p3::PSP3, q::FT, N::FT, ρ_r::FT, F_r::FT) where {FT}
...
+    q::FT,
+    N::FT,
+    T::FT,
+    ρ_a::FT,
+    F_r::FT,
+    ρ_r::FT,
+) where {FT}
+    # Get constants
+    (; ν_air, D_vapor, K_therm) = aps
+    a = FT(0.78)
```
trontrytel, line 1203: Could you remind me wehre are this values coming from? Should we also add them to the `p3` parameters struct?

anastasia-popova: These values are coming from the A. Seifert and K. D. Beheng. paper that is cited in the 1M scheme ventilation factor section. The P3 Scheme only says that they use an approach similar to a paper which just gives an equation that we are not using due to the differentiation in snow, graupel, and ice within that paper. I am not sure whether these are the values we want to use but the P3 paper does not give an answer to this question.

src/P3Scheme.jl
anastasia-popova: Personally, no I do not. I was thinking finding some D' value for the regimes where the area is not in the form of D^2 such that D'^2 = D^sigma and then using D' in this calculation but I am not convinced that this would give us anything helpful.

src/P3Scheme.jl
anastasia-popova: Should this maybe be moved to the same place where we calculate terminal velocities and have a passed in mass and area? This is the only place where phi is not 1 but for future use maybe it would be more helpful there?

docs/src/P3Scheme.md
```diff
@@ -188,3 +188,92 @@ They can be compared with Figure 2 from [MorrisonMilbrandt2015](@cite).
 include("plots/P3TerminalVelocityPlots.jl")
 ```
 ![](MorrisonandMilbrandtFig2.svg)
+
+## Collisions
+
+Collisions are calculated through the following equation:
```
trontrytel, line 194:
```suggestion
Collisions are calculated through the following equation (see the 1-moment accretion documentation for more discussioon):
```

docs/src/P3Scheme.md
```diff
@@ -188,3 +188,92 @@ They can be compared with Figure 2 from [MorrisonMilbrandt2015](@cite).
 include("plots/P3TerminalVelocityPlots.jl")
 ```
 ![](MorrisonandMilbrandtFig2.svg)
+
+## Collisions
+
+Collisions are calculated through the following equation: 
+
+```math 
+\frac{dq_c}{dt} = \int_{0}^{\infty} \! \int_{0}^{\infty} \! \frac{E_ci}{\rho} N'_i(D_p) N'_c(D_c) A(D_i, D_c) m_{collected}(D_{collected}) |V_i(D_i) - V_c(D_c)| \mathrm{d}D_i \mathrm{d}D_c
```
trontrytel, line 197:
```suggestion
\frac{dq_c}{dt} = \int_{0}^{\infty} \! \int_{0}^{\infty} \! \frac{E_{ci}}{\rho} \; N'_i(D_i) \; N'_c(D_c) \;  A(D_i, D_c) \; m_{cld}(D_{cld}) \; |V_i(D_i) - V_c(D_c)| \; \mathrm{d}D_i \mathrm{d}D_c
```

docs/src/P3Scheme.md
```diff
@@ -188,3 +188,92 @@ They can be compared with Figure 2 from [MorrisonMilbrandt2015](@cite).
...
+  - subscript ``c`` - corresponding colliding species
+  - subscript ``collected`` - whichever species is being collected by the collisions 
+  - ``E_ci`` - collision efficiency between particles
+  - ``N'_x`` - distribution of particles of x type
+  - ``A`` - collision kernel between ice and colliding species 
+  - ``m_x(D)`` - mass of given species x at diameter D
+  - ``V_x(D)`` - Chen terminal velocity of given species x at diameter D 
+
+  In our parametrization we take ``E_ci = 1`` and ``A(D_i, D_c) = \pi \frac{(D_i + D_c)^2}{4}``. We also assume that if the temperature is above freezing then ice particles get collected, otherwise the colliding species is collected by the ice.  
+
```
trontrytel, line 210: Maybe something like this:

```
where:
  - ``i`` - denotes ice particles (either cloud ice or precipitating ice),
  - ``c`` - denotes colliding species (cloud liquid water on rain),
  - ``cld`` - denotes the species that are being collected during the collision event,
    (We assume that if the temperature is above freezing then ice particles get collected.
    In temperatures below freezing, cloud and rain particles are collected by the ice.),
  - ``E_ci`` - is the collision efficiency,
  - ``N'_x`` - is the particle distribution,
  - ``A`` - is the assumed crossection of the two colliding species,
  - ``m_x(D)`` - is the assumed particle mass as a function of it's size,
  - ``V_x(D)`` - is the assumed particle terminal velocity as a function of it's size.

!!! note
    TODO - In our parametrization we take ``E_ci = 1``

!!! note
    TODO - For now we assume ``A(D_i, D_c) = \pi \frac{(D_i + D_c)^2}{4}``.
    This is true for spherical particles only, we should be changing this according to the
    size regimes.

In order to compute the above integral,
  we need to know the particle size distributions of cloud and rain particles.
P3 scheme will be used with Seifert Beheng 2-moment warm rain microphysics
  [scheme](https://clima.github.io/CloudMicrophysics.jl/dev/Microphysics2M/#The-Seifert-and-Beheng-(2006)-parametrization).
  which assumes a Gamma distribution of cloud droplet mass and exponential distribution of rain drop mass.
Writing them as functions of droplet maximum dimension (meaning the diameter for spherical drops) we get:
```

docs/src/P3Scheme.md
```diff
@@ -188,3 +188,92 @@ They can be compared with Figure 2 from [MorrisonMilbrandt2015](@cite).
...
+include("plots/P3TendenciesSandbox.jl")
+```
+![](CollisionComparisons.svg)
+
+## Melting 
+
+Melting rates are calculated through the following equation: 
+
+```math 
+\frac{dq}{dt} = \frac{1}{2 \rho_a} \int_{0}^{\infty} \! \frac{dm(D)}{dt} N'(D) \mathrm{d}D
```
trontrytel, line 232: I think it should be

```suggestion
\frac{dq}{dt} = \int_{0}^{\infty} \! \frac{dm(D)}{dt} N'(D) \mathrm{d}D
```
because in the P3 scheme `q` is in `kg/m3` we don't need to divide by air density. Maybe it would be good to write the P3 variables as `rho_air q`. I think this way the notation bewteen P3 and the 1M and 2M schemes would actually be consistent and we could remove one of the todos

trontrytel:
I can help revive this branch after #403 is merged. Lets see if we can make this work with limiters, or if we have to do without

PR #382, opened 2024-04-24
Bump version for a new release

trontrytel:
I don't think. it's a breaking release. But there was a lot of development from the last one. So just to be on the safe side I'm bumping it to `0.19`


nefrathenrici (approved):
I can add a NEWS.md so we can track breaking changes more clearly

trontrytel:
> I can add a NEWS.md so we can track breaking changes more clearly

Sounds great!

PR #384, opened 2024-04-25
Add Cloudy.jl parcel example to docs

edejong-caltech:
## Content

----


trontrytel:
I think it wouldn't be hard to actually have one parcel model and add Cloudy option as another condensation option to dispatch over in `Parcel.Tendencies.jl` As far as I can tell, the only clunky bit would be that we would have to extend the state vector by the number of moments you are using in the example. But on the plus side we wouldn't have to repeat the whole parcel model implementation and the plotting related to reproducing the figure from Rogers.

Lmk if thats something you want to try to do. If not, I'm fine to merge it as is and just open an issue to remember that this could be refactored.

Remember to squash and rebase before merging. And thank you fr adding it!

trontrytel (approved):

trontrytel:
And also, could you add a sentence to the [news](https://github.com/CliMA/CloudMicrophysics.jl/blob/main/NEWS.md) about what is being added here?

PR #385, opened 2024-04-26
AIDA ice nucleation calibrations

amylu00:
## Purpose 

Calibrate water activity based ice nucleation parameterizations to AIDA chamber data.

## To-do

-  error analysis functions to calibration.jl
- separate file which overwrites/calibrates the parameters for each nucleation mode
- figure out how to ensure calibrated parameters aren't in the wrong minimum/dip. i.e. loss function should lead us to global min and not the local min
- documentation?
- call data from Caltech box(? one drive? ClimaArtifacts) instead of dropbox
- probably tag Ollie (EKP) and Gabriele (ClimaArtifacts) for review too

## Content

----


parcel/ParcelModel.jl
```diff
@@ -139,7 +140,7 @@ function parcel_model(dY, Y, p, t)
         a1 * w * Sₗ - (a2 + a3) * Sₗ * dqₗ_dt_v2l -
         (a2 + a4) * Sₗ * dqᵢ_dt_v2i - a5 * Sₗ * dqᵢ_dt_l2i
 
-    dp_air_dt = -p_air * grav / R_air / T * w
+    dp_air_dt = -p_air * grav / R_air / T * w #* 3.5
```
amylu00, line 143:
```suggestion
    dp_air_dt = -p_air * grav / R_air / T * w
```

amylu00:
There are quite a few hard-coded things in the aida calibrations file (i.e. start and end times). Was planning to fix them in a separate PR where I also calibrate the rest of the HOM experiments, but let me know if I should fix that in this PR or its own PR. Also, not sure if we want to plot the before-calibration parcel time evolutions or if that's too messy (for now it is commented out).

PR #391, opened 2024-05-02
add sensitivity checks for r_eff and radarrefl 2M

crocicc:
## Content

----

### DIFF
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -448,7 +448,7 @@ function rain_evaporation(
 end
 
 """ 
-    radar_reflectivity(struct, q_liq, q_rai, N_liq, N_rai, ρ_air, ρ_w)
+    radar_reflectivity(struct, q_liq, q_rai, N_liq, N_rai, ρ_air, ρ_w, τ_q, τ_N)
 
     - `struct` - type for 2-moment rain autoconversion parameterization
     - `q_liq` - cloud water specific humidity 
@@ -457,6 +457,8 @@ end
     - `N_rai` - rain droplet number density 
     - `ρ_air` - air density
     - `ρ_w` - water density
+    - `τ_q` - threshold for minimum specific humidity value
+    - `τ_N` - threshold for minimum number density value
 
 Returns logarithmic radar reflectivity from the assumed cloud and rain particle 
 size distribuions normalized by the reflectivty of 1 millimiter drop in a volume of
@@ -470,6 +472,8 @@ function radar_reflectivity(
     N_rai::FT,
     ρ_air::FT,
     ρ_w::FT,
+    τ_q::FT,
+    τ_N::FT,
 ) where {FT}
 
     # we assume a cloud droplets gamma distribution,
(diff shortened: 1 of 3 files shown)


src/Microphysics2M.jl
```diff
@@ -488,8 +488,11 @@ function radar_reflectivity(
     N_liq *= FT(1e-9)
     N_rai *= FT(1e-9)
 
-    xc = (N_liq == 0) ? FT(0) : ((q_liq * ρ_air) / N_liq)
-    xr = (N_rai == 0) ? FT(0) : ((q_rai * ρ_air) / N_rai)
+    q_liq = (q_liq < maximum(q_liq) * 1e-8) ? FT(0) : q_liq
+    q_rai = (q_rai < maximum(q_rai) * 1e-8) ? FT(0) : q_rai
+
+    xc = (N_liq < 1e-4) ? FT(0) : ((q_liq * ρ_air) / N_liq)
```
trontrytel, line 494: I'm wondering if with those additional checks for q and N not being too small, we could get rid of the unit conversions and keep the whole computation in base SI?

src/Microphysics2M.jl
```diff
@@ -488,8 +488,11 @@ function radar_reflectivity(
     N_liq *= FT(1e-9)
     N_rai *= FT(1e-9)
 
-    xc = (N_liq == 0) ? FT(0) : ((q_liq * ρ_air) / N_liq)
-    xr = (N_rai == 0) ? FT(0) : ((q_rai * ρ_air) / N_rai)
+    q_liq = (q_liq < maximum(q_liq) * 1e-8) ? FT(0) : q_liq
```
trontrytel, line 491: `q_liq` is just a number here, so I think that `maximum(q_liq)` is not really doing anything here

trontrytel: Maybe it would be good to make the cutoffs for N and q the arguments of the Z and r_eff? Those hard-coded limits sometimes actually matter. So it would be good to be able to test that the thresholds we are choosing don't affect the output

src/Microphysics2M.jl
crocicc: I analyzed the function and this is what I think. As we know, the problem lies in the computation of x; specifically, when using the SI units, q and ρ are very small, and N is very large, so x can become very large. This causes a blow up for B, and a subsequent infinite value for A. 
To solve this, I was able to remove the change of unit for the density, but I think that the change of units for N (from m^-3 to mm^-3) is needed. 
I specified it with a comment above it. Let me know what you think!

src/Microphysics2M.jl
crocicc: I agree that it's better to put them as arguments. It allows us to always check that the threshold do not alter the results; and considering that these two products are very sensitive to small modifications of their arguments I think it makes a lot more sense like this.

crocicc: I made the corrections and I also just squashed and rebased. Check that I did it correctly (but I think so), and lmk if the corrections look good to you! :)

trontrytel (approved):

trontrytel:
Thank you!

PR #402, opened 2024-05-23
add check r_eff and Z

crocicc:
## Content

----


trontrytel:
Do we still need anything from here? @crocicc

crocicc:
Nope @trontrytel ! I delete it!

PR #403, opened 2024-05-24
Case without limiters

crocicc:
## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #404, opened 2024-05-24
Issue with rain distributions not returning the values of mass when integrated

anastasia-popova:
## Purpose 

We noticed a difference in distributions from the one that the 2M scheme solves vs the one that the P3 scheme solves for. This can be seen using the integral functions I added to the Microphysics 2M file. 

mass_integral() integrated the SB2006 distribution in terms of mass over all mass values 
diameter_integral() converts the SB2006 distribution to be in terms of diameter and then integrates over all diameter values 
p3_integral() takes the distribution that is solved for in the P3 Scheme (analogous to the process in the "Assumed Particle Distribution" section of the P3 Docs (assuming that rain has an exponential distribution i.e., mu = 0)). 

When run without the mass term, all of the distributions return the total N values.
However, when run with the mass, only the P3 distribution returns the total q value.

specifically with the following values we get these outputs: 
q = 5e-4
N = 1e8
mass_integral() = (0.053242761992414823, 2.1347496010890966e-10)
diameter_integral() = (0.05396701648992669, 5.426008450982034e-10)
p3_integral() =  (0.0004999999999990584, 5.0271524832576724e-12)

Plugging in directly the values for A and lambda from the SB2006 paper instead of using the limited_vars, we get the following: 
mass_integral() = (0.00025670012836447303, 1.728586069347634e-12)
This is still not the correct value but it is not off by a factor of 2 instead of a factor of 100. 

## Content

----

### DIFF
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -80,6 +80,169 @@ function raindrops_limited_vars(
     return (; λr, xr, Ar, Br)
 end
 
+
+
+"""
+    get_override_pdf_r()
+
+Returns the parameters for pdf_r given the override file 
+"""
+function get_override_pdf_r(FT)
+    # Seifert and Beheng 2006 parameters
+    override_file = joinpath(
+        pkgdir(CM),
+        "src",
+        "parameters",
+        "toml",
+        "SB2006_limiters.toml",
+    )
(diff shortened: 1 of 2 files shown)


trontrytel:
Should be fixed in #403

PR #407, opened 2024-05-30
1M radarrefl adjustment

crocicc:
## Purpose 
1-moment Adjustment of radar reflectivity. 
I simply change units for maintaining the accuracy, and set a lowest possible value -300; with this threshold we make sure to avoid obtaining -Inf values.

## Content

----

### DIFF
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -91,7 +91,7 @@ function lambda(
     # mass(size)
     (; r0, m0, me, Δm, χm) = mass
 
-    return q > FT(0) ?
+    return q > eps(FT) ?
            (
         χm * m0 * n0 * SF.gamma(me + Δm + FT(1)) / ρ / q / r0^(me + Δm)
     )^FT(1 / (me + Δm + 1)) : FT(0)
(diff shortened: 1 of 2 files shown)


src/Microphysics1M.jl
```diff
@@ -113,11 +113,15 @@ function radar_reflectivity(
...
 
-    n0 = get_n0(pdf)
-    λ = lambda(pdf, mass, q, ρ)
-    Z₀ = FT(1e-18)
+    # change units for accuracy
+    n0 = get_n0(pdf) * FT(1e-12)
+    λ = lambda(pdf, mass, q, ρ) * FT(1e-3)
 
-    return 10 * log10((720 * n0 / λ^7) / Z₀)
+    Z = (λ == FT(0)) ? FT(0) : (720 * n0 / λ^7)
```
sajjadazimi, line 120: `== FT(0)` is not safe for float numbers. You should do `lambda < eps(FT)` instead

src/Microphysics1M.jl
```diff
@@ -113,11 +113,15 @@ function radar_reflectivity(
...
+    log_10_Z₀ = FT(-18)
+    log_Z = FT(10) * (log10(Z) - log_10_Z₀ - FT(9))
+
+    return (log_Z < FT(-300)) ? FT(-300) : log_Z
```
trontrytel, line 124: Do we have to clip it here to -300? Would it be ok to just output very small numbers and deal with it when postprocessing the output.

Also `return max(FT(-300), log_z)` is maybe shorter?

trontrytel (approved):

src/Microphysics1M.jl
crocicc: It needs a clipping because it could have -Inf values. We just have to choose how low we want the clipping value! + I'll change to `return max(FT(-300), log_z)`

sajjadazimi (approved):

src/Microphysics1M.jl
sajjadazimi: You can check the minimum meaningful number that log10 returns and set the threshold to that.

PR #409, opened 2024-06-06
correction pdf_rain 2Moment

crocicc:
## Content

----

### DIFF
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -138,9 +138,10 @@ function pdf_rain(
     N_rai::FT,
 ) where {FT}
     (; νr, μr, xr_min, xr_max, N0_min, N0_max, λ_min, λ_max, ρw) = pdf_r
+    ϵ_N = FT(10)
 
     L_rai = ρ * q_rai
-    xr_0 = L_rai / N_rai
+    xr_0 = N_rai == 0 ? FT(0) : L_rai / N_rai
     xr_hat = max(xr_min, min(xr_max, xr_0))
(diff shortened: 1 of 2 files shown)


src/Microphysics2M.jl
```diff
@@ -334,10 +339,9 @@ function rain_self_collection(
     (; ρ0, ρw) = pdf
 
     L_rai = ρ * q_rai
-    #λr =
-    #    pdf_rain(pdf, q_rai, ρ, N_rai).λr *
-    #    (SF.gamma(FT(4)) / FT(π) / ρw)^FT(1 / 3)
-    λr = pdf_rain(pdf, q_rai, ρ, N_rai).Br
+    λr =
+        pdf_rain(pdf, q_rai, ρ, N_rai).λr *
+        (SF.gamma(FT(4)) / FT(π) / ρw)^FT(1 / 3)
```
trontrytel, line 344: Why do we need to multiply lambda here?

crocicc: I went back to the version of computation of self collection that we had before our last merge. I had to do it because now we were taking `.Br` instead of not `.λr` from the `pdf_rain()`; however, when I apply the correction for avoiding the NaNs, in this process you need the limited lambda, but in the radar reflectivity and effective radius I need `Ar` and `Br` equal to 0. So, I came back to the version of computation of the self collection that we had before the last merge (and that didn't use Br), so that I can have Br and Ar equal to 0 for radar_refl and effective radius, but at the same time I don't affect this process.

When I didn't do this, running in KiD led to instabilities, so I believe that this is one of the cases where we need the limited values for lambda.

trontrytel:
We should close this PR, if https://github.com/CliMA/CloudMicrophysics.jl/pull/411 gets merged.

PR #411, opened 2024-06-12
More cleanup in limiters

trontrytel:
@crocicc - could you take a look when you have the time. The tests work for the numbers you gave me. 

Also I double checked that the `Br` coefficient that we are returning from `pdf_rain` is correct. So we should use that one in self collection for rain, instead of computing it again.


crocicc:
In the end, wrt to Anna's code, I just added the check for negative values of qr. I did so because making other changes generated numerical instabilities with KiD, and it also made the scheme less coherent in the case with limiters. 

Let me know if that check looks good to you, so that we can proceed with the merge! 
Thank you!

crocicc (approved):

PR #414, opened 2024-06-25
Lognormal droplets in parcel

amylu00:
Adding lognormal droplet distribution option in parcel. Note that the ice distribution will remain as a **gamma** distribution.

### DIFF
--- a/parcel/Example_Immersion_Freezing.jl
+++ b/parcel/Example_Immersion_Freezing.jl
@@ -64,7 +64,8 @@ for DSD in size_distribution_list
         heterogeneous = heterogeneous,
         condensation_growth = condensation_growth,
         deposition_growth = deposition_growth,
-        size_distribution = DSD,
+        liq_size_distribution = DSD,
+        ice_size_distribution = DSD,
     )
     # solve ODE
     local sol = run_parcel(IC, FT(0), t_max, params)
--- a/parcel/ParcelDistributions.jl
+++ b/parcel/ParcelDistributions.jl
@@ -4,29 +4,28 @@ struct Monodisperse{FT} <: CMP.ParametersType{FT} end
 
 struct Gamma{FT} <: CMP.ParametersType{FT} end
 
-# Size distributiom moments
-function distribution_moments(::Monodisperse, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρᵢ)
-    FT = typeof(qₗ)
-    # liquid droplet
-    if Nₗ == FT(0)
-        rₗ = FT(0)
-        Aₗ = FT(0)
-        Vₗ = FT(0)
-    else
-        rₗ = cbrt(qₗ / Nₗ / FT(4 / 3 * π) / ρₗ * ρ_air)
-        Aₗ = 4 * FT(π) * rₗ^2
-        Vₗ = FT(4 / 3 * π) * rₗ^3
-    end
-    # ice crystals
-    if Nᵢ == FT(0)
-        rᵢ = FT(0)
-    else
-        rᵢ = cbrt(qᵢ / Nᵢ / FT(4 / 3 * π) / ρᵢ * ρ_air)
-    end
-    return (; rₗ, Aₗ, Vₗ, rᵢ)
+struct Lognormal{FT} <: CMP.ParametersType{FT}
+    σ_g::FT
 end
 
-function distribution_moments(::Gamma, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρᵢ)
+#! format: off
+# Size distributiom moments
+function distribution_moments(::Monodisperse, q, N, ρ, ρ_air)
+    FT = typeof(q)
+    if N == FT(0) || q == FT(0)
+        r = FT(0)
+        A = FT(0)
+        V = FT(0)
+    else
+        r = cbrt(q / N / FT(4 / 3 * π) / ρ * ρ_air)
+        A = 4 * FT(π) * r^2
+        V = FT(4 / 3 * π) * r^3
+    end
+    return (; r, A, V)
+end
+
+function distribution_moments(::Gamma, q, N, ρ, ρ_air)
+    FT = typeof(q)
     #integral(n(r)) = 1/λ^2
     #integral(r*n(r)) = 2/λ^3
     #<r> = 2/λ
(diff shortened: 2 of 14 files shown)


parcel/ParcelTendencies.jl
```diff
@@ -190,9 +190,10 @@ end
 function condensation(params::CondParams, PSD, state, ρ_air)
     FT = eltype(state)
     (; aps, tps) = params
-    (; Sₗ, T, Nₗ) = state
+    (; Sₗ, T, Nₗ, qₗ) = state
     Gₗ = CMO.G_func(aps, tps, T, TD.Liquid())
-    return 4 * FT(π) / ρ_air * (Sₗ - 1) * Gₗ * PSD.rₗ * Nₗ
+    return Nₗ > FT(0) && qₗ > FT(0) ?
```
trontrytel, line 195: Is this check needed? Maybe it would be better to ensure that `PSD.rl` is zero when q and N are zero? I think we are actually already doing it for `N` but maybe not for `q` for all the supported distributions...

This way it would also work for the deposition growth

parcel/ParcelModel.jl
```diff
@@ -195,8 +198,13 @@ function run_parcel(IC, t_0, t_end, pp)
     info = "\nSize distribution: $(pp.size_distribution)\n"
     if pp.size_distribution == "Monodisperse"
         distr = Monodisperse{FT}()
+        distr_params = nothing
     elseif pp.size_distribution == "Gamma"
         distr = Gamma{FT}()
+        distr_params = nothing
+    elseif pp.size_distribution == "Lognormal"
+        distr = Lognormal{FT}()
```
trontrytel, line 206: Could we instead make the sigma a parameter of `distr`? See  for example like `MohlerAF` struct has four fields that contain parameters (or any other parameter struct that we create)

parcel/ParcelDistributions.jl
```diff
@@ -54,3 +57,27 @@ function distribution_moments(::Gamma, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρ
...
+        rₗ = FT(0)
+        Aₗ = FT(0)
+        Vₗ = FT(0)
+    else
+        r_median = cbrt(3 / 4 / FT(π) / Nₗ * qₗ / ρₗ * ρ_air / exp(9 / 2 * (log(σ_g))^2))
+        rₗ = r_median * exp(0.5 * (log(σ_g))^2)
+        Aₗ = 4 * FT(π) * r_median^2 * exp(2 * (log(σ_g))^2)
+        Vₗ = 4 / 3 * FT(π) * r_median^3 * exp(9 / 2 * (log(σ_g))^2)
+    end
+    # ice crystals; note this is gamma distribution
```
trontrytel, line 74: This is definitely confusing. If we want to have different distributions for both liquid and ice and we should split the `distribution_moments` function to return things for one phase only. And then call it twice in the code for liquid and ice. And make sure to throw an error if someone tries to have Lognormal distribution for ice

parcel/ParcelDistributions.jl
```diff
@@ -54,3 +57,27 @@ function distribution_moments(::Gamma, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρ
...
+
+function distribution_moments(::Lognormal, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρᵢ, distr_params)
+    (; σ_g) = distr_params
+    # liquid droplets
+    if Nₗ == FT(0)
+        rₗ = FT(0)
+        Aₗ = FT(0)
+        Vₗ = FT(0)
+    else
+        r_median = cbrt(3 / 4 / FT(π) / Nₗ * qₗ / ρₗ * ρ_air / exp(9 / 2 * (log(σ_g))^2))
```
trontrytel, line 69: Do you have your derivation notes on how you got those formulae?

amylu00: Yes, should I add it to the documentation somewhere?

trontrytel: Sure! But I just asked, because I was getting something slightly different and I wanted to double check

amylu00:
Adding `&& qₗ == FT(0)` in the distribution moments creates instability in Frostenberg example. Without it, the value of `qₗ` is to the order of `1e-7` or `1e-6`. Not sure if we want to approximate that to be zero and use an `isapprox` statement for the q's. We would need a an absolute tolerance, not a relative tolerance, because we're working with 0 but then the issue becomes "what do we use as our arbitrary tolerance to say that q is approximately zero".

In an earlier commit, I tried to catch any cases where N was 0 but q was non-zero and vice versa because that situation would physically not make sense. I used `isapprox` statements then but the absolute tolerances seemed so arbritary and several examples started getting errors so I backtracked and kept it simple by just adding `&& qₗ == FT(0)`. Maybe it is necessary though?

parcel/ParcelDistributions.jl
```diff
@@ -4,11 +4,17 @@ struct Monodisperse{FT} <: CMP.ParametersType{FT} end
...
+
+#! format: off
 # Size distributiom moments
-function distribution_moments(::Monodisperse, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρᵢ)
+function distribution_moments(::Monodisperse, ::TD.Liquid, qₗ, Nₗ, ρₗ, ρ_air)
     FT = typeof(qₗ)
     # liquid droplet
-    if Nₗ == FT(0)
+    if Nₗ == FT(0) || qₗ == FT(0)
+        sleep(1)
```
amylu00, line 17: get rid of this

parcel/Example_Immersion_Freezing.jl
```diff
@@ -22,6 +22,7 @@ p₀ = FT(800 * 1e2)
 T₀ = FT(251)
 qᵥ = FT(8.1e-4)
 qₗ = Nₗ * 4 / 3 * FT(π) * r₀^3 * ρₗ / FT(1.2) # 1.2 should be ρₐ
+@info(" ", qₗ)
```
trontrytel, line 25: Also leftover from debugging

parcel/ParcelDistributions.jl
trontrytel: LGTM! Thank you

parcel/ParcelDistributions.jl
```diff
@@ -17,16 +23,22 @@ function distribution_moments(::Monodisperse, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, N
...
         rᵢ = FT(0)
     else
         rᵢ = cbrt(qᵢ / Nᵢ / FT(4 / 3 * π) / ρᵢ * ρ_air)
     end
-    return (; rₗ, Aₗ, Vₗ, rᵢ)
+    return (; rᵢ)
 end
 
-function distribution_moments(::Gamma, qₗ, Nₗ, ρₗ, ρ_air, qᵢ, Nᵢ, ρᵢ)
+function distribution_moments(::Gamma, ::TD.Liquid, qₗ, Nₗ, ρₗ, ρ_air)
```
trontrytel, line 40: We are passing in liquid and ice densities as input. So I don't think we need to have a separate second argument for the condensate type. `function distribution_moments(::Gamma, q, N, ρ, ρ_air)` should work for both liquid and ice. And I think the same for other distributions. And then we would just need to call it twice in parcel models to specify separate distributions for liquid droplets and ice crystals

trontrytel (commented):
I left a couple more small comments. I think it's almost ready to be merged. `

PR #415, opened 2024-06-30
Implement a modified terminal velocity as an option in SB2006

sajjadazimi:
## Purpose 
- Implement a modified terminal velocity as an option in SB2006
- Fix abnormal plots (Chen velocity) in documentation

<!---
Review checklist

I have:


src/Microphysics2M.jl
```diff
@@ -421,7 +421,7 @@ Fall velocity of individual rain drops is parameterized:
 """
 function rain_terminal_velocity(
     (; pdf_r)::CMP.SB2006{FT},
-    (; ρ0, aR, bR, cR)::CMP.SB2006VelType{FT},
+    (; ρ0, aR, bR, cR, modified)::CMP.SB2006VelType{FT},
```
trontrytel, line 424: I'm wondering, does it even make sense to keep the `modified=false` as option? It seems to me like we should always be computing it this way. - The code would be simpler and without 4 `if/else` statements there. 

If not, it would be better to define two different versions of this function and dispatch based on the velocity type. This way the choice would happen at the compile time and not via `if/else` during run time.

docs/src/plots/TerminalVelocityComparisons.jl
```diff
@@ -218,23 +232,39 @@ Ch_snow = [snow_terminal_velocity_individual_Chen(snow, Chen2022.snow_ice, ρ_ai
...
+p2 = PL.plot(q_rain_range * 1e3,  SB_rain_bN, linewidth = 3, xlabel = "q_rain [g/kg]", ylabel = "terminal velocity [m/s]", label = "Rain-SB2006 [ND]", linestyle = :dot, color = 1)
+p2 = PL.plot!(q_rain_range * 1e3, Ch_rain_bN, linewidth = 3, xlabel = "q_rain [g/kg]", ylabel = "terminal velocity [m/s]", label = "Rain-Chen2022 [ND]", linestyle = :dot, color = 2)
+p2 = PL.plot!(q_rain_range * 1e3, SB_rain_bM, linewidth = 3, xlabel = "q_rain [g/kg]", ylabel = "terminal velocity [m/s]", label = "Rain-SB2006 [M]", color = 1)
+p2 = PL.plot!(q_rain_range * 1e3, Ch_rain_bM, linewidth = 3, xlabel = "q_rain [g/kg]", ylabel = "terminal velocity [m/s]", label = "Rain-Chen2022 [M]", color = 2)
+p2 = PL.plot!(q_rain_range * 1e3, M1_rain_bM, linewidth = 3, xlabel = "q_rain [g/kg]", ylabel = "terminal velocity [m/s]", label = "Rain-default-1M [M]", color = 3)
+# SB2006 group velocities vs. mean radius
+p3 = PL.scatter(r_mean, SB_rain_bN_rm, ms = 1.0, markerstrokewidth=0, xlabel = "r_mean [μm]", ylabel = "terminal velocity [m/s]", label = "Rain-SB2006 [ND]")
```
trontrytel, line 258: I don't understand why the bottom two panels `Rain-SB2006` (blue color) is a cloud of points and the other two are lines?

src/Microphysics2M.jl
sajjadazimi: The reason I defined this flag is because I thought maybe Caterina is using the original SB2006. So I kept the original as a default. But if not I'm also happy to change the function to the new form.
@crocicc Do you want to keep the original sedimentation equation?

docs/src/plots/TerminalVelocityComparisons.jl
sajjadazimi: For SB2006 with limiters, terminal velocity is a function of both q and N while for the others it is only a function of mean radius. So in the plot we see multiple terminal velocity values for one value of mean radius for the equation with limiters while there is only one value for other cases.

docs/src/Microphysics2M.md
```diff
@@ -539,10 +551,10 @@ $\lambda$ is the size distribution parameter.
...
 and mass-weighted mean terminal velocity when k = 3.
 
-Below, we compare the individual terminal velocity formulas for [Chen2022](@cite) and [SeifertBeheng2006](@cite).
-We also compare bulk number weighted [ND] and mass weighted [MD]
-terminal velocities for both formulas integrated over the size distribution from [SeifertBeheng2006](@cite).
-We also show the mass weighted terminal velocity from the 1-moment scheme.
+Below, in the top-left panel we compare the individual terminal velocity formulas for Chen2022 [Chen2022](@cite), SB2006 [SeifertBeheng2006](@cite) and data from [Gunn1949](@cite).
+In the top-right panel, we compare bulk number weighted [ND] and mass weighted [M]
+terminal velocities for both formulas integrated over the size distribution from SB2006 [SeifertBeheng2006](@cite).
+We also show the mass weighted terminal velocity from the 1-moment scheme. In the bottom panels we compare number-weighted (left) and mass-weighted (right) terminal velocities for the original parameterization of SB2006 [SeifertBeheng2006](@cite) and the modifed parameterzation given by eq. (\ref{eq:sb_modifiedterminalvelocity}) without limiting distribution shape factor $\lambda_r$.
```
trontrytel, line 557: Something doesn't get parsed right here in the docs preview.

docs/src/plots/TerminalVelocityComparisons.jl
trontrytel: This is interesting! Thank you, I guess I wasn't aware of that. All that scatter is only present for the small drops, right? After we get out of the region where the limiters are important, the two options will be quite similar right?

If yes, could we either change the x axis on the bottom panels to extend further till about 2 mm? Or highlight in the description that we are zooming into the region of interest and that outside of that they are the same?

Thank you for catching and fixing all of this!

trontrytel:
I forgot to add, could you also include a short description in NEWS.md file? Apologies for that, I wish it was more automatic

trontrytel (approved):
Thank you!

docs/src/plots/TerminalVelocityComparisons.jl
sajjadazimi: The limiters are important for large droplets too. For medium size droplets when both N and lambda are in the defined range then the two velocities are similar. I extended x axis to 2 mm and now the effect of limiters is more clear on the figure.

PR #418, opened 2024-07-02
Try fixing 1M scheme for ClimaAtmos

trontrytel:
ClimaAtmos 1-moment simulations don't work with the new release of CloudMicrophysics (v0.21). I tracked it down to us changing the "do nothing" condition in 1-moment microphysics scheme from `FT(0)` to `eps(FT)`.

This PR reverts the changes. I will make a new release for CloudMicrophysics after it is merged (and ideally also #417 and #415 )

### DIFF
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -91,20 +91,20 @@ function lambda(
     # mass(size)
     (; r0, m0, me, Δm, χm) = mass
 
-    return q > eps(FT) ?
+    return q > FT(0) ?
            (
         χm * m0 * n0 * SF.gamma(me + Δm + FT(1)) / ρ / q / r0^(me + Δm)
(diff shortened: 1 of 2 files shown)


trontrytel:
cc @szy21 - This should fix Atmos simulations

szy21:
Thanks! It's interesting that this makes a difference to stability. Maybe eps(Float32) is too large as a threshold for things like q_rai?

sajjadazimi (approved):

crocicc (approved):
Looks good to me Anna! Considering that eps(Float32) is quite large, it will even make the computation of radar reflectivity more accurate; so thank you so much for this!

trontrytel:
Agreed. I think for Float32 its not a good condition for our water tracers. One way around it would be to consider not using base SI units for some of the microphysics variables. But for now I'll make sure that the schemes we use in Atmos use `FT(0)`

PR #420, opened 2024-07-04
Bugfix SB2006 evaporation; return 0 instead of NaN for xr = 0

sajjadazimi:
## Purpose 
SB2006 evaporation scheme used to return NaN for xr = 0 when limiters are not applied. The reason is that in the computation of number density rate we have a 0/0 term. The limit of this equation for XR -> 0 is zero. This PR adds a check in the code to catch cases where xr/x_star is small and return 0 instead of NaN.

I have:

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ main
 ------
 <!--- # Add changes since the most recent release here --->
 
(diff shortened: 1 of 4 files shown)


crocicc (approved):
Hi Sajjad! I already tested my KiD simulations with this fix and I don't have any instability anymore -in the case without limiters- for my setup. Thank you so much for finding and fixing the bug!

trontrytel (approved):
Great catch! Thank you.

PR #425, opened 2024-07-23
Add P3 terminal velocity functions

trontrytel:
This PR:
- Adds the rain and ice individual particle terminal velocities needed for P3 scheme numerical integrals. A peel off from https://github.com/CliMA/CloudMicrophysics.jl/pull/381.
- Fixes a typo in one of the velocity difference computations.
- Removes the `TerminalVelocity` file and adds the necessary functions in the Microphysics2M (for rain) or P3_terminal_velocity (for ice) files.


rorlija1 (approved):
I'm not sure if there are rules as to what to comment on in a PR review: But I appreciate the TODOs you added, and I noticed that since the velocity difference function will be used only when we merge the processes, it should be good to merge it even though it's not covered by the tests. Looks great!

PR #426, opened 2024-07-25
Simplify the P3 terminal velocity

trontrytel:
- Don't consider aspect ratio
- Use QuadGK for mass and number weighted terminal velocity integrals

### DIFF
--- a/docs/src/plots/P3TerminalVelocityPlots.jl
+++ b/docs/src/plots/P3TerminalVelocityPlots.jl
@@ -91,17 +91,17 @@ function figure_2()
     args = (color = :black, labels = true, levels = 3, linewidth = 1.5, labelsize = 18)
 
     ax1 = make_axis_top(fig, 1, "Small Dₘ")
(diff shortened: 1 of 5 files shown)


trontrytel:
JET shows some new instabilities from QuadGK. I'll try to debug tomorrow, but I think it's ok to merge if everything else passes and you see no mistakes!

rorlija1 (approved):
It sucks that Anastasia did that work only for us to be using quadgk for all of it now... but the velocity is quite simple now! Also, although it's probably good to merge now, it'll be interesting to continue to investigate whether we should be using get_ice_bound() or double the value it returns as the upper bound of integration.

trontrytel:
> It sucks that Anastasia did that work only for us to be using quadgk for all of it now... but the velocity is quite simple now! Also, although it's probably good to merge now, it'll be interesting to continue to investigate whether we should be using get_ice_bound() or double the value it returns as the upper bound of integration.

I hope to be able to go back to the semi-analytical solutions at some point. But since we discovered that the numerical integrals reproduce the results better I wanted to start with those. And then work on replacing/reviving what we had there while making sure the results are the same

PR #428, opened 2024-07-27
update notation q -> L

trontrytel:
The original paper uses symbol ``q`` to denote the mass of a tracer per volume of air (named mass mixing ratio). In our documentation of the 1-moment and 2-moment schemes we used ``q`` to denote the mass of a tracer per mass of air (named specific humidity). To keep the notation consistent between the 1,2-moment schemes and P3, and to highlight the difference between normalizing by air mass or volume, we denote the mass of a tracer per volume of air as ``L`` (named content).


rorlija1 (approved):
Feels like a big change even though it's only a small notation change! :)

PR #429, opened 2024-07-29
Aerosol activation in parcel

amylu00:
Adding simple aerosol activation in parcel. Takes in one aerosol type at a time (no mixing) and one mode of aerosols only. Should try to replace S_max with the parcel S_l, but will do so in separate PR?


trontrytel:
I'm not sure if thats the way to use the ARG2000 scheme in parcel. - ARG2000 parameterizes the maximum supersaturation ever achieved by an adiabatic parcel and then uses it to compute the number of activated aerosol particles at the cloud base (not throughout the whole cloud). Maybe if we want to stay close to the assumptions of the derivation we would only apply that tendency once when we first reach supersaturation? Not sure how stable numerically it would be, but probably doable in parcel.

I think a much better approach is to use the supersaturation we are solving for in parcel. And follow the ARG2000 math to compute the number of activated aerosol particles at each time step.

trontrytel (changes requested):
I left some longer comment. I would vote to either
- keep ARG2000 as a cloud base only activation scheme
- rewrite it to use our supersaturation as input

amylu00:
> I think a much better approach is to use the supersaturation we are solving for in parcel. And follow the ARG2000 math to compute the number of activated aerosol particles at each time step.

I agree. I've already implemented using supersat from the parcel in a local branch and it's stable! I'll update that in this branch this week

trontrytel:
> > I think a much better approach is to use the supersaturation we are solving for in parcel. And follow the ARG2000 math to compute the number of activated aerosol particles at each time step.
> 
> I agree. I've already implemented using supersat from the parcel in a local branch and it's stable! I'll update that in this branch this week

Thats awesome! Thank you.

parcel/ParcelModel.jl
```diff
@@ -112,16 +123,17 @@ function parcel_model(dY, Y, p, t)
...
     # ... water mass budget
+    dqₗ_dt_a2l = dqₗ_dt_act
     dqₗ_dt_v2l = dqₗ_dt_ce
     dqᵢ_dt_l2i = dqᵢ_dt_imm + dqᵢ_dt_hom
     dqᵢ_dt_v2i = dqᵢ_dt_dep + dqᵢ_dt_ds
 
     # Update the tendecies
     dqᵢ_dt = dqᵢ_dt_v2i + dqᵢ_dt_l2i
-    dqₗ_dt = dqₗ_dt_v2l - dqᵢ_dt_l2i
+    dqₗ_dt = dqₗ_dt_v2l - dqᵢ_dt_l2i + dqₗ_dt_a2l
```
trontrytel, line 136: Should there be a corresponding sink to water vapor as well?

parcel/ParcelModel.jl
```diff
@@ -112,16 +123,17 @@ function parcel_model(dY, Y, p, t)
 
     # number concentration and ...
     dNᵢ_dt = dNᵢ_dt_dep + dNᵢ_dt_imm + dNᵢ_dt_hom
-    dNₐ_dt = -dNᵢ_dt_dep
-    dNₗ_dt = -dNᵢ_dt_imm - dNᵢ_dt_hom
+    dNₐ_dt = -dNᵢ_dt_dep - dNₗ_dt_act
```
trontrytel, line 126: It might become an issue that we are decreasing the number concentration of aerosol particles but not changing the standard deviation of the mode

parcel/Example_AerosolActivation.jl
```diff
@@ -0,0 +1,74 @@
...
+qᵢ = FT(0)
+Sₗ = FT(1.2)
+e = Sₗ * eₛ
+p₀ = e / cᵥ₀
+IC = [Sₗ, p₀, T₀, qᵥ, qₗ, qᵢ, Nₐ, Nₗ, Nᵢ, ln_INPC]
+
+# Simulation parameters passed into ODE solver
+w = FT(1)           # updraft speed
+const_dt = FT(1)    # model timestep
+t_max = FT(30)      # total time
```
trontrytel, line 45: Would it be possible to start below supersaturation and run the simulation long enough to see the supersaturation reach its maximum value and then start decreasing (the same way as you would expect in the cloud going from cloud base aerosol activation inside the cloud conditions)

trontrytel (approved):
Left some small comments - mostly unsure about the water budget. I think there is a sink term missing for water vapor. Also, it would be nice to run the example for a little longer, to test if we are getting the expected behavior.

Thank you for adding it!

parcel/ParcelModel.jl
amylu00: Because we are neglecting the condensation onto the aerosol, the `dqₗ_dt_a2l` term only accounts for the aerosol and not any of the water vapor. So technically what it's doing there is converting a prescribed volume of soluble aerosol into a volume of liquid water. A corresponding sink would be a decrease in "q_a" but since mass of aerosols is not tracked, it's not in the parcel anywhere. I could add it though if that's something we want?

parcel/ParcelModel.jl
amylu00: I'll make a note of that in documentation but maybe we can assume -- for the sake of simple parcel simulations for now -- that the aerosols are all activating with the same probability and that way the shape of the PDF doesn't change? Or is that a terrible assumption?

parcel/ParcelModel.jl
trontrytel: In that case, shouldn't it be multiplied by the aerosol density and not liquid water?

`dqₗ_dt_act = dNₗ_dt_act * 4 * FT(π) / 3 * r_nuc^3 * ρₗ / ρ_air`

Maybe it would be better to not mix the aerosol mass and water mass. We could assume some sort of initial cloud droplet size after activation, and compute the corresponding sink of water vapor based on that

parcel/ParcelModel.jl
```diff
@@ -112,17 +123,18 @@ function parcel_model(dY, Y, p, t)
...
+    dqₗ_dt_a2l = dqₗ_dt_act
+    dqₗ_dt_v2l = dqₗ_dt_ce + dqₗ_dt_act
     dqᵢ_dt_l2i = dqᵢ_dt_imm + dqᵢ_dt_hom
     dqᵢ_dt_v2i = dqᵢ_dt_dep + dqᵢ_dt_ds
 
     # Update the tendecies
     dqᵢ_dt = dqᵢ_dt_v2i + dqᵢ_dt_l2i
-    dqₗ_dt = dqₗ_dt_v2l - dqᵢ_dt_l2i
-    dqᵥ_dt = -dqᵢ_dt - dqₗ_dt
+    dqₗ_dt = dqₗ_dt_v2l - dqᵢ_dt_l2i + dqₗ_dt_a2l
```
trontrytel, line 136: I think you are now counting `dqₗ_dt_a2l` twice? You already added it to `dqₗ_dt_v2l`?

Issue #406, opened 2024-05-28
P3 Liquid Fraction


rorlija1:
**I'll populate this with my notes over the next couple days: just creating an outline now.**

### Sources/sinks

Changes to old sources/sinks:

New sources/sinks:

### Slight modification to mass, area, velocity regimes

```[tasklist]
### Tasks
- [ ] add methods for mass, area, and velocity regimes when `F_{i,liq} != 0`
- [ ] replicate Fig.1? (terminal velocity parameterization)
- [ ] ...
```

PR #431, opened 2024-08-06
P3 Add F_liq to Velocity

rorlija1:
See corresponding [issue](https://github.com/CliMA/CloudMicrophysics.jl/issues/406) and other PRs — this one adds F_liq to terminal velocity with updated docs, with updated L notation, with updated F_rim notation, and with updated tests which test against Cholette et al 2019 Fig1.


trontrytel:
closing and moved to #440

PR #432, opened 2024-08-06
Aspect Ratio Capability in P3 Velocities

rorlija1:
- This PR adds the capability to compute P3 velocities with aspect ratio using a Boolean argument in the terminal velocity functions. Since this choice between using and not using aspect ratio will likely not be maintained in the long term, we can temporarily sacrifice elegance and efficiency in using the Boolean.
- The docs are also modified/expanded here for the aspect ratio (and mascot).
- Since the aspect ratio is property regime-dependent, we add a helper function to compute the ice density in given particle regimes.


trontrytel:
Thank you so much for all the documentation and really digging into understanding the different P3 regimes. Also, I love love love the P3 duck linking to the original repository. 

I modified a couple of small things in the PR:
- made the aspect ratio return zero if `D` is zero (instead of the velocity). I thought maybe it was cleaner this way
- wrapped all the `pi` into FT. I think by default it is `Float64`. 
- added some tests for the velocity with aspect ratio
- deleted commented out code (but I saved it locally in case you want to go back to it)

One bigger issue - based on the aspect ratio we are getting and Fig 2 from Chen, I think we should be assuming our particles are oblate. Which I think means that our kappa power should be 1/3. Notice that I don't have the minus sign in front of it, which I think is a typo in Chen 2022. With that formulation my terminal velocities with aspect ratio are always smaller than when assuming spherical particles. Which in turn agrees with Fig 2 in Chen. 

I will squash/rebase and merge everything. Things to do next for us would be to add some notes on how we derive the aspect ratio and cleanup the documentation plots (sometimes fonts are too small and we have too many plots showcasing the behavior of the numerics of P3 scheme solvers). Maybe we can split the work and I'll do the docs cleanup and you could do the aspect ratio derivation?

trontrytel:
I forgot - in quadgk integrals you can set the relative tolerance with which you integrate. By default its `sqrt(eps(FT))` which is very small for `Float64`. Thats why the functions were so expensive in the performance tests. I think it should be a parameter of the scheme and we should test what value is a good balance between the performance and accuracy, but for now I hardcoded it as relative tolerance of 1e-6.

rorlija1:
Sounds great, thanks for taking the time to look everything over and provide feedback! Yeah, with kappa = 1/3 the terminal velocity plots are definitely different, so I'm curious to see how all of this will affect what we'll see (hopefully) with liquid fraction! I'm not 100% confident on the aspect ratio derivation since I think Anastasia had implemented the general formula, but I'm happy to dig into it. I don't think I can access Chen 2022 for free, so maybe if you could send me that paper I'd appreciate it since it may help with the aspect ratio stuff.

PR #433, opened 2024-08-07
non-equilibrium parcel example

oalcabes:
## Purpose 

Non-equilibrium initial formulation for parcel model

## Content

----


trontrytel:
Lets split this PR into two, if thats ok:
- first one that adds the time integrators version to our source code, docs and tests
- second one that adds an example to parcel

PR #434, opened 2024-08-09
Cleanup in 2M size distributions

trontrytel:
I noticed while doing integrals for the P3 scheme rates that the integral bound for cloud size distribution in SB2006 was returned in `mm`. I think it's ok to keep it in `m`, even though in some other places (aka effective radius and reflectivity calculations) we switch things to `ug` or `mm` for better accuracy. Keeping it in `m` will make it easier for P3 scheme computations.

### DIFF
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -191,9 +191,12 @@ function size_distribution(
     N::FT,
 ) where {FT}
     (; Cc, Ec, ϕc, ψc) = pdf_cloud_parameters(pdf, q, ρₐ, N)
-    # Convert values from mm to m
-    Ec *= 1e9
-    Cc *= 1e36
+    # Convert values from mm to m assuming
+    Ec *= FT(10^(3 * ψc))
+    # TODO - overflow for Float32. We only need it in P3 scheme and it only
+    # works for Float64 right now. The solution is to non-dimensionalize
+    # all distributions by a reference mass/size, similar to the 1M scheme.
+    Cc *= Float64(10^((ϕc + 4) * 3))
     return Cc * D^ϕc * exp(-Ec * D^ψc)
 end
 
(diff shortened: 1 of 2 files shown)


trontrytel:
Also, what I really care about in the tests are the integrals with bounds computed by the `get_size_distribution_bound` function. If you prefer, I can delete the other ones to simplify the tests

src/Microphysics2M.jl
```diff
@@ -206,7 +206,7 @@ end
  - ρₐ - density of air
  - tolerance - tolerance for integration error
 
-    Returns D_max value such that (1 - tolerance) = 1/N * ∫ N'(D) dD from 0 to D_max.
+    Returns D_max value in meters such that (1 - tolerance) = 1/N * ∫ N'(D) dD from 0 to D_max.
```
sajjadazimi, line 209: It would be good to say D_max value is in meters if inputs are in SI.

test/microphysics2M_tests.jl
```diff
@@ -729,27 +734,32 @@ function test_microphysics2M(FT)
...
+        qₗ = FT(1e-3)      # 1 g/kg
 
         # air and liquid water densities in μg/m3
         ρₐ = FT(1.2)
         ρₗ = SB2006.pdf_r.ρw
 
-        # distribution parameters for cloud, units: [Bc] = 1/μg, [Ac] = 1/m3 1/μg3
+        # distribution parameters for cloud
+        # units: [Bc] = 1/μg, [Ac] = 1/m3 1/μg3
+        # units: [Ec] = 1/mm3, [Cc ] = (1/mm3)^4
```
sajjadazimi, line 762: Based on line 774 (below), it seems like unit of Ec must be be 1/mm^\psi_c and unit of Cc also needs to be a function of \phi_c

sajjadazimi (approved):
Thank you for the additions. It looks good to me. I just left two small comments.

PR #435, opened 2024-08-09
Add P3 immersion freezing

trontrytel:
Add ABIFM immersion freezing ice nucleation to P3 scheme

### DIFF
--- a/src/P3_processes.jl
+++ b/src/P3_processes.jl
@@ -1,1 +1,47 @@
-# TODO
+"""
+    het_ice_nucleation(pdf_c, p3, tps, q, N, T, ρₐ, p, aerosol)
+
+ - aerosol - aerosol parameters (supported types: desert dust, illite, kaolinite)
+ - tps - thermodynamics parameters
+ - qₚ - phase partition
+ - N_liq - cloud water number concentration
+ - RH - relative humidity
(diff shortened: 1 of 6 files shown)


src/P3_processes.jl
```diff
@@ -1 +1,56 @@
...
+    dt::FT,
+) where {FT}
+    #TODO - Also consider rain freezing
+
+    # Immersion freezing nucleation rate coefficient
+    J = CM_HetIce.ABIFM_J(aerosol, RH - CO.a_w_ice(tps, T))
+
+    # TODO - should we make it a parameter of ABIFM scheme?
+    # or a function of the droplet size distribution?
+    A_aer = FT(1e-11) # Assumed erosol surface area
```
amylu00, line 37:
```suggestion
    A_aer = FT(1e-11) # Assumed aerosol surface area
```

amylu00 (approved):
I think for desert dust, we could go with 1e-10 to up to 5e-10 m^2 (eyeballing figure 2 [here](https://acp.copernicus.org/articles/21/3491/2021/)). Not sure what a typical surface area would be for airborne illite but for now maybe it's safe to keep it constant for all aerosol types until we figure out a better representation

PR #438, opened 2024-08-12
P3 aspect ratio derivation

rorlija1:
- Added a short explanation of how aspect ratio is derived to the docs page.


trontrytel:
Thank you!

PR #439, opened 2024-08-13
Add a new nonequilibrium cloud formation

trontrytel:
Part 1 from #433 . Adds non-equilibrium formulation, tests and documentation. 

TODO - add a note in docs about the difference between the current evaporation treatment and the added one (i.e. `cp` vs `D/K`...)


docs/src/API.md
oalcabes, line 1: this needs to be changed from MG2008 to MM2015 -- the 2008 paper is only for warm rain whereas the 2015 paper expands the formulation to mixed phase clouds.

trontrytel: done!

PR #440, opened 2024-08-13
P3 velocities with liquid fraction

rorlija1:
This PR:
- changes notation from F_r to F_rim
- adds liquid fraction to P3 code, docs, tests for use with terminal velocity.
See corresponding [issue](https://github.com/CliMA/CloudMicrophysics.jl/issues/406).


rorlija1:
Seems like there are conflicts now, but I think it's just because of the heterogeneous freezing PR that was merged. I'll try to resolve conflicts in github — I don't think that'll mess anything up!

src/P3_size_distribution.jl
```diff
@@ -96,11 +96,13 @@ function get_ice_bound(p3::PSP3, λ::FT, tolerance::FT) where {FT}
...
 
 """
-    L_(p3, ρ, F_r, λ, μ, D_min, D_max)
+    L_(p3, ρ, F_rim, F_liq, λ, μ, D_min, D_max)
 
  - p3 - a struct with P3 scheme parameters
  - ρ - bulk ice density (ρ_i for small ice, ρ_g for graupel) [kg/m^3]
- - F_r - rime mass fraction [L_rim/L_ice]
+ - F_rim - rime mass fraction (L_rim / (L_ice - L_liq)) [-]
+ - F_liq - liquid fraction (L_liq / L_ice) [-]:
```
trontrytel, line 104: It seems to me that the notation is a little confusing between `L_ice` and `L_liq`. Maybe it would be better to have `L_tot = L_ice + L_liq`?

trontrytel: Or `L_p3_tot`? I'm thinking out loud here, not sure.

src/P3_terminal_velocity.jl
```diff
@@ -12,22 +12,28 @@ following Morrison and Milbrandt (2015). The density of nonspherical
...
-
-    mᵢ = p3_mass(p3, D, F_r, th)
-    aᵢ = p3_area(p3, D, F_r, th)
-    ρᵢ = p3_density(p3, D, F_r, th)
+function ϕᵢ(p3::PSP3, D::FT, F_rim::FT, th) where {FT}
+    # Let F_liq = 0 to calculate ϕᵢ for an ice particle with D
+    # of the total particle but treating a mixed-phase particle
+    # as an ice particle with the same D. We need this for the F_liq
+    # weighted average of velocities discussed in the docs.
+    # TODO? should the above comment be here or should it be in the docs?
```
trontrytel, line 20: We can have it in both. In the docstring for this function we can have:

```
Returns the aspect ratio (ϕ) for an ice particle with mass, cross-section area,
and ice density determined using the size-dependent particle property regimes
following Morrison and Milbrandt (2015). The density of nonspherical
particles is assumed to be equal to the particle mass divided by the volume of a
spherical particle with the same D_max.
Assuming zero liquid fraction.
```

src/P3_terminal_velocity.jl
```diff
@@ -51,24 +57,66 @@ function ice_particle_terminal_velocity(
...
+        p3,
+        D,
+        Chen2022.snow_ice,
+        ρₐ,
+        F_rim,
+        th,
+        use_aspect_ratio,
+    )
+    v_r = CM2.rain_particle_terminal_velocity(D, Chen2022.rain, ρₐ)
+    v = (1 - F_liq) * v_i + F_liq * v_r
```
trontrytel, line 100: This `F_liq` weighing shows up in many places through the source code. I was wondering if it would be cleaner to add some sort of helper function to do this and then just call it throughout the code when needed.

test/p3_tests.jl
```diff
@@ -104,33 +106,73 @@ function test_p3_thresholds(FT)
...
+        TT.@test P3.p3_mass(p3, D_2, F_rim, F_liq, th) == P3.mass_nl(p3, D_2)
+
+        # test F_liq != 0
+        F_liq = FT(0.5)
+
+        # test area
+        TT.@test P3.p3_area(p3, D_1, F_rim, F_liq, th) == P3.A_s(D_1)
+        TT.@test P3.p3_area(p3, D_2, F_rim, F_liq, th) ==
+                 (1 - F_liq) * P3.A_ns(p3, D_2) + F_liq * P3.A_s(D_2)
+        # TODO - debug this test
```
trontrytel, line 146: Is it still breaking?

test/p3_tests.jl
```diff
@@ -104,33 +106,73 @@ function test_p3_thresholds(FT)
...
+        TT.@test P3.p3_mass(p3, D_1, F_rim, F_liq, th) ==
+                 (1 - F_liq) * P3.mass_s(D_1, p3.ρ_i) +
+                 F_liq * P3.mass_s(D_1, p3.ρ_l)
+        TT.@test P3.p3_mass(p3, D_2, F_rim, F_liq, th) ==
+                 (1 - F_liq) * P3.mass_nl(p3, D_2) +
+                 F_liq * P3.mass_s(D_2, p3.ρ_l)
+        TT.@test P3.p3_mass(p3, D_3, F_rim, F_liq, th) ==
+                 (1 - F_liq) * P3.mass_nl(p3, D_3) +
+                 F_liq * P3.mass_s(D_3, p3.ρ_l)
+        # TODO - debug this test:
```
trontrytel, line 163: And here

test/p3_tests.jl
```diff
@@ -567,16 +772,17 @@ end
...
+# TODO - only works for Float64 now. We should switch the units inside the solver
 # from SI base to something more managable
-#test_p3_shape_solver(Float32)
+# test_p3_shape_solver(Float32)
 # test_bulk_terminal_velocities(Float32)
 
 println("Testing Float64")
 test_p3_thresholds(Float64)
-test_p3_shape_solver(Float64)
+# test_p3_shape_solver(Float64)
```
trontrytel, line 783: Are those tests still failing?

src/P3_size_distribution.jl
rorlija1: Ah yeah, maybe we can finalize the best way this afternoon. I think `L_tot = L_ice + L_liq` is what I was going to do originally, but if we make `L_tot = L_ice + L_liq`, when we switch into a model like KiD, I was worried that would be confusing when we're talking about `q_tot = q_ice + q_liq + q_rai + ...` where `q_liq` is the cloud liquid mass and not the mass of liquid on ice particles like P3's `L_liq` is. In the paper they have `L_ice,tot = L_ice + L_liq` but that seemed like too much, so I went with `L_ice = L_rim + L_dep + L_liq`. `L_p3_tot` could be a good solution because that would notation would transfer well to models where we have `L_tot = L_p3_tot + L_other`. I'll change to `L_tot_p3` this morning since I like the way that sounds more than `L_p3_tot` 😄, and we can check in again this afternoon.

src/P3_terminal_velocity.jl
rorlija1: Ooh, you're totally right, that would be nice—I'll get on that with my changes

test/p3_tests.jl
rorlija1: These ones I commented out because of the `get_ice_bound()` quick fix, but the other ones I'll check and have an answer ready for our meeting

trontrytel (approved):
LGTM after all the tests are added back. Thank you!

PR #446, opened 2024-08-21
Add ice melting rate to P3

trontrytel:
Add P3 ice melting parameterization.


rorlija1 (commented):
Aside from performance tests, to the best of my limited knowledge things look good. Maybe the melting rate is high since we're still returning D = 0.01 from get_ice_bound? That's my best guess without digging into the nitty gritty of the parameterization! Maybe if we go ahead with things now, after your gamma approximation ice bound fix is done, we can check to see if it has changed a bit?

PR #449, opened 2024-08-24
P3 melting + F_liq

rorlija1:
This PR modifies melting for liquid fraction and adds a P3Rates struct similar to the one in Microphysics2M.

See the docs build for further explanations, and here are some notes:

- I have a doubt in the code that I'd like to highlight: physically, should we be using the ice core fall speed or the whole (mixed-phase) particle fall speed to calculate the melting rates?
- Also, a small TODO before or after merging: the codecov is missing one line because the tests don't cover the case when the integration bound for the PSD is below D_th, and I can't write tests for that line since `get_ice_bound()` returns a constant. Also, the rest of the tests are failing because I changed the thresholds solver to allow rime density of 0 but did not change the tests to match.
- One larger issue I'm thinking about with the code is I'm wondering if we can use the same function called `ice_melt` to also compute (re)freezing of ``L_liq``. In Cholette (2019), the freezing rate (eq. A5 p. 578) is almost identical to the melting rate, except for a sign change and using the latent heat of sublimation, and for freezing we care about the mixed-phase particle mass content ``L_p3_tot`` as well as the mixed-phase particle PSD instead of only caring about ice like we do for melting. Since we switch between melting and freezing based on temperature, maybe it would be nicest to implement them both in one function?
- Finally, for catching cases when we have no particles over which to compute melting, I have currently implemented to compute melting if `L > eps && N > eps` but am unsure whether the condition should be that or ``L > eps || N > eps``.

### DIFF
--- a/docs/src/P3Scheme.md
+++ b/docs/src/P3Scheme.md
@@ -345,22 +345,33 @@ We assume the same ventilation factor parameterization as in [SeifertBeheng2006]
 The ``dm/dD`` derivative is computed for each P3 size regime, without taking into account
   ``L_{liq}``, since melting is dependent only on the ice core and not on the entire
   mixed-phase particle.
-The bulk melting rate is computed by numerically integrating over the particle
-  size distribution:
+The bulk melting rates are computed by piecewise and numerically integrating over the ice core
+  size distribution using the terminal velocity of the mixed-phase particle. Particles with ``D < D_{th}``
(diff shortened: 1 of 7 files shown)


rorlija1:
Hi @trontrytel , just requested your review on this, but no rush since of course this is my last week! 
I have a doubt in the code that I'd like to highlight: physically, should we be using the ice core fall speed or the whole (mixed-phase) particle fall speed to calculate the melting rates?

PR #451, opened 2024-08-26
edits so noneq can run in climaatmos

oalcabes:
## Content

----


oalcabes:
Also fixed a bug where we were multiplying by Gamma rather than dividing (added parenthesis)

Issue #459, opened 2024-09-18
High tolerance needed for perf model IN calibration tests


amylu00:
Within the perfect model calibration tests in the `ice_nucleation_calibration.jl` file under `test`, each test is being run with different, arbitrary tolerances. Calibration might not be working as we want it because some values wouldn't pass the test if we set a reasonable relative tolerance.

For example: 
ABDINM calibrated _c_ coefficient is tested with an arbitrary absolute tolerance of 3. This is because the true value is around ~2, but because of the noise and standard deviation given to the priors, the calibrated value could be as high as 4. This is double the true value and requires around rtol = 0.9 which seems wrong. However, if we run the `perfect_calib.jl` file, we see that the resulting parcel output for ICNC matches fairly well with parcel outputs using the true coefficients.

Need to check with Ollie to see if this is okay. If not, find a way for these unit tests to pass with a relative tolerance of ~0.3.

trontrytel:
All ice calibration tests are now running with `0.3` relative tolerance on the final number concentration. This results in similar tolerance for all free parameters but for the second parameter in `ABDINM` experiment. Is that accuracy acceptable, or should we expect better from the algorithm? 

```
    if IN_mode == "ABDINM"
        TT.@test calibrated_parameters[1] ≈ coeff_true[1] rtol = FT(0.3)
        TT.@test calibrated_parameters[2] ≈ coeff_true[2] rtol = FT(1.5)
        TT.@test calibrated_soln[9, end] ≈ true_soln[9, end] rtol = FT(0.3)
    elseif IN_mode == "ABIFM"
        TT.@test calibrated_parameters[1] ≈ coeff_true[1] rtol = FT(0.3)
        TT.@test calibrated_parameters[2] ≈ coeff_true[2] rtol = FT(0.3)
        TT.@test calibrated_soln[9, end] ≈ true_soln[9, end] rtol = FT(0.3)
    elseif IN_mode == "ABHOM"
        TT.@test calibrated_parameters[1] ≈ coeff_true[1] rtol = FT(0.3)
        TT.@test calibrated_parameters[2] ≈ coeff_true[2] rtol = FT(0.3)
        TT.@test calibrated_soln[9, end] ≈ true_soln[9, end] rtol = FT(0.3)
    end
```

PR #460, opened 2024-09-19
Add UKI to AIDA ice nucleation calibrations

amylu00:
## Purpose 

Add UKI calibration so we can compare to the EKI results and also have some error metrics.

## To-do

- what do we want to have in loss function? Freezing onset, gradient, or final ICNC?

## Content

- Calibrate J UKI function in calibration.jl
- Update Calibrate J EKI function so that it works with newest version of EnsembleKalmanProcesses.jl

----


amylu00:
When loss function is total evolution of ICNC, perfect model does well and passes unit tests. When loss function is catered to end of the simulation, the final ICNC matches better in AIDA calibrations, however, perfect model unit tests are prone to failing.

papers/ice_nucleation_2024/calibration_setup.jl
```diff
@@ -133,7 +133,7 @@ function perf_model_pseudo_data(FT, IN_mode, params, IC)
     G_truth = run_model(params, coeff_true, IN_mode, FT, IC)
     dim_output = length(G_truth)
 
-    Γ = 0.03 * LinearAlgebra.I * (maximum(G_truth) - minimum(G_truth))
+    Γ = 0.001 * LinearAlgebra.I * (maximum(G_truth) - minimum(G_truth))
```
trontrytel, line 136: How do we obtain this number?

papers/ice_nucleation_2024/perfect_calib.jl
```diff
@@ -17,29 +17,29 @@ for IN_mode in IN_mode_list
...
     coeff_true = pseudo_data[3]
 
-    output = calibrate_J_parameters(
+    output = calibrate_J_parameters_EKI(
         FT,
         IN_mode,
         params,
         IC,
-        y_truth,
+        y_truth[end-25:end],
```
trontrytel, line 25: Might be good not to hard code 25 as the number but instead have some top level parameter that defines the window we average over at the end of the simulation. And then based on that compute the index value of the cutoff

papers/ice_nucleation_2024/calibration.jl
```diff
@@ -100,7 +102,8 @@ function run_model(p, coefficients, IN_mode, FT, IC)
 
     # solve ODE
     local sol = run_parcel(IC, FT(0), t_max, params)
-    return sol[9, :] ./ (IC[7] + IC[8] + IC[9])  # frozen fraction
+    # return sol[9, end-25:end] ./ (IC[7] + IC[8] + IC[9])  # frozen fraction
+    return sol[9,end-25:end] ./ (IC[7] + IC[8] + IC[9])
```
trontrytel, line 106: Might be good not to hard code 25 as the number but instead have some top level parameter that defines the window we average over at the end of the simulation. And then based on that compute the index value of the cutoff

test/ice_nucleation_calibration.jl
```diff
@@ -14,16 +14,16 @@ function test_J_calibration(FT, IN_mode)
     y_truth = pseudo_data[1]
     coeff_true = pseudo_data[3]
 
-    output = calibrate_J_parameters(
+    output = calibrate_J_parameters_EKI(
```
trontrytel, line 17: Should we also add UKI to tests?

papers/ice_nucleation_2024/calibration.jl
```diff
@@ -9,6 +9,8 @@ import ClimaParams as CP
 import CloudMicrophysics as CM
 import CloudMicrophysics.Parameters as CMP
 import Thermodynamics as TD
+using StatsBase
```
trontrytel, line 12: Could we do it as 

`import StatsBase` as SB`

and then

`SB.function(...)` 

when we use it in the code? This way it's easier to read and tell where different functions are defined

test/Project.toml
```diff
@@ -25,6 +25,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RootSolvers = "7181ea78-2dcb-4de3-ab41-2b8ab5a31e74"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
```
trontrytel, line 28: Do we need that dependency?

trontrytel:
So the `codecov` error seems to be fixed 🎉 

I'm fine with increasing the relative tolerance for the calibration test as needed. Especially since it seems that it just doesn't matted for the loss function. - The final number concentration is ok.

I would vote to
- fix whatever is still needed in this PR and close issue  https://github.com/CliMA/CloudMicrophysics.jl/issues/459
- open an issue about updating to EKI v2 and then trying to update. Maybe in the process we will be able to bring the calibration errors lower

papers/ice_nucleation_2024/calibration_setup.jl
amylu00: Need to ask AIDA people what the noise in their data is. I am using an arbitrary number here for now

test/Project.toml
amylu00: I don't think so. It's not in my local version of the code right now

papers/ice_nucleation_2024/calibration.jl
amylu00: Will open as an issue if the other AIDA data are shorter than 100 timesteps long. Otherwise, I will have it as a variable called `end_sim` for now if that's fine

amylu00:
> I would vote to
> 
> * fix whatever is still needed in this PR and close issue  [High tolerance needed for perf model IN calibration tests #459](https://github.com/CliMA/CloudMicrophysics.jl/issues/459)
> * open an issue about updating to EKI v2 and then trying to update. Maybe in the process we will be able to bring the calibration errors lower

Seems like EKP wasn't up-to-date only in `test` so I removed the dependency on version 1. Everything should be up-to-date with EKP and working fine now.

amylu00:
git commit --allow-empty

amylu00:
git commit --allow-empty -m "Trigger Build"

PR #461, opened 2024-10-02
Cleaning parcel docs

amylu00:
- fixed some equations so they display correctly
- removed some plots
- rearranged headings for better flow


trontrytel (approved):
Thank you, the docs are much better! I changed a couple more lines, lets see if I didn't break anything.

PR #464, opened 2024-10-09
Remove arch from CI jobs and update actions

Sbozzolo:
Specifying arch forces the macOS runner to run in emulation mode (the correct arch is aarch64), making it much slower than it needs to be.

In the process, I also updated some actions and added dependabot to keep them updated


trontrytel:
Thank you!

PR #475, opened 2024-11-06
Revamp Chen2022 parameterization to allow variable apparent ice/snow density

trontrytel:
The `Chen2022` parameterization for terminal velocity depends on ice particles density. We used to assume that this density is constant and the same for cloud ice and snow. As a result we could precompute more coefficients of `Chen2022` parameterization in `parameters/TerminalVelocity.jl` .

Now we want to:
- use different apparent snow and cloud ice densities in the 1M scheme,
- use the predicted particle density in the P3 scheme.

In order to do that we had to move the computation of `Chen2022` coefficients from `parameters/TerminalVelocity.jl` to `Common.jl`. 

This PR is just refactoring and does not change the actual density values. As a result it does not require any changes in our unit tests. I will update those in a separate PR


src/Common.jl
```diff
@@ -208,80 +208,95 @@ function a_w_ice(tps::TPS, T::FT) where {FT}
...
+    # Table B5
+    Al = A[1] + A[2] * log(ρᵢ) + A[3] * (ρᵢ)^FT(-3 / 2)
+    Bl = exp(B[1] + B[2] * log(ρᵢ)^2 + B[3] * log(ρᵢ))
+    Cl = exp(C[1] + C[2] / log(ρᵢ) + C[3] / ρᵢ)
+    El = E[1] + E[2] * log(ρᵢ) * sqrt(ρᵢ) + E[3] * sqrt(ρᵢ)
+    Fl = F[1] + F[2] * log(ρᵢ) + F[3] * exp(-ρᵢ)
+    Gl = (G[1] + G[2] * log(ρᵢ) * sqrt(ρᵢ) + G[3] / sqrt(ρᵢ))^(-1)
+    Hl = H[1] + H[2] * (ρᵢ)^FT(5 / 2) - H[3] * exp(-ρᵢ)
+    # Table B4
+    #TODO
```
nefrathenrici, line 293: What is this TODO for?

src/parameters/Microphysics1M.jl
```diff
@@ -153,16 +153,17 @@ struct CloudIce{FT, PD, MS} <: CloudCondensateType{FT}
...
-    "ice density [kg/m3]"
-    ρi::FT
+    "cloud ice apparent density [kg/m3]"
+    ρᵢ::FT
 end
 
 CloudIce(::Type{FT}) where {FT <: AbstractFloat} =
     CloudIce(CP.create_toml_dict(FT))
 
+#:cloud_ice_apparent_density => :ρi,
```
nefrathenrici, line 163: Are these comments leftover?

src/Microphysics1M.jl
```diff
@@ -191,18 +191,17 @@ function terminal_velocity(
     # from D=125um to D=625um using B2 and D=625um to inf using B4.
     if q > FT(0)
         # coefficients from Table B4 from Chen et. al. 2022
-        aiu, bi, ciu = CO.Chen2022_vel_coeffs_B4(vel, ρ)
-        #aiu, bi, ciu = CO.Chen2022_vel_coeffs_B2(vel, ρ)
+        aiu, bi, ciu = CO.Chen2022_vel_coeffs_B4(vel, ρₐ, ρᵢ)
+        #aiu, bi, ciu = CO.Chen2022_vel_coeffs_B2(vel, ρₐ, ρᵢ)
```
nefrathenrici, line 195: Why is this function call commented out?

nefrathenrici (approved):
Just a few nitpicks

src/Common.jl
trontrytel: Leftover from debugging. Will delete, thank you!

src/parameters/Microphysics1M.jl
trontrytel: I'll change them in the next PR. I just left them there to remember to what density from `ClimaParams.jl` I want to actually switch to

src/Microphysics1M.jl
trontrytel: Thats a bigger TODO that we may or may not chose to address. The Chen 2022 velocity parameterization switches coefficient values between small and large ice particles (table B2 vs B4). Technically in 1M terminal velocity we should integrate by parts and use both tables to get the bulk velocities for cloud condensate and snow. But instead we use B2 for cloud ice and B4 for snow and keep the simple analytical formulations

amylu00 (approved):

src/Microphysics1M.jl
nefrathenrici: Since the TODO isn't addressed in this PR does it make sense to open an issue or leave a code comment to track it?

PR #480, opened 2024-11-11
HET J defaults to 0 for any aerosol type

amylu00:
For heterogeneous freezing modes, parameterization for J will default to 0.

### DIFF
--- a/src/IceNucleation.jl
+++ b/src/IceNucleation.jl
@@ -80,6 +80,7 @@ Returns the deposition nucleation rate coefficient, `J`, in m^-2 s^-1
 for different minerals in liquid droplets.
 The free parameters `m` and `c` are derived from China et al (2017)
 see DOI: 10.1002/2016JD025817
+Returns zero for unsupported aerosol types.
 """
 function deposition_J(
     dust::Union{CMP.Ferrihydrite, CMP.Feldspar, CMP.Kaolinite},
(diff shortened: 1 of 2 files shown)


src/parameters/AbstractAerosol.jl
```diff
@@ -0,0 +1,15 @@
...
+    AbstractAerosol{FT}
+
+Struct of type AerosolType. Refers to an arbitrary aerosol.
+
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+Base.@kwdef struct AbstractAerosol{FT} <: AerosolType{FT} end
+
+function AbstractAerosol(FT)
```
amylu00, line 13: Probably better to have same form as the other aerosol types to keep a generic calibration code
i.e. `function AbstractAerosol(td::CP.AbstractTOMLDict)`. I just don't want to take up space in ClimaParams just to say `J_deposition_m_AbstractAerosol = FT(0)` though. Maybe there's a way to make a toml file within the function that says m and c values for ABDINM and ABIFM are 0...?

trontrytel:
I don't think we need to create a type with empty parameters. Since all aerosols inherit from the `AerosolType`, we just need to specify the two `J` functions to be zero for it. It should then work for any other aerosol, for which we don't have a more specific definition.

It's good to merge if you agree with the above and want to squash/rebase stuff

trontrytel (approved):

PR #483, opened 2024-11-15
Fix DocumenterCitations error

trontrytel:
Our docs stopped building with `DocumenterCitations v1.3.5` . They were working fine with previous versions. 

Following the discussion in https://github.com/JuliaDocs/DocumenterCitations.jl/issues/78 I removed the TeX commands in favor of unicode in authors names. The actual fix is removing `{` from `{Dal Maso}` in line 400. I want to keep the other changes too, because they are more in line with the above mentioned discussion.

### DIFF
--- a/docs/bibliography.bib
+++ b/docs/bibliography.bib
@@ -55,7 +55,7 @@ doi = "10.1039/D1EA00077B",
 }
 
 @Article{Baumgartner2022,
-author = {Baumgartner, M. and Rolf, C. and Groo{\ss}, J.-U. and Schneider, J. and Schorr, T. and M\"ohler, O. and Spichtinger, P. and Kr\"amer, M.},
+author = {Baumgartner, M. and Rolf, C. and Grooß, J.-U. and Schneider, J. and Schorr, T. and Möhler, O. and Spichtinger, P. and Krämer, M.},
 title = {New investigations on homogeneous ice nucleation: the effects of water activity and water saturation formulations},
 journal = {Atmospheric Chemistry and Physics},
 volume = {22},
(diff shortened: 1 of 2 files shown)


amylu00 (approved):

docs/bibliography.bib
```diff
@@ -397,7 +397,7 @@ @article{Koop2002
 
 @article{Lehtinen2007,
   title = {Estimating nucleation rates from apparent particle formation rates and vice versa: Revised formulation of the Kerminen–Kulmala equation},
-  author = {Kari E.J. Lehtinen and Miikka {Dal Maso} and Markku Kulmala and Veli-Matti Kerminen},
+  author = {Kari E.J. Lehtinen and Miikka Dal Maso and Markku Kulmala and Veli-Matti Kerminen},
```
goerz, line 400: This is not correct: it changes the last name "Dal Maso" to "Maso" with "Dal" as a middle name. The correct way to handle this is to use the "Last, First" format.

```
@article{Lehtinen2007,
  title = {Estimating nucleation rates from apparent particle formation rates and vice versa: Revised formulation of the Kerminen–Kulmala equation},
  author = {Lehtinen, Kari E.J. and Dal Maso, Miikka and Kulmala, Markku and Kerminen, Veli-Matti},
  journal = {Journal of Aerosol Science},
  volume = {38},
  number = {9},
  pages = {988-994},
  year = {2007},
  doi = {10.1016/j.jaerosci.2007.06.009}
}
```

goerz:
By the way, I don't think you're [using the correct CSS](https://juliadocs.org/DocumenterCitations.jl/stable/styling/)

PR #487, opened 2024-11-22
Prescribed thermo for AIDA calibrations

amylu00:
Changing AIDA calibrations for HOM so that we are prescribing cooling rate and pressure.


amylu00:
Cooling rate is not completely linear. Seems like latent heat and/or heating from chamber walls has a pretty significant effect later in the experiment. Should I make cooling rate nonlinear in the model to match or is that unnecessary?

Screenshot 2024-11-21 at 5 04 06 PM

trontrytel:
> Cooling rate is not completely linear. Seems like latent heat and/or heating from chamber walls has a pretty significant effect later in the experiment. Should I make cooling rate nonlinear in the model to match or is that unnecessary?

I would vote for not fitting any lines then and just driving it with what is in the measurements. We can interpolate between the two points in time from the measurements to get the rate at each time step

papers/ice_nucleation_2024/calibration.jl
```diff
@@ -58,18 +49,7 @@ function run_model(p, coefficients, IN_mode, FT, IC, end_sim)
...
-            const_dt = const_dt,
-            w = w,
-            aerosol = overwrite,
-            heterogeneous = heterogeneous,
-            condensation_growth = condensation_growth,
-            deposition_growth = deposition_growth,
-            liq_size_distribution = liq_size_distribution,
-            ice_size_distribution = ice_size_distribution,
-        )
+        aerosol = overwrite
```
trontrytel, line 52: We could just delete this line and have `aerosol = CMP.Kaolinite(kaolinite_calibrated)`? And the same for other two options in this file?

parcel/ParcelCommon.jl
```diff
@@ -8,3 +8,15 @@ eᵥ(qᵥ, p_air, R_air, Rᵥ) = qᵥ * p_air * Rᵥ / R_air
 
 # Saturation pressure over ice given saturation pressure over liquid
 S_i(tps, T, S_liq) = ξ(tps, T) * S_liq
+
+# Interpolating for cooling/expansion rate
+function AIDA_rate(model_t, data)
+    last_index = length(data)
```
trontrytel, line 14: Does this assume that the data from AIDA is saved at every second?

parcel/ParcelCommon.jl
```diff
@@ -8,3 +8,15 @@ eᵥ(qᵥ, p_air, R_air, Rᵥ) = qᵥ * p_air * Rᵥ / R_air
...
+    index = Int32(model_t) + 1
+
+    if index < last_index
+        return (data[index + 1] - data[index]) # ΔT/second; ΔT/second
```
trontrytel, line 18: I think this will only work if the model time step is 1 second?

trontrytel (commented):
I'm not sure about the `AIDA_rate` function. Maybe something like this would be nicer?

```
import Interpolations as Intp
# example values

FT = Float32
t_in  = FT[0.130,  0.464,  0.573,  1.100,  1.653,  2.216,  2.760]
p_in = FT[991.3,  954.2,  942.0, 886.9,  831.5,  778.9, 729.8]

p_of_t = Intp.linear_interpolation(t_in, p_in)

p = p_of_t(0.5)
```
Then you could compute the gradient for any `t` and `dt` combination?

trontrytel (approved):
Thank you!

PR #495, opened 2024-12-10
calibrating AIDA HET experiments

amylu00:
## Purpose 

- adding experiments IN07_01 and IN07_19 to the AIDA calibrations

## To-do

- calculate and plot saturation from raw data to see if it's in deposition nucleation conditions or IMM conditions
- properly find initial N_l
- have m and c fields for deposition and immersion ice nucleation for different aerosol types (so that they are calibrate-able)

## Content

----

### DIFF
--- a/papers/ice_nucleation_2024/AIDA_calibrations.jl
+++ b/papers/ice_nucleation_2024/AIDA_calibrations.jl
@@ -33,11 +33,17 @@ moving_average(data, n) =
     [sum(@view data[i:(i + n)]) / n for i in 1:(length(data) - n)]
 
 # Defining data names, start/end times, etc.
-data_file_names = ["in05_17_aida.edf", "in05_18_aida.edf", "in05_19_aida.edf"]
-plot_names = ["IN0517", "IN0518", "IN0519"]
+data_file_names = [
+    "in05_17_aida.edf",
+    "in05_18_aida.edf",
+    "in05_19_aida.edf",
+    "in07_01_aida.edf",
+    "in07_19_aida.edf",
+]
+plot_names = ["IN0517", "IN0518", "IN0519", "IN0701", "IN0719"]
 end_sim = 25                                            # Loss func looks at last end_sim timesteps only
-start_time_list = [195, 180, 80, 50, 35]                # freezing onset
-end_time_list = [290, 290, 170, 800, 800]               # approximate time freezing stops
+start_time_list = [150, 180, 80, 50, 35]                # freezing onset
+end_time_list = [290, 290, 170, 400, 400]               # approximate time freezing stops
 moving_average_n = 20                                   # average every n points
 updrafts = [FT(1.5), FT(1.4), FT(5), FT(1.5), FT(1.5)]  # updrafts matching AIDA cooling rate
 
@@ -49,6 +55,7 @@ R_d = TD.Parameters.R_d(tps)
 
 
 for (exp_index, data_file_name) in enumerate(data_file_names)
+    @info(data_file_name)
     #! format: off
     ### Unpacking experiment-specific variables.
     plot_name = plot_names[exp_index]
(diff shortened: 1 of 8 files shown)


amylu00:
IN07_01:
image

IN07_19:
Screenshot 2024-12-12 at 4 31 28 PM

parcel/ParcelModel.jl
```diff
@@ -141,6 +141,7 @@ function parcel_model(dY, Y, p, t)
     dqᵥ_dt = -dqₗ_dt_v2l - dqᵢ_dt_v2i
 
     dSₗ_dt =
+    # prescribed_thermodynamics ? AIDA_rate(t, t_profile, T_profile) / (-0.0065) :
```
trontrytel, line 144: Is that line supposed to be commented out?

papers/ice_nucleation_2024/calibration_setup.jl
```diff
@@ -247,8 +246,84 @@ function AIDA_IN05_IC(FT, data_file)
         eₛ = TD.saturation_vapor_pressure(tps, T₀, TD.Liquid())
         Sₗ = FT(e / eₛ)
     end
-    return [Sₗ, p₀, T₀, qᵥ, qₗ, qᵢ, Nₐ, Nₗ, Nᵢ, FT(0)]   #remove the last 2 elements, its J & r_l
+    return [Sₗ, p₀, T₀, qᵥ, qₗ, qᵢ, Nₐ, Nₗ, Nᵢ, FT(0)]
+end
+
+function AIDA_IN07_params(FT, w, t_max, t_profile, T_profile, P_profile, plot_name)
```
trontrytel, line 252: Are you still using anywhere the generic `perf_model_params` and `perf_model_IC`? Or should we switch to only using the case specific ones?

amylu00: I'm using it as a unit test and there are also plots for it in the papers folder. I could get rid of the plots since we already know it is calibrate-able now?

parcel/ParcelModel.jl
amylu00: Yes, it wasn't working right (saturation was order of magnitude 1000) so I'm going to use an approximate of the updraft for now

papers/ice_nucleation_2024/calibration.jl
```diff
@@ -178,13 +218,17 @@ function create_prior(FT, IN_mode, ; perfect_model = false)
         end
     elseif perfect_model == false
         if IN_mode == "ABDINM"
-            # m_stats = [FT(20), FT(1), FT(0), Inf]
-            # c_stats = [FT(-1), FT(1), -Inf, Inf]
-            println("Calibration for ABDINM with AIDA not yet implemented.")
+            if aerosol_type == CMP.ArizonaTestDust(FT)
+                m_stats = [FT(40), FT(20), FT(0), Inf]
```
trontrytel, line 222: Why do we have `Infs` here?

papers/ice_nucleation_2024/AIDA_calibrations.jl
```diff
@@ -131,30 +146,30 @@ for (exp_index, data_file_name) in enumerate(data_file_names)
     
     MK.lines!(ax_parcel_1, EKI_parcel.t, EKI_parcel[1, :], label = "EKI Calib Liq", color = :orange) # label = "liquid"
     MK.lines!(ax_parcel_1, UKI_parcel.t, UKI_parcel[1, :], label = "UKI Calib Liq", color = :fuchsia) # label = "liquid"
-    MK.lines!(ax_parcel_1, parcel_default.t, parcel_default[1, :], label = "default", color = :darkorange2)
+    # MK.lines!(ax_parcel_1, parcel_default.t, parcel_default[1, :], label = "default", color = :darkorange2)
```
trontrytel, line 149: Should we just delete those instead of commenting them out?

trontrytel (approved):
Left some comments. There is one commented out line in the parcel code that I don't think should be commented? Looks good otherwise!

parcel/ParcelModel.jl
trontrytel: Then we should figure out why. We can't run calibrations with the updraft version because our pessure is 1000x ff then?

PR #497, opened 2024-12-12
Add ClimaCore GPU test

trontrytel:


trontrytel:
I still can't reproduce the error in `Atmos`, even though now I'm testing on (I think) the same 3D space.

But It's a step in right direction, so I want to merge the test in.

PR #508, opened 2025-01-29
Nonequilibrium in the parcel model

oalcabes:
## Content

----


parcel/ParcelParameters.jl
```diff
@@ -82,7 +82,29 @@ struct CondParams{FT} <: CMP.ParametersType{FT}
     const_dt::FT
 end
 
+struct NonEqCondParams_Anna{FT} <: CMP.ParametersType{FT}
```
trontrytel, line 85: Could we avoid my name? We could name it `NonEqCondParams_simple` (or stupid). Or just not include it alltogether and just show the MM2015 option.

parcel/ParcelTendencies.jl
```diff
@@ -223,6 +229,40 @@ function condensation(params::CondParams, PSD_liq, state, ρ_air)
...
+    FT = eltype(state)
+    (; T, qₗ, qᵥ, qᵢ) = state
+
+    (; tps, liquid, dt) = params
+
+    q = TD.PhasePartition(qᵥ + qₗ + qᵢ, qₗ, qᵢ)
+
+    cond_rate = MNE.conv_q_vap_to_q_liq_ice_MM2015(liquid, tps, q, ρ_air, T)
+
+    # using same limiter as ClimaAtmos for now
```
trontrytel, line 258: Would be good to do some testing and see if we need the limiter. Maybe we don't? Or if we do, maybe we can limit just by dt?

parcel/Example_NonEq.jl
```diff
@@ -0,0 +1,131 @@
...
+        condensation_growth = condensation_growth,
+        deposition_growth = deposition_growth,
+        const_dt = const_dt,
+        w = w,
+        liquid = liquid,
+        ice = ice,
+    )
+
+    # solve ODE
+    local sol = run_parcel(IC, FT(0), t_max, params)
```
trontrytel, line 89: It would be neat to compare all two (or three depending on if you kick out the simple one) together. And it would be even neater to add a couple of words and the resulting plot to our docs, along with other parcel examples. This way at least your example will be run when we are building the docs during PRs.

trontrytel (approved):
I left a couple of comments. Mostly about removing my name and trying to add the resulting plot to the docs somewhere.

LGTM otherwise! Thank you. Just remember to squash and rebase before merging

parcel/ParcelParameters.jl
oalcabes: Great -- I'm planning to just rename it "simple." I don't mind keeping it in the parcel just for testing reasons. But if we want to eventually get rid of it in the Noneq function all together, we can do that one day.

PR #510, opened 2025-02-06
Add cloud diagnostics

trontrytel:
This PR:
- moves the existing 1M and 2M radar reflectivity and effective radius to a CloudDiagnostics module
- adds a simple constant effective radius function (based on ClimaParams)
- adds a wrapper for Lu and Hallet effective radius parameterization that assumes N=100 (allows it to be used with 0M and 1M schemes)


szy21 (approved):
I didn't go through any equations but the code change looks good to me.

PR #513, opened 2025-02-12
Add more badges

trontrytel:


AlexisRenchon:
I am not sure why it seems your badge is counting CI downloads https://juliapkgstats.com/pkg/CloudMicrophysics (if you click on CI on that web app you will see. The user numbers look normal, CI is very high)

trontrytel:
Yes, I'm not sure why the CI gets counted. The docs on the package states webpage are very sparse, so I'm not sure how to generate the badge excluding CI

PR #514, opened 2025-02-14
Fix averaging on AIDA

amylu00:
## Purpose 

- fix moving average/ smoothing of AIDA data in ice nucleation calibrations

## Content

- fixed moving average
- loss func is the average frozen fraction of the last 25 seconds (essentially a constant value of ff in the last 25 seconds)

----


papers/ice_nucleation_2024/calibration_setup.jl
```diff
@@ -195,51 +195,34 @@ function AIDA_IN05_IC(FT, data_file)
     ϵₘ = R_d / R_v
 
     if data_file == "in05_17_aida.edf"
-        Nₗ = FT(317.559 * 1e6)
-        Nᵢ = FT(0.0 * 1e6)
+        Nₗ = FT(320.876 * 1e6)
```
trontrytel, line 198: Are all those changes to initial conditions needed?

trontrytel (approved):

papers/ice_nucleation_2024/calibration_setup.jl
amylu00: Yeah because before, the way the average was calculated meant the experiment started a few seconds after and ended a few seconds earlier. Now the edges are also fixed so the start time is sooner and has different initial conditions.

PR #516, opened 2025-02-22
P3 refactor

haakon-e:
## Purpose 

Refactor `P3` implementation to be more robust to a wide range of numerical values and formalize API.

## Content

- Rewrite distribution parameter solver in log-space to ensure accurate solution across a range of parameter regimes
- Implement constant slope parameterization which circumvents multiple \lambda solution phenomena of traditional power law parameterization
- Implement exact solution for thresholds
- Formalize API: introduce `P3State`, `P3Distribution` which contains variables obtained at various solution stages
- Update tests
- Document new solution procedure in P3 docs.

----


docs/src/plots/P3Melting.jl
```diff
@@ -1,15 +1,21 @@
...
 import CloudMicrophysics.Parameters as CMP
 import CloudMicrophysics.P3Scheme as P3
-
+import ClimaParams as CP
 FT = Float64
 
 # parameters
 tps = TD.Parameters.ThermodynamicsParameters(FT)
-pps = CMP.ParametersP3(FT)
+params = CMP.ParametersP3(
```
trontrytel, line 12: Won't be needed once we merge the PR in ClimaParams

docs/src/plots/P3Melting.jl
```diff
@@ -1,15 +1,21 @@
-import CairoMakie as PL
-PL.activate!(type = "svg")
+using CairoMakie
```
trontrytel, line 1: I prefer the `import CairoMakie as PL` way. Because it makes it clear in the code from which package the functions are coming from. But I'm not going to argue

src/P3.jl
```diff
@@ -1,15 +1,18 @@
 """
-Predicted particle properties scheme P3 for ice.
-Implementation of Morrison and Milbrandt 2015 doi: 10.1175/JAS-D-14-0065.1
+Predicted particle properties scheme P3 for ice [MorrisonMilbrandt2015](@cite).
 
-Note: Particle size is defined as its maximum length (i.e. max dimesion).
+!!! note "Particle size definition"
```
trontrytel, line 4: This part shows in the API docs, but I think it's not needed there. Maybe we can just move it to the documentation page.

src/P3.jl
```diff
@@ -1,15 +1,18 @@
...
 """
 module P3Scheme
 
+using DocStringExtensions
+
 import SpecialFunctions as SF
 import QuadGK as QGK
 import RootSolvers as RS
 import HCubature as HC
+import LogExpFunctions: logsumexp
```
trontrytel, line 15: I would prefer to keep the pattern and do something like 

```
import LogExpFunctions as LEF

LEF.logsumexp(blah)
```

src/P3.jl
```diff
@@ -21,12 +24,12 @@ import Thermodynamics as TD
 import Thermodynamics.Parameters as TDP
 
 const PSP3 = CMP.ParametersP3
+using CloudMicrophysics.Parameters:
+    MassPowerLaw, AreaPowerLaw, SlopeLaw, SlopePowerLaw, SlopeConstant
```
trontrytel, line 28: Same here. Would prefer to keep the qualifier to see in the code where the parameters are coming from

docs/src/plots/P3Melting.jl
```diff
@@ -1,15 +1,21 @@
...
 import CloudMicrophysics.P3Scheme as P3
-
+import ClimaParams as CP
 FT = Float64
 
 # parameters
 tps = TD.Parameters.ThermodynamicsParameters(FT)
-pps = CMP.ParametersP3(FT)
+params = CMP.ParametersP3(
+    CP.create_toml_dict(
```
trontrytel, line 13: To be changed after ClimaPrams PR is merged

src/parameters/MicrophysicsP3.jl
```diff
@@ -1,63 +1,418 @@
...
+# Examples
 
-ParametersP3(::Type{FT}) where {FT <: AbstractFloat} =
-    ParametersP3(CP.create_toml_dict(FT))
+```julia-repl
+julia> import CloudMicrophysics.Parameters as CMP
+julia> CMP.ParametersP3(Float64)
+ParametersP3
+├── mass: MassPowerLaw
+│   ├── α_va = 0.018537721864540644 [kg μm^(-β_va)]
```
trontrytel, line 279: I'm not sure that showing the actual numbers in the docstring is a good idea. This will become outdated as soon as we change the default values in `ClimaParams` or overwrite anything?

Also, showing how to print values from a struct in `Julia` as part of our docs might be too verbose?

src/parameters/MicrophysicsP3.jl
```diff
@@ -1,63 +1,418 @@
...
+    ParametersP3{FT}
+
+Parameters for P3 bulk microphysics scheme.
+    
+From [MorrisonMilbrandt2015](@cite)
+
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+Base.@kwdef struct ParametersP3{FT, SLOPELAW <: SlopeLaw{FT}} <:
```
trontrytel, line 249: I thought that for some GPU compilation reasons we had to specify all the types here. But I could be wrong

src/Common.jl
```diff
@@ -302,7 +302,7 @@ end
 # Wrapper to cast types from SF.gamma
 # (which returns Float64 even when the input is Float32)
 # TODO - replace with parameterization of our own
-Γ(a::FT) where {FT <: Real} = FT(SF.gamma(a))
+Γ(a::FT) where {FT <: Real} = FT(SF.gamma(a))  # TODO: Don't need this. Float32 is correctly handled.
```
trontrytel, line 305: Then lets delete this

docs/src/P3Scheme.md
```diff
@@ -1,225 +1,351 @@
...
+|small, spherical ice                  | $D < D_{th}$          |                | $\frac{π}{6} ρ_i D^3$                 | $\frac{π}{4} D^2$                                |
+|large, unrimed ice                    | $D_{th} < D$          | $L_{rim} = 0$  | $α_{va} D^{β_{va}}$                   | $γ D^{σ}$                                        |
+|dense nonspherical ice                | $D_{th} < D < D_{gr}$ | $L_{rim} > 0$  | $α_{va} D^{β_{va}}$                   | $γ D^{σ}$                                        |
+|graupel (completely rimed, spherical) | $D_{gr} < D < D_{cr}$ | $L_{rim} > 0$  | $\frac{π}{6} ρ_g D^3$                 | $\frac{π}{4} D^2$                                |
+|partially rimed ice                   | $D_{cr} < D$          | $L_{rim} > 0$  | $\frac{α_{va}}{1-F_{rim}} D^{β_{va}}$ | $F_{rim} \frac{π}{4} D^2 + (1-F_{rim})γ \ D^{σ}$ |
+
+where $D_{th}$, $D_{gr}$, $D_{cr}$, and $ρ_g$ are determined by relations below; $ρ_i$, $β_{va}$, $α_{va}$, $γ$, and $σ$ are fixed.
+
+!!! details "Symbol definitions"
```
trontrytel, line 45: Love the collapsable boxes!

docs/src/P3Scheme.md
```diff
@@ -1,225 +1,351 @@
...
-  ``A = \frac{x_1 - x_2}{y_1 - y_2}``.
-``y_1`` and ``y_2`` define ``log(\lambda)`` values for three ``L/N`` ranges
+With this choice, it appears that some values of $\log(L/N)$ gives rise to multiple solutions for $λ$, as seen in the plot below.
 
-|             L/N                |         ``y_1``        |      ``y_2``         |
-|:-------------------------------|:-----------------------|:---------------------|
-|        ``L/N >= 10^-8``        |         ``1``          |     ``6 * 10^3``     |
-|   ``2 * 10^9 <= L/N < 10^-8``  |     ``6 * 10^3``       |     ``3 * 10^4``     |
-|        ``L/N < 2 * 10^9``      |     ``4 * 10^4``       |       ``10^6``       |
+!!! todo "TODO: Add plot of multiple solutions"
```
trontrytel, line 292: That would be nice to add

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+
+Create a [`P3State`](@ref) object from a [`CMP.ParametersP3`](@ref) object and rime state parameters.
+
+# Arguments
+- `params`: [`CMP.ParametersP3`](@ref) object
+- `F_rim`: rime mass fraction
+- `ρ_r`: rime density
 
- - p3 - a struct with P3 scheme parameters
+!!! note "Floating point type"
```
trontrytel, line 52: Not sure we need that. The floating point type of everything have to be the same. I.e. all parameters, cached variables and state variables

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+    # rime mass fraction must always be non-negative ...
+    # ... and there must always be some unrimed part
+    @assert 0 ≤ F_rim < 1 "Rime mass fraction, `F_rim`, must be between 0 and 1"
+
+    (; mass, ρ_i, ρ_l) = params
+
+    D_th = get_D_th(mass, ρ_i)
+
+    if iszero(F_rim)
+        ρ_g = D_gr = D_cr = FT(NaN)
```
trontrytel, line 73: Should they be NaN or just zero in this case?

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+    else
+        # rime density must be positive ...
+        # ... and as a bulk ice density can't exceed the density of water
+        @assert 0 < ρ_r <= ρ_l
+
+        ρ_d = get_ρ_d_exact(mass, F_rim, ρ_r)
```
trontrytel, line 79: Maybe just `get_ρ_d` ?

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+    return P3State(; params, F_rim, ρ_r, ρ_g, D_th, D_gr, D_cr)
+end
+
+Base.eltype(::P3State{FT}) where {FT} = FT
+Base.broadcastable(state::P3State) = tuple(state)
+
+get_parameters(state::P3State) = state.params
+
+isrimed(state::P3State) = state.F_rim > 0
+isunrimed(state::P3State) = !isrimed(state)
```
trontrytel, line 95: I think it's enough to just provide one option?

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+- `D_cr`: `ρ = ρ_g * (1 - F_rim)` (see [`get_D_cr`](@ref))
+
+# Arguments
+- `params`: [`CMP.MassPowerLaw`](@ref) parameters
+- `ρ`: density [kg/m³]
 """
-ρ_g_helper(ρ_r::FT, F_rim::FT, ρ_d::FT) where {FT} =
-    F_rim * ρ_r + (1 - F_rim) * ρ_d
+get_threshold((; α_va, β_va)::MassPowerLaw, ρ) =
+    (6α_va / (π * ρ))^(1 / (3 - β_va))
```
trontrytel, line 192:
```suggestion
    (6α_va / (FT(π) * ρ))^(1 / (3 - β_va))
```

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
-        return (6 * α_va_si(p3)) / (FT(π) * (1 - F_rim)) * D^(p3.β_va - 3)
+    volume_sphere(D)
+
+Calculate the volume of a sphere with diameter D.
+
+```math
+V = \\frac{4}{3} π D^3
+```
+"""
+volume_sphere(D) = D^3 * π / 6
```
trontrytel, line 244:
```suggestion
volume_sphere(D) = D^3 * FT(π) / 6
```

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
-        return α_va_si(p3) / (1 - F_r) * p3.β_va * D^(p3.β_va - 1)
-    end
+    area_spherical(D)
+
+Calculate the cross-sectional area of a spherical particle.
+
+# Arguments
+- `D`: maximum particle dimension [m]
+"""
+area_spherical(D) = D^2 * π / 4
```
trontrytel, line 386:
```suggestion
area_spherical(D) = D^2 * FT(π) / 4
```

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+- `ρ_d`: density of the unrimed portion of the particle [kg/m³], see [`get_ρ_d_exact`](@ref)
+
+# Returns
+- `ρ_g`: density of total (deposition + rime) ice mass for graupel [kg/m³]
+
+# Examples
+
+```julia
+FT = Float64
+ρ_r, F_rim, ρ_d = FT(916.7), FT(0.5), FT(916.7)
```
trontrytel, line 164: Not sure if those examples are helpful or distracting here. We have plenty of use examples in the unit tests. Plus, thanks to the CI they get updated all the time. I'm worried that those examples here will become outdated soon

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
-Returns the liquid fraction weighted average of X_ice and X_liq.
+Return the weighted average of `a` and `b` with fraction `f_a`,
+    
+```math
+f_a ⋅ a + (1 - f_a) ⋅ b
+```
 """
-function p3_F_liq_average(F_liq::FT, X_ice::FT, X_liq::FT) where {FT}
-    return (1 - F_liq) * X_ice + F_liq * X_liq
+function weighted_average(f_a, a, b)
```
trontrytel, line 231: Looks like it's only used once, so maybe we don't need to define a special function for it?

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
-    elseif th.D_gr > D >= D_th_helper(p3)
-        # dense nonspherical ice
-        return (6 * α_va_si(p3)) / FT(π) * D^(p3.β_va - 3)
-    elseif th.D_cr > D >= th.D_gr
-        # graupel
-        return th.ρ_g
-    else #elseif D >= th.D_cr
-        # partially rimed ice
-        return (6 * α_va_si(p3)) / (FT(π) * (1 - F_rim)) * D^(p3.β_va - 3)
+    volume_sphere(D)
```
trontrytel, line 236: Could we put it in Common? I can double check later if we can reuse it in 1M or 2M schemes

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+    mass_spherical(ρ, D)
+
+Calculate the mass as a function of size for spherical particles, used for small ice, liquid, and graupel.
+
+# Arguments
+- `ρ`: bulk density [kg m⁻³]
+- `D`: maximum particle dimension [m]
+
+"""
+mass_spherical(ρ, D) = ρ * volume_sphere(D)  # TODO: used to be: `mass_s` (update tests)
```
trontrytel, line 256: Is this TODO still needed?

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+"""
+    mass_nonspherical(mass, D)
+
+Calculate the mass as a function of size for large nonspherical ice or dense nonspherical ice.
+
+# Arguments
+- `mass`: [`CMP.MassPowerLaw`](@ref) parameters
+- `D`: maximum particle dimension [m]
+"""
+mass_nonspherical((; α_va, β_va)::MassPowerLaw, D) = α_va * D^β_va  # TODO: used to be: `mass_nl`
```
trontrytel, line 267: Same as above

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+    mass_rimed(mass, D, F_rim)
+
+Calculate the mass as a function of size for partially rimed ice.
+
+# Arguments
+- `mass`: [`CMP.MassPowerLaw`](@ref) parameters
+- `D`: maximum particle dimension [m]
+- `F_rim`: rime mass fraction [`L_rim/L_ice`]
+"""
+mass_rimed((; α_va, β_va)::MassPowerLaw, D, F_rim) = α_va * D^β_va / (1 - F_rim)  # TODO: used to be: `mass_r`
```
trontrytel, line 279: And here

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+
+"""
+    ice_mass(state, D)
+
+Return the mass of a particle based on where it falls in the particle-size-based properties regime.
+
+Calls the appropriate mass function based on the particle state,
```
trontrytel, line 286:
```suggestion
```

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+"""
+mass_rimed((; α_va, β_va)::MassPowerLaw, D, F_rim) = α_va * D^β_va / (1 - F_rim)  # TODO: used to be: `mass_r`
+
+"""
+    ice_mass(state, D)
+
+Return the mass of a particle based on where it falls in the particle-size-based properties regime.
+
+Calls the appropriate mass function based on the particle state, 
+    [`mass_spherical`](@ref), [`mass_nonspherical`](@ref), [`mass_rimed`](@ref).
```
trontrytel, line 287:
```suggestion
```

src/P3_terminal_velocity.jl
```diff
@@ -1,25 +1,21 @@
 """
-    ice_particle_terminal_velocity(p3, D, Chen2022, ρₐ, F_rim, th, use_aspect_ratio)
+    ice_particle_terminal_velocity(state, D, Chen2022, ρₐ, use_aspect_ratio)
```
trontrytel, line 2: `state` might be confusing once we couple to the Atmos model, because there we use it to describe the whole model state (i.e. all prognostic variables). Maybe something like `p3_state`?

src/P3_terminal_velocity.jl
```diff
@@ -95,134 +83,99 @@ cloud or rain drop as a function of their sizes. It uses Chen 2022 velocity
...
-    F_rim::FT,
-    F_liq::FT,
-    th,
     use_aspect_ratio = true,
 ) where {FT}
     # velocity difference for cloud-ice collisions
     return abs(
         p3_particle_terminal_velocity(
-            p3,
+            state,
```
trontrytel, line 120: Actually, this should now use the sedimentation terminal velocity for cloud liquid. You can fix it now or open an issue and we can do it separately

src/P3_terminal_velocity.jl
```diff
@@ -95,134 +83,99 @@ cloud or rain drop as a function of their sizes. It uses Chen 2022 velocity
...
-    N::FT,
-    ρ_r::FT,
-    F_rim::FT,
-    F_liq::FT,
     ρₐ::FT,
-    use_aspect_ratio = true,
+    use_aspect_ratio = true;
+    accurate = false,
 ) where {FT}
+    state = get_state(dist)
```
trontrytel, line 156: `p3_state` ?

src/P3_particle_properties.jl
```diff
@@ -1,333 +1,498 @@
...
+
+```julia
+FT = Float64
+params = ParametersP3(FT)
+state = get_state(params; F_rim = FT(0.5), ρ_r = FT(916.7))
+```
 """
-D_th_helper(p3::PSP3{FT}) where {FT} =
-    (FT(π) * p3.ρ_i / 6 / α_va_si(p3))^(1 / (p3.β_va - 3))
+function get_state(params::PSP3{FT}; F_rim, ρ_r) where {FT}
```
trontrytel, line 63: Thinking about coupling with the Atmos model I would suggest to avoid caching too much. So rather than precomputing `F_rim` and passing it around, it is better to compute on the fly. 

Now that I think about this, I wonder if we even want to cache the thresholds into the struct. I used it do this because they were solved for rootsolver. Bu now that we have analytical formulae for them, it might be more efficient to recompute them when needed

src/P3_processes.jl
```diff
@@ -64,53 +64,46 @@ end
...
 
-    # Get the P3 diameter distribution...
-    th = thresholds(p3, ρ_rim, F_rim)
-    (λ, N_0) =
-        distribution_parameter_solver(p3, L_ice, N_ice, ρ_rim, F_rim, F_liq_)
-    N(D) = N′ice(p3, D, λ, N_0)
-    # ... and D_max for the integral
-    bound = get_ice_bound(p3, λ, N_ice, FT(1e-6))
+    state = get_state(dist)
+    params = get_parameters(dist)
```
trontrytel, line 83: Not sure if its a good idea to "repackage" the parameters into a bigger struct. Maybe it would be better to keep passing smaller structs and here pass only the ones we need.

test/p3_tests.jl
```diff
@@ -1,42 +1,61 @@
...
-        end
+function test_p3_state_creation(FT)
+    @testset "P3State Creation and Properties" begin
+        # Test creating a state with valid parameters
+        FT = Float64
+        params = CMP.ParametersP3(FT)
+
+        # Test unrimed state
+        state_unrimed = P3.get_state(params; F_rim = FT(0.0), ρ_r = FT(400))
+        @test P3.isunrimed(state_unrimed)
```
trontrytel, line 18: I don't think we need both?

test/p3_tests.jl
```diff
@@ -1,42 +1,61 @@
...
+            F_rim = FT(0.5),
+            ρ_r = FT(-400),
+        )
     end
 end
 
 function test_thresholds_solver(FT)
 
-    p3 = CMP.ParametersP3(FT)
+    params = CMP.ParametersP3(FT)
```
trontrytel, line 56: Thinking about coupling with Atmos model. `params` is usually a name used for the struct with all parameters. We might want to use a different name here, maybe?

test/p3_tests.jl
```diff
@@ -45,216 +64,169 @@ function test_thresholds_solver(FT)
...
-                            μ_ex = P3.DSD_μ(p3, λ_ex)
-                            N₀_ex = P3.DSD_N₀(p3, N, λ_ex)
-                            # Find the P3 scheme  thresholds
-                            th = P3.thresholds(p3, ρ_r, F_rim)
-                            # Convert λ to ensure it remains positive
-                            x = log(λ_ex)
+    slope_laws = (:constant, :powerlaw)
+    for slope_law in slope_laws
+        override_file = if slope_law == :constant
+            Dict("Heymsfield_mu_coeff1" => Dict("value" => 3.0))
```
trontrytel, line 185: Not needed anymore

test/p3_tests.jl
```diff
@@ -772,50 +549,44 @@ function test_p3_melting(FT)
...
+        @test rate.dNdt == Nᵢ
+        @test rate.dLdt == Lᵢ
     end
 end
 
 for FT in [Float32, Float64]
     @info("Testing " * string(FT))
 
+    # state creation
+    test_p3_state_creation(FT)
```
trontrytel, line 587: Given all the changes to the internals of P3 scheme, do any of the tests below work for `Float32` also?

src/P3_size_distribution.jl
```diff
@@ -1,361 +1,339 @@
...
-    # Get thresholds for given F_rim, ρ_r
-    th = thresholds(p3, ρ_r, F_rim)
+    log_LdN₀(state, log_λ)
 
-    # Get min and max lambda values
-    λ_0 = μ_to_λ(p3, FT(0))
-    λ_6 = μ_to_λ(p3, p3.μ_max)
+Compute `log(L/N₀)` given the `state` and `log_λ`
+"""
+function log_LdN₀(state::P3State{FT}, log_λ) where {FT}
```
trontrytel, line 129: I'm wondering if `log_L_over_N₀` or ``log_L_div_N₀` would be better. `dN` could be confusing because you are using `dD1` for integrating over things in other places

trontrytel (approved):
Amazing job! Thank you so much for taking this on. 

I left some small comments for your consideration thinking about how this would look integrated in Atmos. But feel free to disregard them. We can always change the parameters and state structs when we actually get to coupling things.

Some of the doc strings I would see better as some sort of tutorial page. But we can move them later once the whole scheme is done

trontrytel:
And LMK if you need a second pair of eyes for the GPU errors. In the next PR we could also consider adding more to the ClimaCore GPU tests.

src/P3_particle_properties.jl
haakon-e: this will be needed for liquid fraction calculations.

src/P3_particle_properties.jl
haakon-e: How about:
```julia
volume_sphere(D::FT)::FT where {FT} = D^3 * π / 6
```
this ensures the return type of the function is `FT`.

src/parameters/MicrophysicsP3.jl
```diff
@@ -1,63 +1,418 @@
...
     FT = CP.float_type(td)
-    return ParametersP3{FT}(; parameters...)
+    return ParametersP3{FT, typeof(slope)}(; mass, area, slope, vent, params...)
+end
+
+### ----------------- ###
+### ----- UTILS ----- ###
+### ----------------- ###
+
+function Base.show(
```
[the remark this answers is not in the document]
haakon-e: This is what enables pretty printing of all structs in the Union below

src/parameters/MicrophysicsP3.jl
haakon-e: I showed this to Dennis. We are specifying all (parameterized) types. The types not specified up here are not needed because they are concrete types, as opposed to abstract types.

src/parameters/MicrophysicsP3.jl
```diff
@@ -1,63 +1,418 @@
...
+├── vent: VentilationSB2005
+│   ├── vent_a = 0.78 [-]
+│   └── vent_b = 0.308 [-]
+├── ρ_i = 916.7 [kg m⁻³]
+├── ρ_l = 1000.0 [kg m⁻³]
+└── T_freeze = 273.15 [K]
+```
+"""
+function ParametersP3(td::CP.AbstractTOMLDict; slope_law = :powerlaw)
+    @assert slope_law in (:constant, :powerlaw)
```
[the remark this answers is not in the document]
haakon-e: Do you mean changing the variable name `slope_law` to `is_constant`, and the value to be `true` or `false`?

src/P3_particle_properties.jl
haakon-e: I personally don't find it intuitive to search around unit tests to learn about method functionality. 
I am updating the examples so the code is run alongside building docs. 
Output is automatically checked if it's correct, and if not (e.g. a code change), can be easily updated by running:
```julia
using Documenter, CloudMicrophysics
# doctest(CloudMicrophysics, fix=false) # default: only reports inconsistencies
doctest(CloudMicrophysics, fix=true)  # overwrites the existing `jldoctest`s with the new method output
```
Let me know what you think about this approach.

src/P3_particle_properties.jl
haakon-e: I checked other parts of the source code, and it appears that volume is variably calculated from radius or diameter.
I added two methods to Common, one for each case, named `volume_sphere_D` and `volume_sphere_R`. I haven't entirely convinced myself that this is the right approach, but I want to make sure no one mistankedly calculates volume using the wrong initial quantity.

test/p3_tests.jl
haakon-e: All do, with some minor adjustments!

haakon-e (commented):
response to most review comments.

docs/src/P3Scheme.md
haakon-e: Added to backlog. Will do in next PR, ref #529

PR #539, opened 2025-03-31
specific contents jargon clarifications (to avoid humidity of snow, ice, rain, ...)

slayoo:
## Purpose 
Avoid misleading jargon featuring humidities of snow, humidities of water, etc by reserving the term "humidity" for humidity, and using specific contents for all other solid and liquid phase water contents.

## To-do
n/a

## Content
jargon changes in docs and code comments

### DIFF
--- a/docs/src/Glossary.md
+++ b/docs/src/Glossary.md
@@ -2,14 +2,14 @@
 
 A scheme that predicts one moment of the particle size distribution (PSD).
 Typically it's the 3rd moment of PSD.
-The scheme's prognostic variable is the specific humidity (mass fraction)
+The scheme's prognostic variable is the specific content (mass fraction)
   of water in each category, which is proportional to the 3rd moment of PSD.
 
 ## 2-moment scheme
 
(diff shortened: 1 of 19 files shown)


trontrytel:
@slayoo - Thanks. I have struggled with a good name here. 

I was worried that `liquid water content` will more often than not be associated with `kg/m3`. And I was trying to highlight that this is in `kg/kg`. But the LWC definition does not really specify by what you are normalizing, so I guess it's fine.

Not sure why buildkite was not triggered. But I'll try to make him

slayoo:
> @slayoo - Thanks. I have struggled with a good name here.
> 
> I was worried that `liquid water content` will more often than not be associated with `kg/m3`. And I was trying to highlight that this is in `kg/kg`. But the LWC definition does not really specify by what you are normalizing, so I guess it's fine.
> 
> Not sure why buildkite was not triggered. But I'll try to make him

I'd say that indeed "liquid water content" is a per-volume quantity, but "liquid water specific content" is defined per-mass, as other "specific" quantities (specific humidity, specific heat capacities, etc)

PR #543, opened 2025-04-04
Add bulk collision rate with liquid for P3 distribution

haakon-e:
### TL;DR

Added a new function to calculate bulk collision rates between ice and liquid particles in the P3 microphysics scheme.

### What changed?

Implemented the `bulk_collision_rate_with_liquid` function in `P3_processes.jl` that:
- Calculates collision rates between ice distribution and liquid particles (rain or cloud)
- Computes changes in number concentration and mass for both ice and liquid phases
- Handles collision cross-section, terminal velocities, and particle size distributions
- Returns changes in ice number concentration, ice mass, rime mass, rime volume, and liquid properties

### How to test?

- Create instances of `P3Distribution` and liquid distribution (either `RainParticlePDF_SB2006` or `CloudParticlePDF_SB2006`)
- Call the function with appropriate parameters (ice distribution, velocity type, liquid distribution, air density, liquid mass, liquid number concentration)
- Verify the returned rates match expected physical behavior

### Why make this change?

This function is essential for modeling ice-liquid interactions in mixed-phase clouds. It enables the simulation of collision processes between ice particles and liquid droplets, which is a key mechanism for precipitation formation and cloud evolution in the P3 microphysics scheme.

### DIFF
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -337,6 +337,39 @@ function Chen2022_exponential_pdf(a::FT, b::FT, c::FT, λ_inv::FT, k::Int) where
 end
 
 """
+    liquid_particle_terminal_velocity(velocity_params, ρₐ)
+    liquid_particle_terminal_velocity(velocity_params, ρₐ, D)
+
+Compute the terminal velocity of a liquid particle as a function of its size 
+    (maximum dimension, `D`) using the Chen 2022 parametrization.
+
+# Arguments
+- `velocity_params`: a struct with terminal velocity parameters from Chen 2022
+- `ρₐ`: air density [kg/m³]
+- `D`: maximum particle dimension [m]
+
+# Returns
+- The first method returns a function of `D`, while the second evaluates at a specific `D`.
+
+Needed for numerical integrals in the P3 scheme.
+
+!!! note
+    We use the same terminal velocity parametrization for cloud and rain water.
+"""
+function liquid_particle_terminal_velocity(velocity_params::CMP.Chen2022VelTypeRain, ρₐ)
+    (ai, bi, ci) = Chen2022_vel_coeffs_B1(velocity_params, ρₐ)
+    v_term(D) = sum(@. sum(ai * D^bi * exp(-ci * D)))
+    return v_term
+end
+liquid_particle_terminal_velocity(velocity_params::CMP.Chen2022VelType, ρₐ) =
+    liquid_particle_terminal_velocity(velocity_params.rain, ρₐ)
+
+function liquid_particle_terminal_velocity(velocity_params, ρₐ, D)
+    v_term = liquid_particle_terminal_velocity(velocity_params, ρₐ)
+    return v_term(D)
+end
+
+"""
     volume_sphere_D(D)
 
 Calculate the volume of a sphere with diameter D.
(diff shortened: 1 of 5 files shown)


haakon-e:
* **#543** <a href="https://app.graphite.dev/github/pr/CliMA/CloudMicrophysics.jl/543?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/CliMA/CloudMicrophysics.jl/543?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

haakon-e:
Questions:
1. For the wet growth rate. (From Musil, 1970):

> The rate at which a wet hailstone is able to transfer heat to its surroundings is
$$4 \pi R a \left( -KT + LD \Delta\rho_w \right)$$,
where the first term in the parentheses refers to heat conduction away from the stone and the second to evaporative cooling.

The $4\pi R$ term includes an implicit cancellation of $R$ from $T$[^1]. This expression assumes heat loss in a spherical particle. How would this be different with a spheroid (ellipsoid of revolution, oblate or prolate)? Does this matter? What is done in the literature?

[^1]: The cancellation comes from the fact that $T$ is the result of approximating heat conduction as `-K(T_air - T_surface) / R` at the surface of the (assumed spherical) ice particle. A similar approximation applies to the vapor term.

haakon-e:
To above: For now, let's stick with the formulation as-is. That is, be fine with Musil's derivation being based on spherical assumptions.

----

Another point:

In Morrison and Milbrandt 2015, they compute a bulk wet growth rate[^1] and compare this to a bulk dry growth rate to decide the partition between ice growth and shedding. We will instead do dry/wet rate comparisons at the ice-particle level. 

The expression for the wet growth rate `[kg/s]` is 

$$\frac{dL_{wet}}{dt} = \frac{2\pi D_{ice} a (-K_T \Delta T + L K_\rho \Delta \rho)}{L + C \Delta T}$$

the expression for the dry growth rate `[kg/s]` is

$$\frac{dL_{dry}}{dt} = \int E A | v_{ice} - v_{liq} | L_{liq} N\prime_{liq} dD_{liq}$$

This means that instead of doing the comparison:

$$\int \frac{dL_{wet}}{dt} N\prime_{ice} dD_{ice} < \int \frac{dL_{dry}}{dt} N\prime_{ice} dD_{ice}$$

as Morrison and Milbrandt 2015 do, I propose we do:

$$\frac{dL_{wet}}{dt} < \frac{dL_{dry}}{dt}$$

then integrate over $D_{ice}$[^2]. 

[^1]: I suspect they do it this way because the bulk wet rate can be computed once and stored in a lookup table for convenient evaluation in-simulation. We instead compute on-the-fly (eventually) on the GPU.
[^2]: On a technical level, `HCubature` can handle vector-valued integrands, so we can compute all bulk rates in one "swoop" with their `hcubature` method.

**Questions**:
1. Investigate in literature whether others have done this already
    - I doubt this since most others will want to use lookup tables, but good to check.
2. Will need to do calculations with cloud and rain at the same time for the thermodynamic argument to hold.
    - This is fine if we follow MM15's assumption that: "When the dry growth rate exceeds the wet growth rate the difference between the rates is shed as _1-mm-sized raindrops_" (emphasis mine). That is: all shedding moves into the _rain_ category.

PR #548, opened 2025-04-18
Get rid of FT in noneq cloud formation

trontrytel:
Hi @dennisYatunin . Here is a PR that fixes things for the two functions you see in the Jacobian.

I tried adding a test that checks the output type for all the functions in performance tests that I have. However I do have more complicated outputs than just FT:

```
 Expression: typeof(foo(args...)) == FT
   Evaluated: @NamedTuple{au::CloudMicrophysics.Microphysics2M.LiqRaiRates{Float64}, sc::Float64} == Float64

```
Do you see an easy way to include that sort of output? If not, thats fine, I'll undo  the performance tests change and just add the one for cloud formation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.22.10"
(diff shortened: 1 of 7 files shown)


dennisYatunin (approved):
Thank you for freeing atmos from the oppressive chains of FTs!

PR #551, opened 2025-04-23
Compute inverse lambda, safer cloud sources in non-eq.

trontrytel:
The (so far) learned lessons from debugging AMIP model configuration:
- Due to numerics it's easier to store inverse of lambda (size distribution parameter). Also, because of numerics I clip it at 1e-8. None of the tests in `CloudMicrophysics.jl` were affected by that change, so I'm planning to merge it as is. But I also intend to check if the clipping value affects the simulations in `Atmos`
- Only do cloud formation sources and sinks if there is water vapor or cloud present. This avoids some issues with cold temperatures high above, where we have no moisture anyway.

### DIFF
--- a/src/Microphysics2M.jl
+++ b/src/Microphysics2M.jl
@@ -579,8 +579,8 @@ function rain_terminal_velocity(
     λ = pdf_rain_parameters(pdf_r, q_rai, ρ, N_rai).λr
 
     # eq 20 from Chen et al 2022
-    vt0 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, λ, 0))
-    vt3 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, λ, 3))
+    vt0 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, 1 / λ, 0))
+    vt3 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, 1 / λ, 3))
 
     vt0 = N_rai < eps(FT) ? FT(0) : max(FT(0), vt0)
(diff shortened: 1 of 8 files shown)


src/Microphysics2M.jl
```diff
@@ -579,8 +579,8 @@ function rain_terminal_velocity(
     λ = pdf_rain_parameters(pdf_r, q_rai, ρ, N_rai).λr
 
     # eq 20 from Chen et al 2022
-    vt0 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, λ, 0))
-    vt3 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, λ, 3))
+    vt0 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, 1 / λ, 0))
+    vt3 = sum(CO.Chen2022_exponential_pdf.(aiu, bi, ciu, 1 / λ, 3))
```
amylu00, line 583: Just curious as to why it's `1 / λ` here but everywhere else, you call something of the form ` λ_inv = CM1.lambda_inverse(pdf, mass, q, ρ)`

amylu00 (approved):

src/Microphysics2M.jl
trontrytel: I changed the 1-moment microphysics to store the inverse lambda and following that I changed the Chen 2022 velocity parameterization to expect 1/lambda. However I left the 2-moment microphysics scheme unchanged, meaning that it still returns just plain old lambda. 

I didn't want to change 2M scheme without any reason. In case we need to change it, I'll do it when we couple it to Atmos.

PR #553, opened 2025-04-26
loss func adjusted to max frozen frac

amylu00:
## Purpose 

Loss function for AIDA ice nucleation calibrations switched from frozen fraction in last 25 seconds to max frozen fraction

## To-do

- adjust Gamma accordingly

## Content

----


papers/ice_nucleation_2024/plots.jl
```diff
@@ -46,29 +46,21 @@ function plot_AIDA_ICNC_data(
 end
 
 """
-    plot_calibrated_coeffs(batch_name, EKI_n_iterations, calibrated_ensemble_means)
+    plot_calibrated_coeffs(
+        batch_name,
+        UKI_n_iterations, UKI_n_ensembles, UKI_all_params,
+    )
```
trontrytel, line 52: The doc string below still says EKI

trontrytel (approved):

Issue #544, opened 2025-04-07
GPU error in `terminal_velocity`


charleskawczynski:
Reproducer:

```julia
FT = Float32
ENV["CLIMACOMMS_DEVICE"]="CUDA";
import ClimaComms
ClimaComms.@import_required_backends
import CloudMicrophysics
import CloudMicrophysics.Microphysics1M as CM1
function do_work(ᶜwᵣ, pr, rain, ρ, ρq_rai)
    @. ᶜwᵣ = CM1.terminal_velocity(
        pr,
        rain, #cmp.tv.rain,
        ρ,
        max(zero(ρ), ρq_rai / ρ),
    )
end

pr = CloudMicrophysics.Parameters.Rain{Float32, CloudMicrophysics.Parameters.ParticlePDFIceRain{Float32}, CloudMicrophysics.Parameters.ParticleMass{Float32}, CloudMicrophysics.Parameters.ParticleArea{Float32}, CloudMicrophysics.Parameters.Ventilation{Float32}, CloudMicrophysics.Parameters.Acnv1M{Float32}}(CloudMicrophysics.Parameters.ParticlePDFIceRain{Float32}(1.6f7), CloudMicrophysics.Parameters.ParticleMass{Float32}(0.001f0, 4.188791f-6, 3.0f0, 0.0f0, 1.0f0), CloudMicrophysics.Parameters.ParticleArea{Float32}(3.1415932f-6, 2.0f0, 0.0f0, 1.0f0), CloudMicrophysics.Parameters.Ventilation{Float32}(1.5f0, 0.53f0), CloudMicrophysics.Parameters.Acnv1M{Float32}(1000.0f0, 0.0005f0, 2.0f0), 0.001f0)
rain = CloudMicrophysics.Parameters.Chen2022VelTypeRain{Float32, 3}(0.115231f0, (0.044612f0, -0.263166f0, 4.7178f0), -0.47335f0, (2.2955f0, 2.2955f0, 1.1451f0), 0.038465f0, (0.0f0, 0.184325f0, 0.184325f0));

ρ = ones(FT, 3);
ᶜwᵣ = zeros(FT, 3);
ρq_rai = ones(FT, 3) .* FT(1e-6);
AT = ClimaComms.array_type(ClimaComms.device());
do_work(AT(ᶜwᵣ), pr, rain, AT(ρ), AT(ρq_rai))
```

PR #558, opened 2025-05-07
functionalizing particle operations

haakon-e:
## Purpose
This pull request introduces several improvements and refactorings to the `src/Common.jl`, `src/Microphysics2M.jl`, and `src/P3_terminal_velocity.jl` files, focusing on code clarity, reusability, and performance. Key changes include refactoring terminal velocity computations, improving documentation for functions, and introducing new utility functions for particle size distributions and ventilation factor.

A common pattern that is now introduced for functions that can be evaluated at specific diameters `D` is to have them return a function in `D` by default[^1]:
```julia
function particle_function(state, vel, aps)
    (a, b) = get_coeffs(state, vel, aps)
    func(D) = D^a * exp( - b * D)
end
function particle_function(state, vel, aps, D)
    func = particle_function(state, vel, aps) 
    return func(D)
end
```

### `Common.jl`
- Updated docs for Chen2022 Table coefficients + precompute `log` when it is used frequently.
- Added `liquid_particle_terminal_velocity` (replacement for, now removed `rain_particle_terminal_velocity` in `Microphysics2M.jl`) with an additional signature that returns a function, as described above.

### `Microphysics2M.jl`
- Removed `FT` where not necessary.
- Updated docstrings
- Added "functionalized" version of `size_distribution` (as above)
- Added function for ventilation factor (useful for a future PR)

### `P3_terminal_velocity.jl`
- Added "functionalized" version of `ice_particle_terminal_velocity` (as above)
- Removed `p3_particle_terminal_velocity` (will add back once we return to liquid fraction)
- Removed `velocity_difference` (Not really needed)

### `src/parameters/Microphysics2M.jl`
- Added a constant alias for the union of `RainParticlePDF_SB2006` and `RainParticlePDF_SB2006_limited`.

----

[^1]: We could consider not having the signature that evaluates directly, for simplicity. Either way is fine with me.

### DIFF
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -210,16 +210,15 @@ end
 """
     Chen2022_vel_coeffs_B1(coeffs, ρₐ)
 
- - coeffs - a struct with terminal velocity free parameters
- - ρₐ - air density
-
 Returns the coefficients from Table B1 Appendix B in Chen et al 2022
+
+# Arguments
+ - `coeffs`: a struct with terminal velocity free parameters
+ - `ρₐ`: air density [kg/m³]
+
 DOI: 10.1016/j.atmosres.2022.106171
 """
-function Chen2022_vel_coeffs_B1(
-    coeffs::CMP.Chen2022VelTypeRain,
-    ρₐ::FT,
-) where {FT}
+function Chen2022_vel_coeffs_B1(coeffs::CMP.Chen2022VelTypeRain, ρₐ)
     (; ρ0, a, a3_pow, b, b_ρ, c) = coeffs
     # Table B1
     q = exp(ρ0 * ρₐ)
@@ -235,11 +234,13 @@ end
 """
     Chen2022_vel_coeffs_B2(coeffs, ρₐ, ρᵢ)
 
- - coeffs - a struct with terminal velocity free parameters
- - ρₐ - air density
- - ρᵢ - apparent density of ice particles
-
 Returns the coefficients from Table B2 Appendix B in Chen et al 2022
+
+# Arguments
+ - `coeffs`: a struct with terminal velocity free parameters
+ - `ρₐ`: air density [kg/m³]
+ - `ρᵢ`: apparent density of ice particles [kg/m³]
+
 DOI: 10.1016/j.atmosres.2022.106171
 """
 function Chen2022_vel_coeffs_B2(
@@ -249,12 +250,13 @@ function Chen2022_vel_coeffs_B2(
 ) where {FT}
     (; A, B, C, E, F, G) = coeffs
     # Table B3
-    As = A[2] * (log(ρᵢ))^2 − A[3] * log(ρᵢ) + A[1]
-    Bs = FT(1) / (B[1] + B[2] * log(ρᵢ) + B[3] / sqrt(ρᵢ))
+    log_ρᵢ = log(ρᵢ)
+    As = A[2] * log_ρᵢ^2 − A[3] * log_ρᵢ + A[1]
+    Bs = FT(1) / (B[1] + B[2] * log_ρᵢ + B[3] / sqrt(ρᵢ))
     Cs = C[1] + C[2] * exp(C[3] * ρᵢ) + C[4] * sqrt(ρᵢ)
-    Es = E[1] - E[2] * (log(ρᵢ))^2 + E[3] * sqrt(ρᵢ)
-    Fs = -exp(F[1] - F[2] * (log(ρᵢ))^2 + F[3] * log(ρᵢ))
-    Gs = FT(1) / (G[1] + G[2] / (log(ρᵢ)) - G[3] * log(ρᵢ) / ρᵢ)
+    Es = E[1] - E[2] * log_ρᵢ^2 + E[3] * sqrt(ρᵢ)
+    Fs = -exp(F[1] - F[2] * log_ρᵢ^2 + F[3] * log_ρᵢ)
+    Gs = FT(1) / (G[1] + G[2] / log_ρᵢ - G[3] * log_ρᵢ / ρᵢ)
     # Table B2
     ai = (Es * ρₐ^As, Fs * ρₐ^As)
     bi = (Bs + ρₐ * Cs, Bs + ρₐ * Cs)
@@ -268,11 +270,13 @@ end
 """
     Chen2022_vel_coeffs_B4(coeffs, ρₐ, ρᵢ)
 
- - coeffs - a struct with terminal velocity free parameters
- - ρₐ - air density
- - ρᵢ - apparent density of ice particles
-
 Returns the coefficients from Table B4 Appendix B in Chen et al 2022
+
+# Arguments
+ - `coeffs`: a struct with terminal velocity free parameters
+ - `ρₐ`: air density [kg/m³]
+ - `ρᵢ`: apparent density of ice particles [kg/m³]
+
 DOI: 10.1016/j.atmosres.2022.106171
 """
 function Chen2022_vel_coeffs_B4(
@@ -282,13 +286,14 @@ function Chen2022_vel_coeffs_B4(
 ) where {FT}
     (; A, B, C, E, F, G, H) = coeffs
     # Table B5
-    Al = A[1] + A[2] * log(ρᵢ) + A[3] * (ρᵢ)^FT(-3 / 2)
-    Bl = exp(B[1] + B[2] * log(ρᵢ)^2 + B[3] * log(ρᵢ))
-    Cl = exp(C[1] + C[2] / log(ρᵢ) + C[3] / ρᵢ)
-    El = E[1] + E[2] * log(ρᵢ) * sqrt(ρᵢ) + E[3] * sqrt(ρᵢ)
-    Fl = F[1] + F[2] * log(ρᵢ) - exp(log(-F[3]) - ρᵢ)
-    Gl = (G[1] + G[2] * log(ρᵢ) * sqrt(ρᵢ) + G[3] / sqrt(ρᵢ))^(-1)
-    Hl = H[1] + H[2] * (ρᵢ)^FT(5 / 2) + exp(log(-H[3]) - ρᵢ)
+    log_ρᵢ = log(ρᵢ)
+    Al = A[1] + A[2] * log_ρᵢ + A[3] * ρᵢ^FT(-3 / 2)
+    Bl = exp(B[1] + B[2] * log_ρᵢ^2 + B[3] * log_ρᵢ)
+    Cl = exp(C[1] + C[2] / log_ρᵢ + C[3] / ρᵢ)
+    El = E[1] + E[2] * log_ρᵢ * sqrt(ρᵢ) + E[3] * sqrt(ρᵢ)
+    Fl = F[1] + F[2] * log_ρᵢ - exp(log(-F[3]) - ρᵢ)
+    Gl = (G[1] + G[2] * log_ρᵢ * sqrt(ρᵢ) + G[3] / sqrt(ρᵢ))^(-1)
+    Hl = H[1] + H[2] * ρᵢ^FT(5 / 2) + exp(log(-H[3]) - ρᵢ)
     # Table B4
     ai = (Bl * ρₐ^Al, El * ρₐ^Al * exp(Hl * ρₐ))
     bi = (Cl, Fl)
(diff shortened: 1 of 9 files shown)


trontrytel:
Looks like some tests are broken:

```
2M_microphysics - Seifert and Beheng 2006 cloud distribution sanity checks: Test Failed at /Users/runner/work/CloudMicrophysics.jl/CloudMicrophysics.jl/test/microphysics2M_tests.jl:752
  Expression: ≈(ND_bounded, Nₗ, rtol = FT(1.0e-7))
   Evaluated: 0.0 ≈ 1.0e9 (rtol=1.0e-7)
```

src/Common.jl
```diff
@@ -330,6 +335,39 @@ function Chen2022_exponential_pdf(a::FT, b::FT, c::FT, λ_inv::FT, k::Int) where
...
+- The first method returns a function of `D`, while the second evaluates at a specific `D`.
+
+Needed for numerical integrals in the P3 scheme.
+
+!!! note
+    We use the same terminal velocity parametrization for cloud and rain water.
+"""
+function liquid_particle_terminal_velocity(velocity::CMP.Chen2022VelTypeRain, ρₐ)
+    (ai, bi, ci) = Chen2022_vel_coeffs_B1(velocity, ρₐ)
+    v_term(D) = sum(@. sum(ai * D^bi * exp(-ci * D)))
```
trontrytel, line 360: Will this compile on the GPUs in Atmos? I remember fixing something similar here:
https://github.com/CliMA/CloudMicrophysics.jl/pull/545/files

trontrytel: Here is the issue for it: https://github.com/CliMA/CloudMicrophysics.jl/issues/544

src/Common.jl
```diff
@@ -330,6 +335,39 @@ function Chen2022_exponential_pdf(a::FT, b::FT, c::FT, λ_inv::FT, k::Int) where
...
+    return v_term
+end
+liquid_particle_terminal_velocity(velocity::CMP.Chen2022VelType, ρₐ) =
```
trontrytel, line 363: Why do we need this wrapper?

From my experience of working with Thermodynamics package, once you define a bunch of function wrappers that all have the same name and call each other depending on their argument type and number, it is very difficult to see which function is actually being used in the code. I really woudn't mind keeping a different names between per-particle and bulk velocities

src/Common.jl
```diff
@@ -330,6 +335,39 @@ function Chen2022_exponential_pdf(a::FT, b::FT, c::FT, λ_inv::FT, k::Int) where
...
 
+"""
+    liquid_particle_terminal_velocity(velocity, ρₐ)
+    liquid_particle_terminal_velocity(velocity, ρₐ, D)
+
+Compute the terminal velocity of a liquid particle as a function of its size 
+    (maximum dimension, `D`) using the Chen 2022 parametrization.
+
+# Arguments
+- `velocity`: a struct with terminal velocity parameters from Chen 2022
```
trontrytel, line 346: I would vote to keep the naming convention of params, coefficients, etc. Maybe `velocity_params` or something like that if you want to be more precise?

src/Common.jl
```diff
@@ -330,6 +335,39 @@ function Chen2022_exponential_pdf(a::FT, b::FT, c::FT, λ_inv::FT, k::Int) where
     return a * exp(δ * log(1 / λ_inv) - (b + δ) * log(1 / λ_inv + c)) * SF.gamma(b + δ) / SF.gamma(δ)
 end
 
+"""
+    liquid_particle_terminal_velocity(velocity, ρₐ)
+    liquid_particle_terminal_velocity(velocity, ρₐ, D)
+
+Compute the terminal velocity of a liquid particle as a function of its size
```
trontrytel, line 342: Right now it's either that, or compute the bulk velocity?

src/Microphysics2M.jl
```diff
@@ -65,10 +65,7 @@ end
     assumed cloud droplet mass distribution.
 """
 function pdf_cloud_parameters(
-    pdf_c::CMP.CloudParticlePDF_SB2006{FT},
-    qₗ::FT,
-    ρₐ::FT,
-    Nₗ::FT,
+    pdf_c::CMP.CloudParticlePDF_SB2006{FT}, qₗ, ρₐ, Nₗ,
```
trontrytel, line 68: At some point I was told that it's more efficient to put the thing you are dispatching over first. Thats why we typically put parameter structs first in the argument list. Hopefully we don't have to change that?

src/Microphysics2M.jl
```diff
@@ -159,36 +150,35 @@ function pdf_rain_parameters(
...
-    particle size
+# Arguments
+- `pdf`: Struct containing size distribution parameters of cloud or rain
+- `q`: Cloud or rain water specific content [kg/kg]
+- `ρₐ`: Density of air [kg/m³]
+- `N`: Cloud or rain water number concentration [1/m³]
+- `D`: Particle size (i.e. maximum dimension of particle) [m]
+
+# Returns
+- The first method returns a function that computes the size distribution value
```
trontrytel, line 166: Could we define the function separately and use that to compute the value? Whats the benefit for having a separate function that returns a function?

src/Microphysics2M.jl
```diff
@@ -159,36 +150,35 @@ function pdf_rain_parameters(
...
-    pdf::Union{
-        CMP.RainParticlePDF_SB2006{FT},
-        CMP.RainParticlePDF_SB2006_limited{FT},
-    },
-    D::FT,
-    q::FT,
-    ρₐ::FT,
-    N::FT,
-) where {FT}
+function size_distribution(pdf::CMP.RainParticlePDF_SB2006_MaybeLimited, q, ρₐ, N, D)
```
trontrytel, line 171: I don't think that maybe limited is a good name. We can just pass here a `Union` of the two SB PDF options.

src/P3_particle_properties.jl
```diff
@@ -479,6 +479,34 @@ function ϕᵢ(state::P3State, D)
...
+function ventilation_factor(state, vel, ρₐ, aps)
+    (; ν_air, D_vapor) = aps
+    (; vent_a, vent_b) = state.params.vent
+    N_sc = ν_air / D_vapor
+    v_term = ice_particle_terminal_velocity(state, vel, ρₐ)
+
+    # Reynolds number
+    N_Re(D) = D * v_term(D) / ν_air
+    # Ventilation factor
+    vent_factor(D) = vent_a + vent_b * cbrt(N_sc) * sqrt(N_Re(D))
```
trontrytel, line 506: Is the same function being used in 1M scheme? Could we move it to common and re-use here?

src/P3_processes.jl
```diff
@@ -76,31 +76,23 @@ function ice_melt(
...
+    (; T_freeze) = state.params
 
-    # Ice particle terminal velocity
-    v(D) = ice_particle_terminal_velocity(state, D, Chen2022, ρₐ)
-    # Reynolds number
-    N_Re(D) = D * v(D) / ν_air
-    # Ventillation factor
-    (; vent_a, vent_b) = params.vent
-    F_v(D) = vent_a + vent_b * N_sc^FT(1 / 3) * N_Re(D)^FT(1 / 2)
+    F_v = ventilation_factor(state, Chen2022, ρₐ, aps)
```
trontrytel, line 85: Maybe `P3_state` instead of `state`? I still find it confusing given that we use `state` for the whole model state in Atmos

src/P3_terminal_velocity.jl
```diff
@@ -1,141 +1,59 @@
+
 """
-    ice_particle_terminal_velocity(state, D, Chen2022, ρₐ, use_aspect_ratio)
-Returns the terminal velocity of a single ice particle as a function
-    of its size (maximum dimension, D) using the Chen 2022 parametrization.
+    ice_particle_terminal_velocity(state, velocity, ρₐ; use_aspect_ratio)
```
trontrytel, line 3: Same as above. I don't understand why we need two options, instead of just having the function defined and then using it to obtain a value.

src/P3_terminal_velocity.jl
```diff
@@ -1,141 +1,59 @@
...
-        (ai, bi, ci) = CO.Chen2022_vel_coeffs_B2(Chen2022.small_ice, ρₐ, ρᵢ)
-    else
-        (ai, bi, ci) = CO.Chen2022_vel_coeffs_B4(Chen2022.large_ice, ρₐ, ρᵢ)
+    state::P3State, velocity::CMP.Chen2022VelType, ρₐ; use_aspect_ratio = true,
+)
+    function v_term(D::FT) where {FT}
+        (; small_ice, large_ice) = velocity
+        ρᵢ = FT(916.7) # ρᵢ = p3_density(p3, D, F_rim, th) # TODO: tmp
+        ϕ_factor = use_aspect_ratio ? cbrt(ϕᵢ(state, D)) : FT(1)
+        (ai, bi, ci) = if D <= small_ice.cutoff
```
trontrytel, line 28: Is that better than what it was before? Or just your personal style preference?

src/parameters/Microphysics2M.jl
```diff
@@ -419,6 +419,11 @@ function RainParticlePDF_SB2006(td::CP.AbstractTOMLDict)
     return RainParticlePDF_SB2006{FT}(; parameters...)
 end
 
+const RainParticlePDF_SB2006_MaybeLimited{FT} = Union{
```
trontrytel, line 422: I don't think it improves code readability. I would vote to just pass the union when needed

test/microphysics2M_tests.jl
```diff
@@ -615,37 +615,20 @@ function test_microphysics2M(FT)
...
-        D_max_limited =
-            CM2.get_size_distribution_bound(SB2006.pdf_r, qᵣ, Nᵣ, ρₐ, eps(FT))
+        D_max_lim = CM2.get_size_distribution_bound(SB2006.pdf_r, qᵣ, Nᵣ, ρₐ, eps(FT))
         D_max = CM2.get_size_distribution_bound(
-            SB2006_no_limiters.pdf_r,
-            qᵣ,
-            Nᵣ,
-            ρₐ,
-            eps(FT),
+            SB2006_no_limiters.pdf_r, qᵣ, Nᵣ, ρₐ, eps(FT),
```
trontrytel, line 620: We want the dispatch type first

src/Microphysics2M.jl
haakon-e: The primary benefit, as I see it, is that I can create a reference to the size distribution once. In this case, that includes precomputing the line `(; N₀r, λr) = pdf_rain_parameters(pdf, q, ρₐ, N)`. Then I can evaluate it many times during integration by only passing in the diameter. As I see it, it leads to code that more closely resembles the equations from the relevant papers.

We already implicitly do this when we write things like:
```julia
x -> CM2.size_distribution(
    SB2006_no_limiters.pdf_r,
    FT(x),
    qᵣ,
    ρₐ,
    Nᵣ,
)
```
which defines an anonymous function in `x`. The idea with this PR is to make code like that more explicit and reusable.

src/P3_particle_properties.jl
haakon-e: Sounds good.

src/P3_processes.jl
haakon-e: ok, I'll revert to p3_state.

src/P3_terminal_velocity.jl
haakon-e: See above, but can abandon this if it seems too confusing/not useful/hard to maintain.

src/P3_terminal_velocity.jl
haakon-e: Let's chat about what the "what" you're referring to is here.

src/P3_terminal_velocity.jl
```diff
@@ -1,141 +1,59 @@
...
-        ),
-    )
+function ice_particle_terminal_velocity(state, velocity, ρₐ, D; use_aspect_ratio = true)
+    v_term = ice_particle_terminal_velocity(state, velocity, ρₐ; use_aspect_ratio)
+    return v_term(D)
 end
 
 """
-    ice_terminal_velocity(dist, Chen2022, ρₐ, use_aspect_ratio)
+    ice_terminal_velocity(dist, velocity, ρₐ, use_aspect_ratio)
```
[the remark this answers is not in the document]
haakon-e: sure.

src/parameters/Microphysics2M.jl
haakon-e: Ok, I'll revert.

test/microphysics2M_tests.jl
haakon-e: I don't understand this comment.

src/Common.jl
haakon-e: We should add some tests for this, somehow.

haakon-e (commented):
Thanks for your feedback! I appreciate it :) 
Let's chat when we both have a minute.

src/Microphysics2M.jl
haakon-e: git diff was weird. leave as is.

src/Common.jl
haakon-e: Follow pattern from here:
https://github.com/CliMA/CloudMicrophysics.jl/blob/main/test/gpu_clima_core_test.jl#L97-L105

PR #561, opened 2025-05-08
Sinks for evaporation and sublimation in parcel

amylu00:
- added number sinks to parcel from evaporation and sublimation
- initial number of droplets for HOM exps is observed N_l + N_a / turn aerosol activation off


trontrytel (approved):
Thank you!

PR #562, opened 2025-05-09
Tiny math rendering improvement: \rm{max}

glwagner:
Some people like to keep their words non-italics, who knows why


trontrytel:
Blasphemy!

PR #567, opened 2025-05-20
Add more interface options to 0M

trontrytel:


src/Microphysics0M.jl
```diff
@@ -16,28 +16,39 @@ export remove_precipitation
...
 The threshold for when to remove `q_tot` is defined either by the
 condensate specific content or supersaturation.
-The thresholds and the relaxation timescale are defined in
-ClimaParams.
+The thresholds and the relaxation timescale are defined `Parameters0M` struct.
 """
 remove_precipitation(
     (; τ_precip, qc_0)::CMP.Parameters0M,
     q::TD.PhasePartition,
 ) = -max(0, (q.liq + q.ice - qc_0)) / τ_precip
```
glwagner, line 35:
```suggestion
remove_precipitation(
    p::CMP.Parameters0M,
    q::TD.PhasePartition,
) = remove_precipitation(p, q.liq, q.ice)
```

could also use this to reduce code paths

glwagner: + reduce need for additional testing

trontrytel: Sure, that also works. Tbh, I think I can/will get rid of `PhasePartition` on the `Atmos` side and just support the simple version of passing in individual tracers.

PR #575, opened 2025-05-29
Refactor: ice terminal velocity method signature

haakon-e:
- Rewrote `ice_particle_terminal_velocity` as a function that returns `v_term(D)`: a 1-argument function in diameter `D`. Very useful for composing with other functions for numerical integration.
- Refactored `ice_terminal_velocity` to follow the new form of `ice_particle_terminal_velocity`

PR #576, opened 2025-05-29
Refactor: liquid (cloud/rain) terminal velocity

haakon-e:
- Renamed `rain_particle_terminal_velocity` to `liquid_particle_terminal_velocity` since it is currently used for both cloud and rain.
- and moved it to Common to reflect it's broader applicability.


haakon-e:

- #575
    - #576 ◀

sajjadazimi (commented):
Maybe `particle_terminal_velocity` might be a better name since it can apparently be used for ice too. `liquid` can be confusing because we normally use it to refer to cloud liquid particles.

PR #578, opened 2025-05-29
refactor: set P3 integral bounds using quantiles

haakon-e:
## Purpose 

Use distribution quantiles to set the integration bounds for the P3 size distribution.

The existing approach was to integrate from 0 to 1, including P3 thresholds.
To achieve higher accuracy, we passed a high `order` and/or low `rtol` to `quadgk`.

This PR uses the `p`-th and `1-p`-th quantiles of the size distribution to set the integration bounds.
This is not likely more accurate, but it is more convenient and likely more computationally efficient.

## Content

----


haakon-e:

- #575
    - #576
        - #578 ◀

PR #581, opened 2025-06-05
Remove nonequilibrium phase partition dependency

oalcabes:
Remove nonequilibirum phase partition dependency

### DIFF
--- a/src/MicrophysicsNonEq.jl
+++ b/src/MicrophysicsNonEq.jl
@@ -83,10 +83,8 @@ function conv_q_vap_to_q_liq_ice_MM2015(
         Rᵥ = TD.Parameters.R_v(tps)
         Lᵥ = TD.latent_heat_vapor(tps, T)
 
-        q = TD.PhasePartition(q_tot, q_liq + q_rai, q_ice + q_sno)
-        #cₚ_air = TD.cp_m(tps, q_tot, q_liq + q_rai, q_ice + q_sno)
-        cₚ_air = TD.cp_m(tps, q)
-        qᵥ = TD.vapor_specific_humidity(q)
+        qᵥ = q_tot - q_liq - q_ice - q_rai - q_sno
+        cₚ_air = TD.cp_m(tps, q_tot, q_liq + q_rai, q_ice + q_sno)
 
         pᵥ_sat_liq = TD.saturation_vapor_pressure(tps, T, TD.Liquid())
         qᵥ_sat_liq = TD.q_vap_saturation_from_density(tps, T, ρ, pᵥ_sat_liq)
@@ -115,10 +113,8 @@ function conv_q_vap_to_q_liq_ice_MM2015(
         Rᵥ = TD.Parameters.R_v(tps)
         Lₛ = TD.latent_heat_sublim(tps, T)
 
-        q = TD.PhasePartition(q_tot, q_liq + q_rai, q_ice + q_sno)
-        #cₚ_air = TD.cp_m(tps, q_tot, q_liq + q_rai, q_ice + q_sno)
-        cₚ_air = TD.cp_m(tps, q)
-        qᵥ = TD.vapor_specific_humidity(q)
+        qᵥ = q_tot - q_liq - q_ice - q_rai - q_sno
+        cₚ_air = TD.cp_m(tps, q_tot, q_liq + q_rai, q_ice + q_sno)
 
         pᵥ_sat_ice = TD.saturation_vapor_pressure(tps, T, TD.Ice())
         qᵥ_sat_ice = TD.q_vap_saturation_from_density(tps, T, ρ, pᵥ_sat_ice)


src/MicrophysicsNonEq.jl
```diff
@@ -76,18 +76,20 @@ The formulation is based on Morrison and Grabowski 2008 and
 Morrison and Milbrandt 2015
 """
 function conv_q_vap_to_q_liq_ice_MM2015(
-    (; τ_relax)::CMP.CloudLiquid,
+    (; τ_relax)::CMP.CloudLiquid{FT},
```
trontrytel, line 79: FT is not defined here. I think thats why none of the tests are passing right now with this PR

src/MicrophysicsNonEq.jl
```diff
@@ -101,18 +103,38 @@ function conv_q_vap_to_q_liq_ice_MM2015(
     end
 end
 function conv_q_vap_to_q_liq_ice_MM2015(
-    (; τ_relax)::CMP.CloudIce,
+    (; τ_relax)::CMP.CloudLiquid{FT},
```
trontrytel, line 106: Same as the above

parcel/ParcelTendencies.jl
```diff
@@ -311,9 +313,9 @@ function deposition(params::NonEqDepParams, PSD, state, ρ_air)
 
     (; tps, ice, dt) = params
 
-    q = TD.PhasePartition(qᵥ + qₗ + qᵢ, qₗ, qᵢ)
+    qₜ = qᵥ + qₗ + qᵢ
 
-    dep_rate = MNE.conv_q_vap_to_q_liq_ice_MM2015(ice, tps, q, ρ_air, T)
+    dep_rate = MNE.conv_q_vap_to_q_liq_ice_MM2015(ice, tps, qₜ, qₗ, qᵢ, ρ_air, T)
```
trontrytel, line 318: You chose to return the unlimited tendency for condensation but you are keeping the old limiters for deposition?

src/MicrophysicsNonEq.jl
```diff
@@ -101,18 +103,38 @@ function conv_q_vap_to_q_liq_ice_MM2015(
...
     tps::TDP.ThermodynamicsParameters,
     q::TD.PhasePartition,
     ρ,
     T,
 )
+
+    override_file = Dict(
```
trontrytel, line 113: Why do you need to create and override any dictionaries here? You already get the timescale as one of the arguments

src/MicrophysicsNonEq.jl
trontrytel, line 106: This should be for `CloudIce`?

src/MicrophysicsNonEq.jl
```diff
@@ -101,18 +103,38 @@ function conv_q_vap_to_q_liq_ice_MM2015(
...
+        "condensation_evaporation_timescale" =>
+            Dict("value" => τ, "type" => "float"),
+    )
+    liquid_toml_dict = CP.create_toml_dict(FT; override_file)
+    liquid = CMP.CloudLiquid(liquid_toml_dict)
+
+    conv_q_vap_to_q_liq_ice_MM2015(liquid, tps, q.tot, q.liq, q.ice, ρ, T)
+
+end
+function conv_q_vap_to_q_liq_ice_MM2015(
```
trontrytel, line 123: Could you provide wrapper functions that still accept `PhasePartition` as argument for both liquid and ice?

test/gpu_tests.jl
```diff
@@ -119,7 +119,9 @@ end
         output[1, i] = CMN.conv_q_vap_to_q_liq_ice_MM2015(
             liquid,
             tps,
-            TD.PhasePartition(qᵥ_sl[i]),
```
trontrytel, line 122: Could we keep both the `PhasePartition` and the new version in the GPU tests?

test/microphysics_noneq_tests.jl
```diff
@@ -66,25 +66,25 @@ function test_microphysics_noneq(FT)
         pᵥ_si = TD.saturation_vapor_pressure(tps, T, TD.Ice())
         qᵥ_si = TD.q_vap_saturation_from_density(tps, T, ρ, pᵥ_si)
 
-        qₚ(qᵥ) = TD.PhasePartition(FT(qᵥ))
+        #qₚ(qᵥ) = TD.PhasePartition(FT(qᵥ))
```
trontrytel, line 69: Could we keep both the new one and `PhasePartition` one in tests? Or if you want to have one, then I would keep the `PhasePartition` one. Since internally it should be calling the new version

test/performance_tests.jl
```diff
@@ -284,7 +284,7 @@ function benchmark_test(FT)
     bench_press(
         FT,
         CMN.conv_q_vap_to_q_liq_ice_MM2015,
-        (liquid, tps, TD.PhasePartition(FT(0.00145)), FT(0.8), FT(263)),
+        (liquid, tps, FT(0.00145), FT(0), FT(0), FT(0.8), FT(263)),
```
trontrytel, line 287: Same as for the other tests - either lets keep the `PhasePartition` version (as it calls the new one) or both

oalcabes: thanks for all the notes and for catching some of the typos! i shall add

src/MicrophysicsNonEq.jl
oalcabes: nvm i am dumb

trontrytel (approved):
LGTM! Feel free to merge it in if you would like. If not I'll include it in my thermodynamics refactor PR

PR #582, opened 2025-06-06
rft: remove P3Distribution struct

haakon-e:
## Purpose 

This PR makes necessary changes to have P3 sources be compatible with broadcast expressions in `ClimaAtmos.jl`. Notably, I completely remove the `P3Distribution` struct, which was intended to simplify (and reduce errors) in internal calculations. Unfortunately, because this struct bundled the `P3State`, which bundled the `CMP.ParametersP3`, a broadcast expression that would find the P3 distribution parameters would store unnecessarily much data (notably copies of all the P3 parameters).

Further, because `F_rim` in `P3State` is computed from `L_rim / L_ice`, it's a bit odd to store `L_ice` in `P3Distribution` and `F_rim` in `P3State`. Thus, it makes more sense for `L_ice` (and `N_ice`) to live in `P3State` (they are, after all, the prognostic "P3 State" variables). This leaves us with only `log_λ` and `log_N₀` left in `P3Distribution`. Since the latter is computed from the former, it can also be left out. Thus we're left with little reason to have a `P3Distribution` struct at all.

The new setup is thus to store `params`, `L_ice`, `N_ice`, `F_rim`, `ρ_rim` in `P3State`. The `P3State` is only used internally, and should not be included in any "external API". There should be minimal overhead by using this struct instead of separate arguments to internal functions.

For distribution-dependent functions, we now pass `state, logλ` where we previously passed `dist` (`P3Distribution`).

## Content

P3_particle_properties.jl:
- move `L` and `N` to `P3State` struct (now called `L_ice` and `N_ice`)
- update `get_state` to take `L_ice` and `N_ice` as keyword arguments
- remove `get_parameters`, `threshold_tuple`
- add `get_segments`. used for analytical integration over mass
P3_size_distribution.jl:
- rename `log∫DⁿmN′dD` -> `logmass_gamma_moment`
    - refactored to fetch mass coeffs using `ice_mass_coeffs`
- rename `log_integrate_moment_psd` -> `loggamma_inc_moment`
- rename `log_L_div_N` -> `logLdivN`
- remove `log_L_div_N₀`, `log_N_div_N₀` (see `logLdivN` for calculation)
- remove `P3Distribution` struct
  - implies that these functions now take `state, logλ` as arguments:
    `logN′ice`, `N′ice`, `get_μ`, `logmass_gamma_moment`, `logLdivN`,
    `∫fdD`, `integral_bounds`, `D_m`, `ice_terminal_velocity_number_weighted`,
    `ice_terminal_velocity_mass_weighted`
- remove `get_distribution_parameters` (returned `P3Distribution` struct)
- replaced by `get_distribution_logλ` (now returns `logλ`, Float value)
- similarly, replace `get_distribution_parameters_all_solutions` by
    `get_distribution_logλ_all_solutions` (now returns vector of `logλ` values)
Elsewhere:
- update docs / docs-plots
- update tests

----

### DIFF
--- a/src/P3_processes.jl
+++ b/src/P3_processes.jl
@@ -47,10 +47,11 @@ function het_ice_nucleation(
 end
 
 """
-    ice_melt(p3, Chen2022, aps, tps, Tₐ, ρₐ, dt; ∫kwargs...)
+    ice_melt(state, logλ, Chen2022, aps, tps, Tₐ, ρₐ, dt; ∫kwargs...)
 
 # Arguments
- - `dist`: a [`P3Distribution`](@ref) object
+ - `state`: a [`P3State`](@ref) object
+ - `logλ`: the log of the slope parameter [log(1/m)]
  - `Chen2022`: struct containing Chen 2022 velocity parameters
  - `aps`: air properties
  - `tps`: thermodynamics parameters
@@ -64,7 +65,7 @@ end
 Returns the melting rate of ice (QIMLT in Morrison and Mildbrandt (2015)).
 """
 function ice_melt(
-    dist::P3Distribution, Chen2022::CMP.Chen2022VelType,
+    state::P3State, logλ, Chen2022::CMP.Chen2022VelType,
     aps::CMP.AirProperties, tps::TDP.ThermodynamicsParameters,
     Tₐ, ρₐ, dt;
     ∫kwargs = (;),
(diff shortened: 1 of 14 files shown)


src/P3_processes.jl
```diff
@@ -75,25 +76,26 @@ function ice_melt(
...
     # only consider melting (not fusion)
     dLdt = max(0, dLdt)
     # compute change of N_ice proportional to change in L
-    dNdt = N / L * dLdt
+    dNdt = N_ice / L_ice * dLdt
 
     # ... and don't exceed the available number and mass of water droplets
-    dNdt = min(dNdt, N / dt)
-    dLdt = min(dLdt, L / dt)
+    dNdt = min(dNdt, N_ice / dt)  # TODO: Apply limiters in CA.jl
```
trontrytel, line 98: Do you want to keep that limiter here?

trontrytel (approved):

src/P3_processes.jl
haakon-e: No, I don't think I really want any limiters in CM, but I'll go through and remove them in a future PR (maybe even the next one!)

PR #585, opened 2025-06-10
Update non-equilibrium microphysics API

trontrytel:
This PR adds an option to call non-equilibrium cloud formation without using the `TD.PhasePartition`. This is a breaking change for the API.

### DIFF
--- a/src/MicrophysicsNonEq.jl
+++ b/src/MicrophysicsNonEq.jl
@@ -31,41 +31,33 @@ deposition of cloud ice.
 τ_relax(p::CMP.CloudIce) = p.τ_relax
 
 """
-    conv_q_vap_to_q_liq_ice(liquid, q_sat, q)
-    conv_q_vap_to_q_liq_ice(ice, q_sat, q)
+    conv_q_vap_to_q_liq_ice(params, q_sat, q_)
 
- - `liquid` or `ice` - a struct with cloud water or ice free parameters
- - `q_sat` - PhasePartition at equilibrium
- - `q` - current PhasePartition
+ - `params` - a struct with cloud water or ice free parameters
+ - `q_sat` - liquid or ice equilibrium specific contents
+ - `q_`` - current liquid or ice specific contants
 
 Returns the cloud water tendency due to condensation and evaporation
 or cloud ice tendency due to sublimation and vapor deposition.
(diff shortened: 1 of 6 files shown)


trontrytel:
Not sure why, but the tests in the CI are running with the v0.23 version. I'll fix it tomorrow

haakon-e (approved):
Looks very reasonable to me!

src/MicrophysicsNonEq.jl
```diff
@@ -31,41 +31,34 @@ deposition of cloud ice.
 τ_relax(p::CMP.CloudIce) = p.τ_relax
 
 """
-    conv_q_vap_to_q_liq_ice(liquid, q_sat, q)
-    conv_q_vap_to_q_liq_ice(ice, q_sat, q)
+    conv_q_vap_to_q_liq_ice(params, q_sat, q_)
+    conv_q_vap_to_q_liq_ice(params, q_sat::PhasePartition, q_::PhasePartition)
```
sajjadazimi, line 35: I don't find the implementation of this function (with phase partition)!

trontrytel: Thanks for catching it! I decided to keep it for now, just in case we need something simple for testing. But I deleted the option to pass in `PhasePartition`

src/MicrophysicsNonEq.jl
```diff
@@ -78,15 +71,22 @@ Morrison and Milbrandt 2015
 function conv_q_vap_to_q_liq_ice_MM2015(
     (; τ_relax)::CMP.CloudLiquid,
     tps::TDP.ThermodynamicsParameters,
-    q::TD.PhasePartition,
+    q_tot,
+    q_liq,
+    q_ice,
+    q_rai,
+    q_sno,
```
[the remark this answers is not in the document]
trontrytel: Agreed! I followed the pattern from 1-moment microphysics and Atmos, where I assumed I have those four variables. But from the point of view of the non-equilibrium module this is not necessary. 

One tiny detail is that right now I compute any tendency if I have vapor or cloud condensate present (excluding precipitation). This (I think) makes sense from the 1-moment microphysics rain formation point of view, where rain rain evaporation and snow sublimation are handled by different terms. 

So I would vote to keep it as is for now and revisit once we have a single category scheme for warm rain processes

Issue #572, opened 2025-05-28
`ExampleNonEq.jl` crashes in Julia 1.11.5


haakon-e:
Run from `docs` environment:
```Julia
julia> include("../parcel/Example_NonEq.jl")
┌ Info: using τ value 
└   τ = 10.0
┌ Info: relaxations:
│   liquid.τ_relax = 10.0
└   ice.τ_relax = 10.0
┌ Warning: Using arrays or dicts to store parameters of different types can hurt performance.
│ Consider using tuples instead.
└ @ SciMLBase ~/.julia/packages/SciMLBase/Ha7rZ/src/performance_warnings.jl:33

[1133] signal 11 (2): Segmentation fault: 11
in expression starting at ~/CloudMicrophysics.jl/parcel/Example_NonEq.jl:85
_ZL13getSameOpcodeN4llvm8ArrayRefIPNS_5ValueEEERKNS_17TargetLibraryInfoEj at ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/lib/julia/libLLVM.dylib (unknown line)
Allocations: 38727004 (Pool: 38725628; Big: 1376); GC: 76
zsh: segmentation fault  julia
```

haakon-e:
I traced it to running
```Julia
ODE.solve(
        problem,
        ODE.Euler(),
        dt = pp.const_dt,
        reltol = 100 * eps(FT),
        abstol = 100 * eps(FT),
    )
```
within `CM.jl/parcel/ParcelModel.jl:L344`:
```Julia
infil> ODE.solve(
               problem,
               ODE.Euler(),
               dt = pp.const_dt,
               reltol = 100 * eps(FT),
               abstol = 100 * eps(FT),
           )

[3019] signal 11 (2): Segmentation fault: 11
in expression starting at none:1
_ZL13getSameOpcodeN4llvm8ArrayRefIPNS_5ValueEEERKNS_17TargetLibraryInfoEj at ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/lib/julia/libLLVM.dylib (unknown line)
Allocations: 45669279 (Pool: 45667770; Big: 1509); GC: 42
zsh: segmentation fault  julia --proj
```

PR #589, opened 2025-06-12
Add Julia v1.11 to CI

trontrytel:
This PR updates our non-buildkite CI (gihub-CI, docs and code-coverage) to Julia v.1.11. 

Also fixes: https://github.com/CliMA/CloudMicrophysics.jl/issues/572

I have no idea why - but what fixed the segfault locally for me was to replace 

```
    return ifelse(
        cond_rate > FT(0),
        min(cond_rate, limit(qᵥ, dt, 1)),
        -min(abs(cond_rate), limit(qₗ, dt, 1)),
    )
```
by

```
    cond_limit = min(cond_rate, limit(qᵥ, dt, 1))
    evap_limit = min(abs(cond_rate), limit(qₗ, dt, 1))
    ret = ifelse(cond_rate > FT(0), cond_limit, -1 * evap_limit)
    return ret
```

Lets see if that also works in the cloud

### DIFF
--- a/.github/workflows/CodeCov.yml
+++ b/.github/workflows/CodeCov.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Julia
       uses: julia-actions/setup-julia@latest
       with:
-        version: '1.10'
+        version: '1.11'
 
     - name: Test with coverage
       env:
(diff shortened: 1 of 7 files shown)


oalcabes (approved):
Looks good to me!

PR #595, opened 2025-06-21
rft: wrap all tests in @testset

haakon-e:
## Purpose 

This pull request wraps the entire test suite in `@testset`. This ensures that even if one `@testset` fails, the rest of the testsuite attempts to run, so that you get an overview of _all_ issues with a proposed change from running the tests just once.
- There might be specific cases where this is undesirable. For those cases, we can add the option `failfast=true` to the `@testset` ([ref](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@testset)) which has the effect:
    - `failfast`: if `true`, any test failure or error will cause the testset and any child testsets to return immediately (the default is `false`).

Additionally:
- all tests now consistently `import Test as TT` (instead of e.g. `using Test`)
- any `@testset` that loops over options, e.g. `for FT in (Float64, Float32)`, does so with a consistent pattern that is supported by `Test`, e.g.:
```Julia
TT.@testset "Aerosol Activation Tests ($FT)" for FT in (Float64, Float32)
    test_aerosol_activation(FT)
end
```
- these kinds of `@testset` loops is an example of a nice place to add `failfast=true`... presumably if a test fails on `Float64`, it does so too on `Float32`.
- add `artifact_calling_tests.jl` (was missing)

To compare how a _successful_ test suite would look like, compare this PR's buildkite ([link](https://buildkite.com/clima/cloudmicrophysics-ci/builds/2140#0197904c-2a33-49ce-8fbd-2caaa8925713/184-1624)) to e.g. this PR's buildkite ([link](https://buildkite.com/clima/cloudmicrophysics-ci/builds/2143#01979dc7-adf5-4e68-96bb-498a01c7f397/184-187))
If any tests fail, the `Test Summary` will expand to show much more detail about the failing test(s). In addition, the stack trace of each failing test is printed, of course:
<details> <summary> click to see example of failing test </summary>

```Julia
# [...]

Melting Smoke Test: Test Failed at /Users/haakon/Documents/CliMA/clima-codes/CloudMicrophysics.jl/test/p3_tests.jl:542
  Expression: false

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/Test/src/Test.jl:679 [inlined]
 [2] macro expansion
   @ ~/Documents/CliMA/clima-codes/CloudMicrophysics.jl/test/p3_tests.jl:542 [inlined]
 [3] macro expansion
[...]

PR #577, opened 2025-05-29
feat: add P3 collision rates

haakon-e:
## Purpose 

P3 collision rates (moved from #543)

Implements source terms for the P3 scheme.
- calculates the rate of collision between ice and liquid particles, numerically integrates over the size distributions, returns bulk source terms
- computes dry and wet growth rates, partitioning sources into the ice and liquid phase

## To-do

- Add tests
- check docs

## Content

- added function signatures
- functions return bulk source terms

----
- [x] I have read and checked the items on the review checklist.

PR #596, opened 2025-06-21
add supporting P3 collisions sub-parameterizations

haakon-e:
This P3 implements:
- dry freeze rate from Musil (1970) [[link](https://doi.org/10.1175/1520-0469(1970)027%3C0474:CMOHGI%3E2.0.CO;2)]
- local rime density from Cober and List (1993) [[link](https://doi.org/10.1175/1520-0469(1993)050%3C1591:MOTHAM%3E2.0.CO;2)]

I would find it helpful with a second pair of eyes on the papers to verify my implementation. The docstring and inline comments are quite verbose, so finding your way around the paper is hopefully easy.

----

### DIFF
--- a/src/P3_processes.jl
+++ b/src/P3_processes.jl
@@ -47,17 +47,17 @@ function het_ice_nucleation(
 end
 
 """
-    ice_melt(state, logλ, Chen2022, aps, tps, Tₐ, ρₐ, dt; ∫kwargs...)
+    ice_melt(velocity_params::CMP.Chen2022VelType, aps, tps, Tₐ, ρₐ, dt, state, logλ; ∫kwargs...)
 
 # Arguments
- - `state`: a [`P3State`](@ref) object
- - `logλ`: the log of the slope parameter [log(1/m)]
- - `Chen2022`: struct containing Chen 2022 velocity parameters
- - `aps`: air properties
+ - `velocity_params`: [`CMP.Chen2022VelType`](@ref)
+ - `aps`: [`CMP.AirProperties`](@ref)
  - `tps`: thermodynamics parameters
  - `Tₐ`: temperature (K)
  - `ρₐ`: air density
  - `dt`: model time step (for limiting the tendnecy)
+ - `state`: a [`P3State`](@ref) object
+ - `logλ`: the log of the slope parameter [log(1/m)]
 
 # Keyword arguments
  - `∫kwargs`: Named tuple of keyword arguments passed to [`∫fdD`](@ref)
@@ -65,9 +65,7 @@ end
 Returns the melting rate of ice (QIMLT in Morrison and Mildbrandt (2015)).
 """
 function ice_melt(
-    state::P3State, logλ, Chen2022::CMP.Chen2022VelType,
-    aps::CMP.AirProperties, tps::TDI.PS,
-    Tₐ, ρₐ, dt;
+    velocity_params::CMP.Chen2022VelType, aps::CMP.AirProperties, tps::TDI.PS, Tₐ, ρₐ, dt, state::P3State, logλ;
     ∫kwargs = (;),
 )
     # Note: process not dependent on `F_liq`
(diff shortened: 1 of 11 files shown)


haakon-e:

- #595
    - #596 ◀
        - #577

src/P3_processes.jl
```diff
@@ -99,3 +99,130 @@ function ice_melt(
...
+
+```math
+ρ'_{rim} = 51 + 114 R_i - 5.5 R_i^2
+```
+where
+```math
+R_i = \\frac{ D_{liq} ⋅ |v_{liq} - v_{ice}| }{ 2 T_{sfc} }
+```
+and ``T_{sfc}`` is the surface temperature [°C], ``D_{liq}`` is the liquid particle
+diameter [m], ``v_{liq/ice}`` is the particle terminal velocity [m/s].
```
sajjadazimi, line 168: In the reference paper (Cober and List, 1993), diameter (radius there) is given in micrometer.

src/P3_processes.jl
```diff
@@ -99,3 +99,130 @@ function ice_melt(
...
+So the units of ``R_i`` are [m² s⁻¹ K⁻¹]. The units of ``ρ'_{rim}`` are [kg/m³].
+
+They assume for simplicity that ``T_{sfc}`` equals ``T``, the ambient air temperature.
+For real graupel, ``T_{sfc}`` is slightly higher than ``T`` due to latent heat release 
+of freezing liquid particles onto the ice particle. Morrison & Milbrandt (2013) 
+found little sensitivity to "realistic" increases in ``T_{sfc}``.
+
+Implementation follows Cober and List (1993), Eq. 16 and 17.
```
sajjadazimi, line 176: If I understand the reference paper correctly, Eq. 16 represents `R_i` as a function of ambient velocity. However, Eq. 14 represents `R_i` as a function of impact velocity (which could be given by `v_liq - v_ice`). But if you use Eq. 14 then for the rime density Eq. 15 must be used (which has different coefficients).

sajjadazimi (commented):
Looks good in general. I left a couple of comments on the rime density computations from Cober and List (1993).

src/P3_processes.jl
```diff
@@ -99,3 +99,130 @@ function ice_melt(
...
+then equals the solid bulk ice density, ``ρ^⭒ = 900 kg/m^3``.
+
+Note that Morrison & Milbrandt (2015) [MorrisonMilbrandt2015](@cite) only uses this 
+parameterization for collisions with cloud droplets.
+For rain drops, they use the solid bulk ice density, ``ρ^* = 900 kg/m^3``.
+We do not consider this distinction, and use this parameterization for all liquid particles.
+"""
+function compute_local_rime_density(state, velocity_params, ρₐ, T)
+    (; T_freeze) = state.params
+    T°C = T - T_freeze  # Convert to °C
```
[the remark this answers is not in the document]
haakon-e: My understanding is that we use variables for global constants, even as we expect them to never change.
For example, we have a variable for `grav` even though you could write 9.81.
If it's correct that `T_freeze`, in the context of microphysics schemes, can reasonably be set to some other value, then I'd be happy to reconsider how we reference the "freezing temperature".

src/P3_processes.jl
```diff
@@ -99,3 +99,131 @@ function ice_melt(
...
+Returns the local rime density [kg/m³] for a given Rᵢ [m² s⁻¹ °C⁻¹].
+
+Based on Cober and List (1993), Eq. 16 and 17 (valid `for 1 ≤ Rᵢ ≤ 8`).
+For `8 < Rᵢ ≤ 12`, linearly interpolate between `ρ′_rim(8) ≡ 611 kg/m³` and `ρ⭒ = 900 kg/m³`.
+See also the P3 fortran code, `microphy_p3.f90`, Line 3315-3323.
+"""
+function ρ′_rim_P3(Rᵢ)
+    # TODO: Externalize these parameters
+    a, b, c = 51, 114, -11 // 2 # coeffs for Eq. 17 in Cober and List (1993), converted to [kg / m³]
+    ρ⭒ = 900  # ρ^⭒: density of solid bulk ice
```
trontrytel, line 213: We should probably be using the ice density we already have in [parameters](https://github.com/CliMA/ClimaParams.jl/blob/c66047dcc360f288780d76ca3a40dfbb4aad3a12/src/parameters.toml#L2134) ?

PR #599, opened 2025-06-23
Remove NonEquilibrium clippings

oalcabes:
Here, I have removed the test for checking that some water vapor and condensate exists within the NonEq parameterization. I added the test to parcel so that behavior should not change there.


src/MicrophysicsNonEq.jl
```diff
@@ -79,23 +79,20 @@ function conv_q_vap_to_q_liq_ice_MM2015(
     T,
 )
     FT = eltype(tps)
-    if q_tot - q_ice - q_rai - q_sno > FT(0) #If there is water vapor or liquid
```
trontrytel, line 82: Could you add some words about the change of logic of how we handle small negative numbers in the docs and docstrings?

oalcabes: done!

PR #609, opened 2025-06-30
Update 2M scheme to working fluid changes

trontrytel:
This PR updates the 2M scheme to follow the working fluid changes. Meaning that when interacting with Thermodynamics when computing supersaturation `q_liquid = q_cloud_liquid + q_rain`, and `q_ice = q_cloud_ice + q_sno`. 

This results in changing the signature of the rain evaporation function in 2M scheme to include `q_sno`. It also changes the reference values in one of the tests, because `TDI.supersaturation_over_liquid(tps, qₜ, qₗ + qᵣ, qᵢ + qₛ, ρ, T)` is not the same as ``TDI.supersaturation_over_liquid(tps, qₜ, qₗ, qᵢ , ρ, T)`

This is a breaking change, but I noticed that even though we bump the version of CM.jl to `0.25` we did not make a release. So I will keep the `Project.toml` unchanged for this one and then register the new version when this is merged.


sajjadazimi (approved):
Thank you!

PR #615, opened 2025-07-11
Sa/make arg local

sajjadazimi:
## Purpose 
Extend ARG activation scheme for local applications

----

### DIFF
--- a/docs/src/AerosolActivation.md
+++ b/docs/src/AerosolActivation.md
@@ -140,11 +140,9 @@ where:
 
 ## Maximum Supersaturation
 
-Maximum supersaturation reached by the system ``S_{max}`` governs
-  what aerosol sizes are activated and what aren't.
-We estimate ``S_{max}`` by considering a parcel of air raising adiabatically
-  with a constant velocity, see for example [Rogers1975](@cite).
-The time rate of change of ``S`` is given by eq (10) in [Rogers1975](@cite)
+Maximum supersaturation reached by the system ``S_{max}`` governs what aerosol sizes are activated and what aren't.
+We estimate ``S_{max}`` by considering a parcel of air raising adiabatically with a constant velocity, see for example [KorolevMazin2003](@cite).
+The time rate of change of ``S`` is given by eq (A11) in [KorolevMazin2003](@cite)
 
 ```math
 \begin{equation}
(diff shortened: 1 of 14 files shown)


docs/src/AerosolActivation.md
```diff
@@ -285,3 +285,100 @@ The CloudMicrophysics package offers an advanced feature for predicting the aero
...
+```
+
+- Ice sink:
+```math
+\frac{d\chi}{dt}_{ice} = 4 \pi \rho_i N_i r_i G_i S_i,
+```
+
+with
+```math
+S_i = \xi (S + 1) - 1, \quad \xi = \frac{e_{s,i}}{e_{s,\ell}}.
```
trontrytel, line 319: Shouldn't it be `e_sl / e_si` ?

docs/src/AerosolActivation.md
```diff
@@ -285,3 +285,100 @@ The CloudMicrophysics package offers an advanced feature for predicting the aero
...
+
+- Varying the liquid droplet number concentration with no ice particles present,
+- Varying the ice crystal number concentration with no liquid droplets present,
+- Varying the initial droplet radius with no ice particles present,
+- Varying the initial ice particle radius with no liquid droplets present.
+
+In all cases, the parcel model assumes an initial zero supersaturation and includes activation, condensational growth, and depositional growth during adiabatic ascent. A background population of sulfate aerosol with an initial concentration of 500 cm⁻³ is present in all scenarios. For simplicity, ice particles are assumed to be monodisperse. Liquid droplets formed through activation and those present initially are represented as two separate monodisperse populations.
+
+The figure below shows the ratio of the maximum supersaturation reached in the parcel model to that predicted by the modified ARG scheme. This comparison highlights the conditions under which the modified ARG provides a good approximation to the more detailed parcel-model results.
+
```
trontrytel, line 379: Maybe it would be good o leave a sentence here about the need for additional clipping when coupled to a full model and treating it as a local parameterization?

parcel/ParcelCommon.jl
```diff
@@ -21,3 +21,13 @@ function AIDA_rate(model_t, data_t, data)
         return 0
     end
 end
+
+# Return activated particle radius
```
trontrytel, line 25:
```suggestion
# Return activated particle radius. See below eq. 19 in [Abdul-Razzaketal1998](@cite)

File: src/AerosolActivation.jl [hunk 4]
```diff
@@ -175,12 +179,38 @@ function max_supersaturation(
...
 
-    return FT(1) / sqrt(tmp)
+    ρ_air::FT = TDI.air_density(tps, T, p, q_tot, q_liq, q_ice)
+    r_liq::FT = N_liq < eps(FT) ? FT(0) : cbrt(ρ_air * q_liq / N_liq / ap.ρ_w / FT(4 / 3 * π))
+    K_liq::FT = FT(4 * π) * ap.ρ_w * N_liq * r_liq * G * γ
+
+    r_ice::FT = N_ice < eps(FT) ? FT(0) : cbrt(ρ_air * q_ice / N_ice / ap.ρ_i / FT(4 / 3 * π))
+    Gᵢ::FT = CO.G_func_ice(aip, tps, T)
+    ξ::FT = TDI.saturation_vapor_pressure_over_liquid(tps, T) / TDI.saturation_vapor_pressure_over_ice(tps, T)
+    K_ice::FT = FT(4 * π) * N_ice * r_ice * Gᵢ * γ
```
trontrytel, line 191: Is ice density missing here?

docs/src/AerosolActivation.md
```diff
@@ -285,3 +285,100 @@ The CloudMicrophysics package offers an advanced feature for predicting the aero
...
+- ``\rho_w`` and ``\rho_i`` are the densities of liquid water and ice.
+
+We define:
+```math
+K_l = 4 \pi \rho_w N_l r_l G_l, \quad K_i = 4 \pi \rho_i N_i r_i G_i,
+```
+
+yielding the modified supersaturation equation at peak supersaturation:
+```math
+0 = \alpha w - \gamma f(S_{max}) - K_l S_{max} - K_i \big[ \xi (S_{max} + 1) - 1 \big].
```
trontrytel, line 335: Should Kl and Ki be multiplied by gamma?

parcel/ParcelModel.jl
```diff
@@ -38,6 +38,9 @@ Base.@kwdef struct parcel_params{FT} <: CMP.ParametersType{FT}
     sampling_interval = FT(1)
     γ = FT(1)
     ip = CMP.Frostenberg2023(FT)
+    Nₐ = FT(1e8)
+    qₗ₀ = FT(1e-3)
```
trontrytel, line 42: Could we either hide them in MixedModes size distribution parameters, or change the names to indicate that those parameters are only used for that one distribution type?

trontrytel:
@sajjadazimi - your PR made me go back to (again) the parcel model equations. I went onto a little bit of a rabbit hole. But here are my notes:

PXL_20250714_234800194

Code issues:
- [Here](https://github.com/CliMA/CloudMicrophysics.jl/blob/15ddc710492a6f82121517f5d5c86acc218d96d9/src/AerosolActivation.jl#L157) I think first `R_m` should actually be `R_v`
- I think we should keep the `1/S` term, instead of dropping it and assuming it's close to 1. Especially since it seems it only disappeared from some terms. 
- I think `1/rho` is fine, it's just coming from reporting the rates per volume or mass of air. 

 It would also be great to go over the aerosol activation, parcel model and non-equilibrium microphysics and make the notation consistent among all three. Right now we have basically the same equation written out three times in three slightly different ways. 
 
Overall, the notation changes we can leave for later. But it would be good to fix the `R_v` and the missing `S` terms sooner. Either in this PR or next? Assuming you agree with my derivation notes.

trontrytel (approved):

docs/src/AerosolActivation.md
sajjadazimi: They include gamma

PR #617, opened 2025-07-23
fix: limiting behavior of 2M and P3 PSDs

haakon-e:
## Purpose 

Some reasonable limits for distributions. Needed in particular for the cases when there's no particles and no mass, because then the ratio `L/N` is undefined.

## Content

----


haakon-e:

- #617 ◀
    - #618
        - #619
            - #620

PR #618, opened 2025-07-23
feat: chebyshev-gauss quadrature instead of quadgk

haakon-e:
## Purpose 

Allocation-free & GPU-friendly integration.

Currently lives in the P3 module, but can be moved in the future if other modules need this.

## Content

----


haakon-e:

- #617
    - #618 ◀
        - #619
            - #620

test/p3_tests.jl
```diff
@@ -448,31 +458,44 @@ function test_numerical_integrals(FT)
...
             g(D) = v_term(D) * N′(D)
-            vel_N_estim = P3.∫fdD(g, state, logλ; p) / N_ice
-            vel_m_estim = P3.∫fdD(state, logλ; p) do D
-                g(D) * P3.ice_mass(state, D) / L_ice
-            end
+            gm(D) = g(D) * P3.ice_mass(state, D)
+            vel_N_estim_cheb = P3.integrate(g, bnds...; quad = P3.ChebyshevGauss(10)) / N_ice
+            vel_m_estim_cheb = P3.integrate(gm, bnds...; quad = P3.ChebyshevGauss(10)) / L_ice
+            @test vel_N ≈ vel_N_estim_cheb rtol = 0.005
+            @test vel_m ≈ vel_m_estim_cheb rtol = 0.05
```
trontrytel, line 480: Do you know why the tolerance increased so badly in those tests? And should we care?

test/microphysics2M_tests.jl
```diff
@@ -577,20 +578,24 @@ function test_microphysics2M(FT)
...
 
             # Sanity checks for number concentrations for rain
-            ND = QGK.quadgk(f_D, D_min, D_max)[1]
-            Nx = QGK.quadgk(f_x, x_min, x_max)[1]
-            ND_psd = QGK.quadgk(psd, D_min, D_max)[1]
-            TT.@test ND ≈ Nᵣ rtol = 1e-5
-            TT.@test Nx ≈ Nᵣ rtol = 5e-3
-            TT.@test ND_psd ≈ Nᵣ rtol = 1e-5
+            ND = P3.integrate(f_D, D_min, D_max; quad = P3.ChebyshevGauss(1000))
+            Nx = P3.integrate(f_x, x_min, x_max; quad = P3.ChebyshevGauss(100_000))
```
trontrytel, line 582: Is that a realistic number of points for actual simulations? Shoul we be testing with such high numbers?

src/P3_processes.jl
```diff
@@ -83,7 +83,8 @@ function ice_melt(
 
     # Integrate
     fac = 4 * K_therm / L_f * (Tₐ - T_freeze)
-    dLdt = fac * ∫fdD(state, logλ; ∫kwargs...) do D
+    bnds = integral_bounds(state, logλ; p = 1e-6)
```
trontrytel, line 86: Is `p` here the tolerance? And if yes, could we name it so?

test/microphysics2M_tests.jl
haakon-e: No, it's way too many for practical simulations. I don't yet know why I needed so many points to get reasonable accuracy. but could try to investigate. I suspect it has to do with the physical quantities/units involved in the integral. In other words, rescaling/non-dimensionalizing the integrands could lead to better performance.

In any case, we don't actually ever integrate the distribution over mass, so it doesn't matter too much. The key point is that we should get the same number integrating over mass or diameter.

PR #619, opened 2025-07-23
fix: bulk_liquid_ice_collision_sources issues

haakon-e:
## Purpose 

a few issues discovered during integration to kinematicdriver.

Note: With regards to changes in arguments passed to `bulk_liquid_ice_collision_sources`, I have added a test for this method in #621.

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #620, opened 2025-07-23
feat: ice terminal velocity function aliases

haakon-e:
## Purpose 

I haven't fully settled on the API here (hopefully this will not be needed with some upcoming improvements to ClimaCore), but adding some very verbose signatures for now to allow for computing terminal velocity without the `P3State`.

## Content

----


haakon-e:

- #617
    - #618
        - #619
            - #620 ◀

PR #626, opened 2025-08-09
Cloud terminal velocity for 2M 

sajjadazimi:
## Purpose 
Add cloud terminal velocity for 2M based on Stokes regime fall speed and gamma size dist assumption

----


docs/src/plots/TerminalVelocity.jl
```diff
@@ -98,6 +98,25 @@ function rain_terminal_velocity_individual_SB(
...
+"""
+    rstokes_regime_terminal_velocity(param_set, ρ, r)
+
+ - `param_set` - set with free parameters
+ - `ρ` - air density
+ - `r` - radius of the cloud droplets
+
+Returns the fall velocity of a cloud droplet in the Stokes regime (Re < 1)
+"""
+function stokes_regime_terminal_velocity(
```
trontrytel, line 110: I think there is a use case to add it to `src`. @haakon-e needs particle-level formulation of sedimentation velocity, because he is doing numerical integrals in P3 scheme to compute the collision rate. Maybe we could add it to `Common` next to `Chen` particle velocities? 

Along the same lines, do you think SB scheme would be ok to use with Chen 2022 for sedimentation velocity for rain? That one doesn't have the issue of negative numbers close to zero, so it's easier to plug in into the P3 scheme integrals. 

The relevant code is here: https://github.com/CliMA/CloudMicrophysics.jl/blob/7d30216674cda6668c959f8eba40bad1220370d7/src/P3_processes.jl#L133

src/parameters/TerminalVelocity.jl
```diff
@@ -116,6 +115,9 @@ Base.@kwdef struct SB2006VelType{FT} <: TerminalVelocityType{FT}
     aR::FT
     bR::FT
     cR::FT
+    ρw::FT
+    ν_air::FT
```
trontrytel, line 119: Would it make sense to create two types: one for rain sedimentation parameters and one for cloud? Just in case if we want to use SB2006 scheme with Stokes for cloud sedimentation but Chen 2022 for rain sedimentation?

trontrytel (approved):
Thank you! I just left some small comments about parameters and docs.

PR #636, opened 2025-08-25
fix(2M): cloud_terminal_velocity type instability

haakon-e:
This pull request updates the calculation of the terminal velocity prefactor in the `cloud_terminal_velocity` function. The change adjusts the order of division and type of exponent to ensure type stability for `Float32` inputs

Physics calculation correction:

* In `src/Microphysics2M.jl`, the formula for `terminal_velocity_prefactor` in `cloud_terminal_velocity` was updated to divide by `π` after `ρw`, and the exponent was changed to use rational division for improved precision.


sajjadazimi (approved):
Great catch, thank you!

Issue #608, opened 2025-06-27
Finish Thermodynamics revamp


trontrytel:
Left to do in follow up PRs:
- [ ] Update 2M microphysics to working fluid.
- [ ] Delete `TD.PhasePartition` from function signatures in `Atmos` and then delete the wrappers.
- [ ] Change the notation in `CloudMicrophysics` to `q_tot` = `q_vap` + `q_liq` + `q_ice`. And `q_liq` = `q_lcl` + `q_rai`, `q_ice` = `q_icl` + `q_sno`
- [ ] Update the documentation in `CloudMicrophysics` to explain the thermodynamics interface
- [ ] Following @haakon-e [suggestion](https://github.com/CliMA/CloudMicrophysics.jl/pull/590#issuecomment-2993028042) create internal thermodynamics parameters set

PR #646, opened 2025-09-16
The Great Rename

trontrytel:
Next step on the issue https://github.com/CliMA/CloudMicrophysics.jl/issues/608:
  - Introducing `q_lcl` and `q_icl` for cloud liquid water and cloud ice. This is done to differentiate from `q_liq` and `q_ice` that represent the total condensed liquid water and total ice.
  - Adding a small docs page that explains the basics of coupling with thermodynamics of a host model.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.27.4"
+version = "0.28.0"
 
 [deps]
 ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"


docs/src/API.md
```diff
@@ -8,8 +8,8 @@ CurrentModule = CloudMicrophysics
 ```@docs
 MicrophysicsNonEq
 MicrophysicsNonEq.τ_relax
-MicrophysicsNonEq.conv_q_vap_to_q_liq_ice
-MicrophysicsNonEq.conv_q_vap_to_q_liq_ice_MM2015
+MicrophysicsNonEq.conv_q_vap_to_q_lcl_icl
```
oalcabes, line 11: Would these cloud sources really only be applying to the cloud liquid and ice values, and not also the rain and snow? I guess I don't have a lot of insight into how it's being used downstream but my impression was that it's happening to the entire liquid and ice fields.

docs/src/IceNucleation.md
```diff
@@ -45,7 +45,7 @@ The other parameterization models the nucleation rate of ice
   see eq. (5) in [Mohler2006](@cite).
 ```math
 \begin{equation}
-\frac{dn_{ice}}{dt} = N_{aer} a \frac{dS_i}{dt}
+\frac{dn_{icl}}{dt} = N_{aer} a \frac{dS_i}{dt}
```
oalcabes, line 48: I guess another question here -- why would this only apply to cloud ice and not also snow?

docs/src/plots/Microphysics1M_plots.jl
```diff
@@ -140,10 +140,10 @@ q_sat = ϵ * p_sat / (p + p_sat * (ϵ - 1.0))
 q_rain_range = range(1e-8, stop = 5e-3, length = 100)
 q_tot = 15e-3
 q_vap = 0.15 * q_sat
-q_ice = 0.0
-q_liq = q_tot - q_vap - q_ice
+q_icl = 0.0
+q_lcl = q_tot - q_vap - q_icl
 q_sno = 0.0
-R = TDI.Rₘ(tps, q_tot, q_liq, q_ice)
+R = TDI.Rₘ(tps, q_tot, q_lcl, q_icl)
```
oalcabes, line 146: similarly would need to add rai and sno for R_m computation

parcel/ParcelTendencies.jl
```diff
@@ -253,7 +253,7 @@ function condensation(params::NonEqCondParams, PSD, state, ρ_air)
 
         qₜ = qᵥ + qₗ + qᵢ
 
-        cond_rate = MNE.conv_q_vap_to_q_liq_ice_MM2015(liquid, tps, qₜ, qₗ, qᵢ, FT(0), FT(0), ρ_air, T)
+        cond_rate = MNE.conv_q_vap_to_q_lcl_icl_MM2015(liquid, tps, qₜ, qₗ, qᵢ, FT(0), FT(0), ρ_air, T)
```
oalcabes, line 256: these might only be cloud here bc we consider no precip

docs/src/plots/Microphysics1M_plots.jl
```diff
@@ -140,10 +140,10 @@ q_sat = ϵ * p_sat / (p + p_sat * (ϵ - 1.0))
 q_rain_range = range(1e-8, stop = 5e-3, length = 100)
 q_tot = 15e-3
 q_vap = 0.15 * q_sat
-q_ice = 0.0
-q_liq = q_tot - q_vap - q_ice
+q_icl = 0.0
+q_lcl = q_tot - q_vap - q_icl
```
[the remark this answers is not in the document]
trontrytel: Yes. In the plotting loop down below I subtract the rain to get the actual liquid water in each plot.

docs/src/plots/Microphysics1M_plots.jl
trontrytel: Yes. I'm just being sloppy here. I didn't want to make the plotting script even longer by recomputing the `Rm` per plot.

docs/src/API.md
trontrytel: Yes but also no. I agree with you that in principle it is the same process of condensation/evaporation or deposition/sublimation. However for bigger particles falling at some significant terminal velocities people tend to split it. This allows them to also add the ventilation effects for rain evaporation or snow deposition/sublimation. 

I think in the 1M scheme I would keep it separate. I.e. use the above tendency for cloud formation. And use the other parameterizations from the 1M scheme to do rain evaporation and snow deposition/sublimation.

The P3 scheme will probably apply this term to the whole ice category. And maybe have on top of that some additional correction for ventilation.

docs/src/IceNucleation.md
trontrytel: Yes. Again, I think the P3 scheme will want to use those nucleation terms as a source for the whole ice category. But for the 1M scheme we use ice nucleation to form small ice crystals (lumped into cloud category). And then further growth via deposition on those crystals is what finally leads to snow formation

docs/src/MicrophysicsNonEq.md
```diff
@@ -65,8 +65,8 @@ This formulation is derived from [MorrisonGrabowski2008_supersat](@cite)
 
 ```math
 \begin{equation}
-   \left. \frac{d \, q_{liq}}{dt} \right|_{cond, evap} = \frac{q_{vap} - q_{sl}}{\tau_l \Gamma_l}; \;\;\;\;\;\;\;
-   \left. \frac{d \, q_{ice}}{dt} \right|_{dep, sub}   = \frac{q_{vap} - q_{si}}{\tau_i \Gamma_i}
+   \left. \frac{d \, q_{lcl}}{dt} \right|_{cond, evap} = \frac{q_{vap} - q_{sl}}{\tau_l \Gamma_l}; \;\;\;\;\;\;\;
```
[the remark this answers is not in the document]
trontrytel: The 1M scheme does not consider condensation on rain drops. I think it's mostly from the fact that they are falling fast. For P3 scheme, or other single category schemes, we would use it for the whole category

docs/src/Thermodynamics.md
```diff
@@ -0,0 +1,30 @@
...
+
+  - Gas constants for dry air and water vapor
+  - Gas constant for moist air
+  - Latent heats of vaporization, fusion and sublimation
+  - Specific heat capacity under constant pressure for moist air
+  - Saturation vapor pressure and supersaturation over liquid and ice
+  - Water vapor specific content
+  - Specific content from partial pressure and partial pressure from specific content
+  - Gravitational constant (only used in the adiabatic parcel example)
+  - Air density (only used in tests)
```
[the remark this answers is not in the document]
trontrytel: Maybe I'll add them as a second list. I figured maybe people would still care to be able to generate the docs and run the tests with a hypothetical new thermodynamics.

docs/src/Thermodynamics.md
```diff
@@ -0,0 +1,30 @@
+# Thermodynamics Interface
+
+The `CloudMicrophysics.jl` library is designed to be flexible and agnostic
+  to the specific thermodynamics of the model it is coupled with.
+It requires certain functions and free parameters to be provided by the host model,
+  such as, for example, the formula to compute the saturation vapor pressure.
```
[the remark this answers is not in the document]
trontrytel: Yeah. Thats how it is right now. I think at some point for fun me and @haakon-e will try to turn it into an extension. So right now this is written a little aspirationally

PR #658, opened 2025-12-19
revert 2M epsilon change

sajjadazimi:
## Purpose 
Revert 2M epsilon change for numerical stability. Using `eps(FT)` for 2M microphysics is safer as we have division of variables. These thresholds will be replaced by appropriate variable limits in the future.
This PR will fix the failure of CI in `KinematicDriver.jl` and `ClimaAtmos.jl`.

----


trontrytel (approved):
Would it be better to define another epsilon in common for 2-moment microphysics `CO.ϵ_numerics_2M`? This would allow us to later change everything in one go. Approving as is, in case you disagree and just wan to merge and move on.

PR #662, opened 2026-01-15
Update to the new Thermodynamics.jl interface. Not supporting the old interface anymore.

trontrytel:
Updates CM.jl to the new Thermodynamics.jl interface. I'm deleting the old interface and making it a breaking release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CloudMicrophysics"
 uuid = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 authors = ["Climate Modeling Alliance"]
-version = "0.30.0"
+version = "0.31.0"
 
 [deps]
 ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"


Project.toml
```diff
@@ -33,6 +33,6 @@ MLJ = "0.20"
 RootSolvers = "0.3, 0.4, 1"
 SpecialFunctions = "1, 2"
 StaticArrays = "1.9"
-Thermodynamics = "0.14, 0.15"
+Thermodynamics = "0.15.4"
```
haakon-e: do you want to be this strict? Or would just `"0.15"` be sufficient?

trontrytel: I think the new Thermodynamics interface was introduced in 15.3

tapios: FWIK, `Thermodynamics = "^0.15.4"` is equivalent to ``Thermodynamics = "0.15.4"`. 

We can use `Thermodynamics = ">= 0.15.4"`, which allows all versions after this (which is what I think we want).

docs/src/plots/P3ImmersionFreezing.jl
```diff
@@ -12,9 +12,7 @@ tps = TDI.PS(FT)
 
 # helper functions
 function RH2qₜ(T, RH)
-    eᵥ_sat = TDI.saturation_vapor_pressure_over_liquid(tps, T)
-    eᵥ = RH * eᵥ_sat
-    qᵥ = 1 / (1 - tps.molmass_dryair / tps.molmass_water * (eᵥ - p) / eᵥ)
+    qᵥ = TDI.RHₗ2qᵥ(tps, p, T, RH)
```
tapios: Can we just get rid of `RHₗ2qᵥ` (if for not other reason that I cannot read the subscripts) and use `q_vap_from_RH(, ... TD.Liquid())` instead? (Please no longer use `q_vap_from_RH_liquid`; this will also disappear from TD. The form here with `TD.Liquid()` is equivalent.)

docs/src/plots/P3ImmersionFreezing.jl
```diff
@@ -12,9 +12,7 @@ tps = TDI.PS(FT)
 
 # helper functions
 function RH2qₜ(T, RH)
-    eᵥ_sat = TDI.saturation_vapor_pressure_over_liquid(tps, T)
-    eᵥ = RH * eᵥ_sat
-    qᵥ = 1 / (1 - tps.molmass_dryair / tps.molmass_water * (eᵥ - p) / eᵥ)
+    qᵥ = TDI.q_vap_from_RH_over_liquid(tps, p, T, RH)
```
tapios:
```suggestion
    qᵥ = TD.q_vap_from_RH(tps, p, T, RH, TD.Liquid())
```
(And add TD import.) In general, I think it's safer and easier to change to use TD functions directly, rather than overloaded methods.

src/ThermodynamicsInterface.jl
```diff
@@ -54,9 +58,9 @@ saturation_vapor_pressure_over_ice(tps::PS, T) =
 
 # (only used in tests)
 saturation_vapor_specific_content_over_liquid(tps::PS, T, ρ) =
-    TD.q_vap_saturation_generic(tps, T, ρ, TD.Liquid())
+    TD.q_vap_saturation(tps, T, ρ, TD.Liquid())
 saturation_vapor_specific_content_over_ice(tps::PS, T, ρ) =
-    TD.q_vap_saturation_generic(tps, T, ρ, TD.Ice())
+    TD.q_vap_saturation(tps, T, ρ, TD.Ice())
```
tapios: As above, it's cleaner to use the TD functions directly.

PR #663, opened 2026-01-17
docs: add cond/evap rate plot

haakon-e:
Add condensation/evaporation rate plot to docs

Should look something like:
cloud_sources_ql_z


trontrytel (approved):
Thank you!

PR #665, opened 2026-01-28
Fused tendencies for bulk microphysics

tapios:
## Purpose 
This PR moves the evaluation of microphysics tendencies from ClimaAtmos to CloudMicrophysics (which seems to be the more natural home). It fuses the tendency evaluations into a single function. Reasons:
* Allows quadrature over tendencies, taking SGS fluctuations into account (the central reason)
* May improve performance by fusing many microphysics kernels into 1
* More clearly separates concerns: atmospheric dynamics vs evaluation of microphysics tendencies

Not done in this PR but should be done soon:
* Enable supersaturation-based autoconversion of cloud ice to snow
* Simplify and unify parameter handling and dispatch in microphysics
* Complete 2M implementation of tendencies 

## Content

----


test/bulk_tendencies_tests.jl
```diff
@@ -0,0 +1,702 @@
...
+        @test tendencies isa NamedTuple
+        @test haskey(tendencies, :dq_lcl_dt)
+        @test haskey(tendencies, :dq_icl_dt)
+        @test haskey(tendencies, :dq_rai_dt)
+        @test haskey(tendencies, :dq_sno_dt)
+        @test tendencies.dq_lcl_dt isa FT
+        @test tendencies.dq_icl_dt isa FT
+        @test tendencies.dq_rai_dt isa FT
+        @test tendencies.dq_sno_dt isa FT
+    end
```
haakon-e:
```suggestion
```
This test is functionally equivalent to `@testset "Type stability (@inferred) for GPU safety"`. Both should not be needed.

test/bulk_tendencies_tests.jl
```diff
@@ -0,0 +1,702 @@
...
+end
+
+function test_bulk_microphysics_0m_tendencies(FT)
+    tps = TD.Parameters.ThermodynamicsParameters(FT)
+    params_0M = CMP.Parameters0M(FT)
+    mp = (; params_0M)
+
+    T_freeze = TD.Parameters.T_freeze(tps)
+
+    @testset "BulkMicrophysicsTendencies 0M - Return type" begin
```
haakon-e: Same as my comment for 1M: this `@testset` is redundant with `@testset "BulkMicrophysicsTendencies 0M - Type stability"`

test/bulk_tendencies_tests.jl
```diff
@@ -0,0 +1,702 @@
...
+            FT(1e-3),
+            FT(0),
+            FT(1e-4),
+            FT(0),
+            FT(1e8),
+            FT(1e4),
+        )
+        @test tendencies isa NamedTuple
+        @test haskey(tendencies, :dq_lcl_dt)
+        @test haskey(tendencies, :dn_rai_dt)
```
haakon-e:
```suggestion
        @test tendencies isa @NamedTuple{dq_lcl_dt::FT, dq_rai_dt::FT, dn_lcl_dt::FT, dn_rai_dt::FT}
```
This test passes locally, when written as I suggest. Could you check again on your end?

Project.toml
```diff
@@ -33,6 +33,6 @@ MLJ = "0.20"
 RootSolvers = "0.3, 0.4, 1"
 SpecialFunctions = "1, 2"
 StaticArrays = "1.9"
-Thermodynamics = "0.14, 0.15"
+Thermodynamics = "0.14, 0.15, 0.16"
```
haakon-e:
```suggestion
Thermodynamics = "0.15.3"
```
I think this is the correct compat bound we need here.

haakon-e (approved):
Thank you for the PR! The implementation as-is looks good and I only have a few minor suggestions for improvements. Feel free to apply any subset of them as you see fit.

trontrytel:
Sounds good. Yes, I can do the simple clipping here and we take care of the smooth part in Atmos

PR #667, opened 2026-01-30
Aerosol activation details: use sqrt + literal pow (+ non-short-circuiting formula for smax)

glwagner:
This change avoids a pow in favor of the more efficient `sqrt` + literal pow `^3` (plus a minor change to aerosol activation implementation in parcel tendencies).


glwagner:
~~Although now I am realizing this is not in `src` so may not be that important...~~

Nevermind, I found more things that might be actually useful.

haakon-e (approved):
Thank you for tracking down these performance improvements!

PR #674, opened 2026-02-05
Adds AD compatible input clamping and defensive checks to improve robustness of 1M microphysics

tapios:
## Purpose 
* Adds defensive and AD compatible clamping
* Adds some defensive checks to avoid numerical instabilities
* Introduces a Utilities.jl module for clamping function and other utilities (keeping them in Common.jl would create circular dependency with ThermodynamicsInterface)
* A few code simplifications
* Bump patch release

## Content

----

### DIFF
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -9,6 +9,7 @@ using UnrolledUtilities
 
 import ..Parameters as CMP
 import ..ThermodynamicsInterface as TDI
+import ..Utilities as UT
 const HPS = CMP.H2SO4SolutionParameters
 
 export G_func_liquid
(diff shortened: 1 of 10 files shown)


src/Common.jl
```diff
@@ -9,6 +9,7 @@ using UnrolledUtilities
 
 import ..Parameters as CMP
 import ..ThermodynamicsInterface as TDI
+import ..Utilities: clamp_to_nonneg, ϵ_numerics, ϵ_numerics_2M_M, ϵ_numerics_2M_N
```
trontrytel: I prefer the style of 

`import ..Utilities as UT`

and then in the code `UT.clamp_to_nonneg` I know it's a personal preference, but I think it improved the readability by showing where different functions are coming from

tapios:
> Thank you! Only have a small comment

Thank you for reviewing this!

PR #676, opened 2026-02-11
Add options for NonEq relaxation timescales as functions of N and T

oalcabes:
Draft PR to make new versions of the NonEq scheme, namely tau(T) and tau(N) functionality. Still need to review docs and think carefully about testing, but I'd love comments and thoughts on the PR! This may break atmos as I've made changes to the bulk_microphysics_tendency function.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -10,6 +10,8 @@ CurrentModule = CloudMicrophysics
 MicrophysicsNonEq
 MicrophysicsNonEq.τ_relax
 MicrophysicsNonEq.conv_q_vap_to_q_lcl_icl
+MicrophysicsNonEq.conv_q_vap_to_q_lcl_MM2015
+MicrophysicsNonEq.conv_q_vap_to_q_icl_MM2015
 MicrophysicsNonEq.conv_q_vap_to_q_lcl_icl_MM2015
 MicrophysicsNonEq.INP_limiter
 MicrophysicsNonEq.terminal_velocity
--- a/docs/src/MicrophysicsNonEq.md
+++ b/docs/src/MicrophysicsNonEq.md
@@ -99,7 +99,7 @@ To see this, it is necessary to use the definitions of ``\tau``, ``q_{sl}``, and
 
 ```math
 \begin{equation}
-  \tau = 4 \pi N_{tot} \bar{r}, \;\;\;\;\;\;\;\;
+  \tau = \left( 4 \pi N_{tot} D_v \bar{r} \right)^{-1}, \;\;\;\;\;\;\;\;
   q_{sl} = \frac{e_{sl}}{\rho R_v T}, \;\;\;\;\;\;\;\;
   D_v = \frac{K}{\rho c_p}.
 \end{equation}
@@ -126,6 +126,41 @@ include("plots/NonEqCondEvapRate.jl")
 
 
 
+
+### Liquid and ice relaxation timescales as functions of number concentration
+
+We also provide functionality to set the relaxation timescales of liquid and ice, ``\tau_l`` and ``\tau_i``, as functions of cloud droplet number concentrations ``N_l`` and ``N_i`` following [MorrisonGrabowski2008_supersat](@cite) and [MorrisonMilbrandt2015](@cite). First, we approximate average radius of droplets from mass and number:
+
+```math
+\begin{equation}
+\bar{r} = \left( \frac{3q_c}{4 π N(T) \rho_c} \right)^{1/3} 
+\end{equation}
+```
+
+where ``q_c`` is the mass of condensate and ``ρ_c`` is the density of the condensate (either liquid or water). Then, we calculate the relaxation timescale:
+
+```math
+\begin{equation}
+\tau = \left( 4 \pi N_c D_v \bar{r} \right)^{-1}
+\end{equation}
+```
+
+where ``D_v`` is the thermal diffusivity and ``N_c`` is the droplet number concentration of the condensate, either ``N_l`` or ``N_i``.
+
+### Ice relaxation timescale as function of temperature through Frostenberg 2023
+
+We can also calculate ``\tau_i`` as a function of temperature using the [Frostenberg2023](@cite) parameterization. Namely, we approximate the concentration of ice crystals as equal to the mean amount of ice nucleating particles for a given temperature using [Frostenberg2023](@cite): 
+```math
+N_i(T) = ln(-(b \cdot T)^9 \times 10^{-9})
+```
+
+ ``\sigma^2`` is the variance, ``a`` and ``b`` are coefficients. The parameters defined in [Frostenberg2023](@cite) for marine data sets are ``\sigma=1.37``, ``a=1`` m``^3``, ``b=1``/C.
+
+Then, we use ``N_i`` to solve for ``\tau_i`` as demonstrated above. Here we show an example of how ``\tau_i`` changes with temperature for a set ``q_{icl}=1e-6``:
+```@example
+include("plots/NonEqFrostenberg.jl")
+```
+![](taui_Frostenberg.svg)
 
 
 ## Cloud condensate sedimentation
--- a/src/BulkMicrophysicsTendencies.jl
+++ b/src/BulkMicrophysicsTendencies.jl
@@ -77,6 +77,20 @@ This unified scheme handles:
 """
 struct Microphysics2Moment <: MicrophysicsScheme end
 
+"""
+    NonEq Scheme
+
+Abstract type for NonEquilibrium microphysics parameterizations. Namely,
+this scheme specifies how to set the relaxation timescales (τₗ, τᵢ) for
+liquid condensation/evaporation and ice deposition/sublimation.
+NonEq_Constant sets τₗ and τᵢ are determined by the τ_relax arguments in
+lcl and icl, whereas NonEq_N calculates τₗ and τᵢ from liquid and ice
+number concentration.
+"""
+abstract type NonEqScheme end
+struct NonEq_Constant <: NonEqScheme end
+struct NonEq_N <: NonEqScheme end
+
 # --- Helper Functions ---
 
 """
@@ -100,8 +114,18 @@ end
 
 """
     bulk_microphysics_tendencies(
-        ::Microphysics1Moment, mp, tps,
-        ρ, T, q_tot, q_lcl, q_icl, q_rai, q_sno,
+        ::Microphysics1Moment,
+        noneq_scheme::NonEqScheme,
+        mp,
+        tps,
+        aps,
+        ρ,
+        T,
+        q_tot,
+        q_lcl,
+        q_icl,
+        q_rai,
+        q_sno,
     )
 
 Compute all 1-moment microphysics tendencies in one fused call.
@@ -113,6 +137,7 @@ This is a pure function of local thermodynamic state, suitable for:
 - Unit testing in isolation
 
 # Arguments
+- `noneq_scheme` Type of NonEquilibrium scheme to use
 - `mp`: NamedTuple of microphysics parameters with fields:
   - `lcl`: CloudLiquid parameters
   - `icl`: CloudIce parameters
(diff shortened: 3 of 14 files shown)


docs/src/MicrophysicsNonEq.md
```diff
@@ -123,6 +123,41 @@ include("plots/NonEqCondEvapRate.jl")
...
+\begin{equation}
+\tau = \left( 4 \pi N_c D_v \bar{r} \right)^{-1}
+\end{equation}
+```
+
+where ``D_v`` is the thermal diffusivity and ``N_c`` is the droplet number concentration of the condensate, either ``N_l`` or ``N_i``.
+
+### Ice relaxation timescale as function of temperature through Frostenberg 2023
+
+We can also calculate ``\tau_i`` as a function of temperature using the [Frostenberg2023](@cite) parameterization. Namely, we approximate the concentration of ice cloud droplets as equal to the mean amount of ice nucleating particles for a given temperature using [Frostenberg2023](@cite):
```
trontrytel:
```suggestion
We can also calculate ``\tau_i`` as a function of temperature using the [Frostenberg2023](@cite) parameterization. Namely, we approximate the concentration of ice crystals as equal to the mean amount of ice nucleating particles for a given temperature using [Frostenberg2023](@cite): 
```

docs/src/API.md
```diff
@@ -10,6 +10,8 @@ CurrentModule = CloudMicrophysics
 MicrophysicsNonEq
 MicrophysicsNonEq.τ_relax
 MicrophysicsNonEq.conv_q_vap_to_q_lcl_icl
+MicrophysicsNonEq.conv_q_vap_to_q_lcl_MM2015
+MicrophysicsNonEq.conv_q_vap_to_q_icl_MM2015
```
trontrytel: Do we still want to keep `MicrophysicsNonEq.conv_q_vap_to_q_icl_MM2015`?

oalcabes: Ideally we would keep "MicrophysicsNonEq.conv_q_vap_to_q_lcl_MM2015" and "MicrophysicsNonEq.conv_q_vap_to_q_icl_MM2015" and delete "MicrophysicsNonEq.conv_q_vap_to_q_lcl_icl_MM2015" - I was leaving it now for backwards compatibility but if we're okay with a breaking release I can just delete it

src/BulkMicrophysicsTendencies.jl
```diff
@@ -663,15 +709,18 @@ end
 # --- 2-Moment Microphysics Helper Functions ---
 
 """
-    warm_rain_tendencies_2m(sb, q_lcl, q_rai, ρ, n_lcl, n_rai)
+    warm_rain_tendencies_2m(warm_rain, noneq_scheme, tps, T, q_tot, q_lcl, q_rai, q_ice, ρ, n_lcl, n_rai)
```
oalcabes, line 711: TO MYSELF -- FIX

oalcabes:
Hi @trontrytel and @sajjadazimi -- I'm coming back to this PR. The PR is doing three main things, namely: (1) refactors the NonEquilibrium parameterization to accept tau as an argument rather than calculating it inside of the function (2) introduce a version of tau_i calculated from Frostenberg et. al. (2023) and (3) add parcel and documentation examples for the Frostenberg option. For now I would like to focus most on the refactoring change -- if you all like it we can keep it in this PR, but if you have major comments perhaps I'll do it in a separate PR.

oalcabes:
I specifically desire comments on src/BulkMicrophysicsTendencies.jl

src/BulkMicrophysicsTendencies.jl
```diff
@@ -177,12 +213,22 @@ This is a pure function of local thermodynamic state, suitable for:
 
     # --- Cloud condensate formation (non-equilibrium) ---
 
+    if noneq_scheme isa NonEq_Constant
+        τₗ = lcl.τ_relax
+        τᵢ = icl.τ_relax
+    elseif noneq_scheme isa NonEq_N # this assumes we're using the Frostenberg parameterization for τᵢ
```
oalcabes: To decide: do we want to have liquid calculated like this as well, or just keep the ice relaxation timescale changing instead?

src/BulkMicrophysicsTendencies.jl
```diff
@@ -663,15 +709,18 @@ end
...
 autoconversion, self-collection, accretion, and rain breakup.
 
 Used by both warm-only and warm+ice dispatch methods to reduce code duplication.
 
 # Arguments
-- `sb`: SB2006 parameters
+- `warm_rain`
+    - `sb`: SB2006 parameters
+    - `aps`: air property parameters % OLIVIA CLEAN THIS UP -- also missing ice???
+- `noneq_scheme` Type of NonEquilibrium scheme to use (NOT NECESSARILY NEEDED BUT KEEP IT FOR THE FUTURE)
```
oalcabes: currently passing this flag but not using it yet -- may not be necessary if we decide to keep tau_l always constant for now and only change tau_i

docs/src/MicrophysicsNonEq.md
```diff
@@ -127,6 +127,41 @@ include("plots/NonEqCondEvapRate.jl")
...
+
+We also provide functionality to set the relaxation timescales of liquid and ice, ``\tau_l`` and ``\tau_i``, as functions of cloud droplet number concentrations ``N_l`` and ``N_i`` following [MorrisonGrabowski2008_supersat](@cite) and [MorrisonMilbrandt2015](@cite). First, we approximate average radius of droplets from mass and number:
+
+```math
+\begin{equation}
+\bar{r} = \left( \frac{3q_c}{4 π N(T) \rho_c} \right)^{1/3} 
+\end{equation}
+```
+
+where ``q_c`` is the mass of condensate and ``ρ_c`` is the density of the condensate (either liquid or water). Then, we calculate the relaxation timescale:
```
sajjadazimi, line 140: `liquid or water`?

docs/src/MicrophysicsNonEq.md
```diff
@@ -127,6 +127,41 @@ include("plots/NonEqCondEvapRate.jl")
...
+where ``D_v`` is the thermal diffusivity and ``N_c`` is the droplet number concentration of the condensate, either ``N_l`` or ``N_i``.
+
+### Ice relaxation timescale as function of temperature through Frostenberg 2023
+
+We can also calculate ``\tau_i`` as a function of temperature using the [Frostenberg2023](@cite) parameterization. Namely, we approximate the concentration of ice crystals as equal to the mean amount of ice nucleating particles for a given temperature using [Frostenberg2023](@cite): 
+```math
+N_i(T) = ln(-(b \cdot T)^9 \times 10^{-9})
+```
+
+ ``\sigma^2`` is the variance, ``a`` and ``b`` are coefficients. The parameters defined in [Frostenberg2023](@cite) for marine data sets are ``\sigma=1.37``, ``a=1`` m``^3``, ``b=1``/C.
```
sajjadazimi, line 157: I only see `b` in the eq

src/BulkMicrophysicsTendencies.jl
```diff
@@ -77,6 +77,20 @@ This unified scheme handles:
...
+this scheme specifies how to set the relaxation timescales (τₗ, τᵢ) for
+liquid condensation/evaporation and ice deposition/sublimation.
+NonEq_Constant sets τₗ and τᵢ are determined by the τ_relax arguments in
+lcl and icl, whereas NonEq_N calculates τₗ and τᵢ from liquid and ice
+number concentration.
+"""
+abstract type NonEqScheme end
+struct NonEq_Constant <: NonEqScheme end
+struct NonEq_N <: NonEqScheme end
+
```
sajjadazimi, line 93: Maybe it's better to move this struct definition to NoneqMicrophysics

src/BulkMicrophysicsTendencies.jl
```diff
@@ -100,8 +114,18 @@ end
 
 """
     bulk_microphysics_tendencies(
-        ::Microphysics1Moment, mp, tps,
-        ρ, T, q_tot, q_lcl, q_icl, q_rai, q_sno,
+        ::Microphysics1Moment,
+        noneq_scheme::NonEqScheme,
```
sajjadazimi, line 118: `noneq_scheme` could be an element of Microphysics1Moment. I see Microphysics1Moment as an umbrella that could define what submodels should be including the non equilibrium scheme. 
This is not a strong opinion, so ignore if it's not possible.

src/BulkMicrophysicsTendencies.jl
```diff
@@ -177,12 +212,22 @@ This is a pure function of local thermodynamic state, suitable for:
 
     # --- Cloud condensate formation (non-equilibrium) ---
 
+    if noneq_scheme isa NonEq_Constant
+        τₗ = lcl.τ_relax
+        τᵢ = icl.τ_relax
+    elseif noneq_scheme isa NonEq_N
+        τₗ = CMNonEq.τ_N(q_icl, N_lcl, lcl.ρw, aps.D_vapor)
+        τᵢ = CMNonEq.τ_Frostenberg(icl, aps, ip, q_icl, T)
+    end
```
sajjadazimi, line 221: For code readability, you could dispatch over noneq_scheme and return the tuple of tau_l and tau_i

src/BulkMicrophysicsTendencies.jl
```diff
@@ -685,7 +733,7 @@ Used by both warm-only and warm+ice dispatch methods to reduce code duplication.
 - `dn_lcl_dt`: Cloud number tendency (1/kg/s)
 - `dn_rai_dt`: Rain number tendency (1/kg/s)
 """
-@inline function warm_rain_tendencies_2m(warm_rain, tps, T, q_tot, q_lcl, q_rai, q_ice, ρ, n_lcl, n_rai)
+@inline function warm_rain_tendencies_2m(warm_rain, noneq_scheme, tps, T, q_tot, q_lcl, q_rai, q_ice, ρ, n_lcl, n_rai)
```
sajjadazimi, line 736: same as 1m: noneq_scheme could live under warm_rain model

src/BulkMicrophysicsTendencies.jl
```diff
@@ -790,7 +841,7 @@ For warm rain + P3 ice, see the method that accepts `Microphysics2MParams{FT, WR
 - `db_rim_dt`: Rime volume tendency (always zero for warm-only)
 """
 @inline function bulk_microphysics_tendencies(
-    ::Microphysics2Moment, mp::CMP.Microphysics2MParams{WR, Nothing}, tps,
+    ::Microphysics2Moment, noneq_scheme::NonEqScheme, mp::CMP.Microphysics2MParams{WR, Nothing}, tps,
```
sajjadazimi, line 844: same as above

src/MicrophysicsNonEq.jl
```diff
@@ -37,8 +40,34 @@ Returns the relaxation timescale for phase change processes.
 - Relaxation timescale [s] for condensation/evaporation (liquid) or
   deposition/sublimation (ice)
 """
-@inline τ_relax(p::CondEvap) = p.τ_relax
-@inline τ_relax(p::SubDep) = p.τ_relax
+@inline τ_relax(p::CMP.CloudLiquid) = p.τ_relax
+@inline τ_relax(p::CMP.CloudIce) = p.τ_relax
```
sajjadazimi, line 44: Here, you could dispatch over non-eq method too

sajjadazimi (commented):
Thank you for refactoring and adding the options for relaxation time scale. It looks good overall. I left a few minor comments.

docs/src/MicrophysicsNonEq.md
```diff
@@ -127,6 +127,41 @@ include("plots/NonEqCondEvapRate.jl")
 
 
 
+### Liquid and ice relaxation timescales as functions of number concentration
+
+We also provide functionality to set the relaxation timescales of liquid and ice, ``\tau_l`` and ``\tau_i``, as functions of cloud droplet number concentrations ``N_l`` and ``N_i`` following [MorrisonGrabowski2008_supersat](@cite) and [MorrisonMilbrandt2015](@cite). First, we approximate average radius of droplets from mass and number:
```
trontrytel, line 132:
```suggestion
We also provide functionality to set the relaxation timescales of liquid and ice, ``\tau_l`` and ``\tau_i``, as functions of cloud droplet and ice crystal number concentrations ``N_l`` and ``N_i`` following [MorrisonGrabowski2008_supersat](@cite) and [MorrisonMilbrandt2015](@cite). First, we approximate average radius of droplets from mass and number:
```

src/BulkMicrophysicsTendencies.jl
trontrytel: I agree. Otherwise we will start to have to carry more parameters in Microphysics1Moment than what we actually need for a single simulation

PR #683, opened 2026-02-13
Remove unused NonEq formulation

oalcabes:


trontrytel:
I agree it's not used in Atmos, but we might want to keep it as an option. In case we want to run something with a simple relaxation rate

PR #690, opened 2026-02-24
Add 2M derivatives and tests

trontrytel:
This PR
- adds a "placeholder" 2M derivatives,
- adds derivatives tests,
- rearranges the order of tests for bulk microphysics tendencies (all the old tests stay unchanged, the diff just got very confused)

I'll bump the patch version for the new release when we are ready with it.


tapios (approved):
Looks good. Thank you!

(Please also bump the patch release.)

PR #691, opened 2026-02-26
Added sink derivatives

tapios:
## Purpose 

Adds (diagonal) derivative terms for condensate sinks, to help prevent negative condensates from forming.

## Content

----


trontrytel (approved):
We should also update docs and tests, etc. But I can do that after we have a version that works

PR #708, opened 2026-03-25
Use condensate in evaporation/sublimation

trontrytel:
Writes evaporation and sublimation as proportional to the condensate mass. Deposition and condensation are still proportional to saturation excess.


sajjadazimi (approved):
Thanks!

PR #711, opened 2026-04-14
Average bulk microphysics tendencies

sajjadazimi:
## Purpose 
Add average bulk microphyiscs 1m tendency based on linearized microphysics:
- Transfers → `D * q_donor`
- Vapor sources → constants
- Sinks → `-D * q` (exponential decay over time);

----


src/BulkMicrophysicsTendencies.jl
```diff
@@ -279,6 +279,279 @@ This is a pure function of local thermodynamic state, suitable for:
...
+    e2 = zero(FT)
+    e4 = zero(FT)
+
+    # ------------------------------------------------------------
+    # Vapor -> cloud condensate: constant sources
+    # ------------------------------------------------------------
+    S_lcl_cond = CMNonEq.conv_q_vap_to_q_lcl_icl_MM2015(
+        lcl, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T,
+    )
+    if S_lcl_cond >= zero(FT)
```
trontrytel: I'm never sure about GPU performance. But I think writing it as:

```
is_source = S_lcl_cond >= zero(FT)
D = (-S_lcl_cond) / max(eps(FT), q_lcl)
e1 += ifelse(is_source, S_lcl_cond, zero(FT))
M11 -= ifelse(is_source, zero(FT), D)
```

will avoid branch divergence on GPU.

Same pattern appears in a couple of other places.

src/BulkMicrophysicsTendencies.jl
```diff
@@ -279,6 +279,279 @@ This is a pure function of local thermodynamic state, suitable for:
...
+The system is solved using a specialized sparse structure (specific to the microphysics 1M model):
+- `q_lcl`, `q_icl`: independent scalar solves
+- `(q_rai, q_sno)`: 2×2 coupled solve
+
+Because sinks are linearized as `-D q`, this method effectively integrates
+them as exponential decays over the timestep, improving stability for stiff processes.
+
+Returns a `NamedTuple` of averaged tendencies for all species.
+"""
+@inline function average_bulk_microphysics_tendencies(
```
trontrytel: Do we want to export it? (Same for the linearized operator)

src/BulkMicrophysicsTendencies.jl
```diff
@@ -279,6 +279,279 @@ This is a pure function of local thermodynamic state, suitable for:
...
+        M34 += D
+    else
+        # rain -> snow
+        D = S_accr_rai_sno / max(eps(FT), q_rai)
+        M33 -= D
+        M43 += D
+    end
+
+    # ------------------------------------------------------------
+    # Rain evaporation: sink to vapor if negative, else constant source
```
trontrytel: I think we don't consider condensational growth on rain. So it should be just zero, not constant source

src/BulkMicrophysicsTendencies.jl
```diff
@@ -279,6 +279,279 @@ This is a pure function of local thermodynamic state, suitable for:
...
+    if is_warm
+        M31 += D
+    else
+        M41 += D
+    end
+
+    α = warm_accretion_melt_factor(tps, sno, T)
+    S_accr_melt = α * S_accr_lcl_sno
+    # this is snow -> rain melt induced by warm collected liquid
+    # donor is snow
```
trontrytel, line 394: Isn't cloud liquid also a donor in this process? (liq+sno->rai). I think it's ok as is, but may lead to some inconsistent damping maybe?

src/BulkMicrophysicsTendencies.jl
```diff
@@ -279,6 +279,279 @@ This is a pure function of local thermodynamic state, suitable for:
...
+
+    q_lcl_new = b1 / a11
+    q_icl_new = b2 / a22
+
+    # Reduced 2x2 system for q_rai_new, q_sno_new
+    r3 = muladd(-a31, q_lcl_new, b3)
+    r4 = muladd(-a41, q_lcl_new, muladd(-a42, q_icl_new, b4))
+
+    det = muladd(-a34, a43, a33 * a44)
+    q_rai_new = (r3 * a44 - a34 * r4) / det
```
trontrytel, line 528: Could you leave a comment somewhere explaining why det > 0

test/bulk_tendencies_tests.jl
```diff
@@ -631,6 +631,262 @@ function test_bulk_microphysics_1m_tendencies(FT)
 
 end
 
+function test_average_bulk_microphysics_1m_tendencies(FT)
```
trontrytel, line 634: Could we add a test or a documentation plot that shows how the average rate depends on dt going from 1s to 100s or something? To try to visualize the average damping we are getting. 

Would be good to also add a test when all inputs are zero.

And add a documentation page. But this we can do once we are settled on the final solution

trontrytel (approved):
Hitting approve, so I don't slow you down when away!

src/BulkMicrophysicsTendencies.jl
sajjadazimi: Good point. The liquid contribution is already accounted for in the preceding liq/snow accretion term (liq → sno or liq → rai depending on temperature). Here this term represents the additional conversion of snow to rain triggered by those collisions (i.e., melt of snow in warm conditions). So in this step the effective donor is only snow.

test/bulk_tendencies_tests.jl
sajjadazimi: I added two unit tests for dt damping and zero input. I leave adding a documentation page for later when things work.

PR #713, opened 2026-05-08
`@noinline` `average_bulk_microphysics_tendencies` to reduce register pressure

### LINKED ISSUE #2632: Improve performance of kernel on L1013 of ClimaAtmos `set_microphysics_tendency_cache!`
```julia
        @. ᶜmp_tendency⁰ = microphysics_tendencies_1m(
            BMT.Microphysics1Moment(), sgs_quad, cmp, thp, ᶜρ⁰, ᶜT⁰,
            ᶜq_tot_nonneg⁰, ᶜq_lcl⁰, ᶜq_icl⁰, ᶜq_rai⁰, ᶜq_sno⁰,
            ᶜT′T′, ᶜq′q′, corr_Tq, ᶜsgs_moments.λ_lagrange, α,
            dt, nsubs_quad, ᶜλ⁰, ᶜmu_S⁰,
        )
```

As of now, this kernel is 18% of the flagship AMIP GPU time.

Putting this issue in ClimaCore so we can attempt to solve it as low as possible in the stack.

## Potential solutions

1. CUDA launch bounds instead of max regs (#2601)
2. Kernel fusion, noinlining in CloudMicrophysics (https://github.com/CliMA/CloudMicrophysics.jl/pull/769)
3. `foreach_point` macro (https://github.com/CliMA/ClimaAtmos.jl/issues/4745)
4. Reduce SGS quadrature order from 3 to 2

## Experiments

My experiment index is [here](https://calkit.io/petebachant/clima-gpu-profiling/tables?path=experiments.csv).

petebachant:
This kernel is now the hottest in prog EDMF 1M AMIP by a long shot, and this change produces a ~10% speedup ([kernel analysis notebook](https://calkit.io/petebachant/clima-gpu-profiling/notebooks?ref=04492d5&path=notebooks%2Fcompare-kernels.ipynb)). Disclaimer: Explanatory comments written by Claude--I don't yet have a deep understanding of what's going on here!

Resolves https://github.com/CliMA/ClimaCore.jl/issues/2632


dennisYatunin (approved):
Claude's explanation is a bit suspicious given that the quadrature loop isn't being unrolled, but a 10% speedup sounds great! I'll think about how we can turn this into a simpler example for ClimaCore's compiler stress tests.

petebachant:
Hypothesis: If Claude is correct, the issue comes from the quadrature loop being unrolled, so we de-unroll that we may be able to get these benefits without the performance hit on CPU.

Might be possible by dropping quadrature order from the type. Move from type into the value. Type can be int. Value is number of quadrature loops.

trontrytel:
Is this PR something that should be merged or closed?

petebachant:
I opened https://github.com/CliMA/ClimaAtmos.jl/pull/4503 to retain the GPU performance gains and move changes to Atmos and avoid the 1.12 regression, but that one is a little uglier. Any preference from your end?

trontrytel:
No preference. Whichever option you think is better?

trontrytel:
Although I'm making some changes here: https://github.com/CliMA/CloudMicrophysics.jl/pull/717 where are grouped some output into a tuple. And Claude thinks that the performance will not be affected only if I keep inlining...

petebachant:
I'm actually curious if the compiler will do a better job on either device if there is no macro on the function. I will give it a try, and if not, close this PR and focus on Atmos. It's not great to have to dig this deep into the stack for the performance gains.

petebachant:
Running with no macro produced no performance change. Closing this and pursuing in Atmos.

trontrytel:
Thank you!

PR #714, opened 2026-05-14
Adding microphysics options to bulk tendency (+ temperature dependent cloud ice formation)

trontrytel:
### Core Design Change: Dispatch-Based Process Options

This PR introduces a **`Microphysics1MOptions`**. Each microphysical process (condensation, autoconversion, evaporation, etc.) is now selected at configuration time via a singleton type, enabling compile-time specialization from the point of view of ClimaAtmos.

**8 configurable processes:**

| Process | Options |
|---------|---------|
| Cloud liquid formation | `ConstantTimescaleCloudLiquidFormation` |
| Cloud ice formation | `ConstantTimescaleCloudIceFormation`, `TemperatureDependentCloudIceFormation` (**new**) |
| Cloud ice melt | `CloudIceMeltToLiquid` (**new**), `NoCloudIceMelt`  |
| Liquid autoconversion | `LiquidAutoconv1M`, `LiquidAutoconv2M` |
| Snow autoconversion | `SnowAutoconvNoSupersat`, `SnowAutoconvWithSupersat` |
| Rain evaporation | `EvaporationOnly` |
| Snow sublimation | `SublimationOnly`, `DepositionSublimation` |
| Snow melt | `SnowMelt` |

The change on the ClimaAtmos side is very minimal and consists of:
- adding flags to config yaml file
- adding code to parse those flags into microphysics options struct
- creating microphysics parameters based on options.

See the draft here: https://github.com/CliMA/ClimaAtmos.jl/pull/4502

### New Physics

1. **Frostenberg (2023) temperature-dependent ice formation** — Uses INP concentration from `HetIceNucleation` to compute a physically-based relaxation timescale for ice deposition (replacing the constant timescale). Selected via `TemperatureDependentCloudIceFormation`.

2. **Cloud ice melt to liquid** — New process where cloud ice melts to cloud liquid above freezing (`conv_q_icl_to_q_lcl`). Previously not represented. Adds `M12` coupling in the implicit solver Jacobian.

### API Refactoring

- **`Microphysics1MParams`** now carries an `options::Microphysics1MOptions` field and conditionally builds sub-parameters (`autoconv_2M`, `frostenberg2023`) based on the selected options.
- **Stale kwarg removed**: `with_2M_autoconv = true` → `options = Microphysics1MOptions(...)`.
- **Unified function signatures**: Process functions like `evaporation_sublimation`, `snow_melt`, `conv_q_icl_to_q_sno_no_supersat` are renamed to consistent `conv_X_to_Y(option, mp, tps, micro, thermo)` form, where `micro = (; q_tot, q_lcl, q_icl, q_rai, q_sno)` and `thermo = (; ρ, T)`.
- **`BulkMicrophysicsTendencies.jl`** dispatches on `mp.options` for every process call, making the tendency computation fully configurable from the options struct.

### Implicit Solver Update

The Jacobian matrix in `average_bulk_microphysics_tendencies` gains an `M12` entry for ice→liquid melt coupling, and autoconversion/sublimation entries are updated to use the new dispatched functions.

### DIFF
--- a/src/BulkMicrophysicsTendencies.jl
+++ b/src/BulkMicrophysicsTendencies.jl
@@ -167,7 +167,11 @@ This is a pure function of local thermodynamic state, suitable for:
     ce = mp.collision
     aps = mp.air_properties
     vel = mp.terminal_velocity
-    var = mp.autoconv_2M
+    opts = mp.options
+
+    # Construct state tuples (reused across all process calls)
+    micro = (; q_tot, q_lcl, q_icl, q_rai, q_sno)
+    thermo = (; ρ, T)
 
     # Initialize tendencies
     dq_lcl_dt = zero(T)
@@ -178,25 +182,28 @@ This is a pure function of local thermodynamic state, suitable for:
     # --- Cloud condensate formation (non-equilibrium) ---
 
     # Condensation/evaporation of cloud liquid
-    S_lcl_cond = CMNonEq.conv_q_vap_to_q_lcl_icl_MM2015(lcl, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T)
+    S_lcl_cond = CMNonEq.conv_q_vap_to_q_lcl(opts.cloud_liquid_formation, mp, tps, micro, thermo)
     dq_lcl_dt += S_lcl_cond
 
-    # Deposition/sublimation of cloud ice (INP limiter applied inside conv_q_vap_to_q_lcl_icl_MM2015)
-    S_icl_dep = CMNonEq.conv_q_vap_to_q_lcl_icl_MM2015(icl, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T)
+    # Deposition/sublimation of cloud ice
+    S_icl_dep = CMNonEq.conv_q_vap_to_q_icl(opts.cloud_ice_formation, mp, tps, micro, thermo)
     dq_icl_dt += S_icl_dep
 
     # --- Autoconversion (cloud → precipitation) ---
 
     # Cloud liquid → rain
-    # Use 2M autoconversion when N_lcl > 0 and var provided, otherwise 1M
     S_acnv_1M = CM1.conv_q_lcl_to_q_rai(rai.acnv1M, q_lcl, true)
-    S_acnv_2M = isnothing(var) ? S_acnv_1M : CM2.conv_q_lcl_to_q_rai(var, q_lcl, ρ, N_lcl)
-    S_acnv_lcl = ifelse(N_lcl > zero(N_lcl), S_acnv_2M, S_acnv_1M)
+    S_acnv_lcl = if opts.cloud_liquid_autoconversion isa CMP.LiquidAutoconv2M
+        S_acnv_2M = CM2.conv_q_lcl_to_q_rai(mp.autoconv_2M, q_lcl, ρ, N_lcl)
+        ifelse(N_lcl > zero(N_lcl), S_acnv_2M, S_acnv_1M)
+    else
+        S_acnv_1M
+    end
     dq_lcl_dt -= S_acnv_lcl
     dq_rai_dt += S_acnv_lcl
 
-    # Cloud ice → snow (no supersaturation version for simplicity)
-    S_acnv_icl = CM1.conv_q_icl_to_q_sno_no_supersat(sno.acnv1M, q_icl, true)
+    # Cloud ice → snow
+    S_acnv_icl = CM1.conv_q_icl_to_q_sno(opts.snow_autoconversion, mp, tps, micro, thermo)
     dq_icl_dt -= S_acnv_icl
     dq_sno_dt += S_acnv_icl
 
@@ -261,17 +268,22 @@ This is a pure function of local thermodynamic state, suitable for:
     # --- Evaporation and sublimation ---
 
     # Rain evaporation
-    S_evap_rai = CM1.evaporation_sublimation(rai, vel.rain, aps, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T)
+    S_evap_rai = CM1.conv_q_rai_to_q_vap(opts.rain_evaporation, mp, tps, micro, thermo)
     dq_rai_dt += S_evap_rai  # negative tendency (evaporation)
 
-    # Snow sublimation/deposition
-    S_subl_sno = CM1.evaporation_sublimation(sno, vel.snow, aps, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T)
-    dq_sno_dt += S_subl_sno  # can be positive (deposition) or negative (sublimation)
+    # Snow sublimation (or sublimation + deposition, depending on option)
+    S_subl_sno = CM1.conv_q_sno_to_q_vap(opts.snow_sublimation, mp, tps, micro, thermo)
+    dq_sno_dt += S_subl_sno
 
     # --- Melting ---
 
+    # Cloud ice melt → cloud liquid
+    S_melt_icl = CM1.conv_q_icl_to_q_lcl(opts.cloud_ice_melt, mp, tps, micro, thermo)
+    dq_icl_dt -= S_melt_icl
+    dq_lcl_dt += S_melt_icl
+
     # Snow melt → rain
-    S_melt_sno = CM1.snow_melt(sno, vel.snow, aps, tps, q_sno, ρ, T)
+    S_melt_sno = CM1.conv_q_sno_to_q_rai(opts.snow_melt, mp, tps, micro, thermo)
     dq_sno_dt -= S_melt_sno
     dq_rai_dt += S_melt_sno
 
@@ -314,7 +326,11 @@ Returns a `NamedTuple` containing the nonzero entries of `M` and `e`.
     ce = mp.collision
     aps = mp.air_properties
     vel = mp.terminal_velocity
-    var = mp.autoconv_2M
+    opts = mp.options
+
+    # Construct state tuples (reused across all process calls)
+    micro = (; q_tot, q_lcl, q_icl, q_rai, q_sno)
+    thermo = (; ρ, T)
 
     FT = promote_type(
         typeof(ρ), typeof(T), typeof(q_tot), typeof(q_lcl),
@@ -322,6 +338,7 @@ Returns a `NamedTuple` containing the nonzero entries of `M` and `e`.
     )
 
     M11 = zero(FT)
+    M12 = zero(FT)
     M22 = zero(FT)
     M31 = zero(FT)
     M33 = zero(FT)
@@ -340,33 +357,41 @@ Returns a `NamedTuple` containing the nonzero entries of `M` and `e`.
     # ------------------------------------------------------------
     # Vapor -> cloud condensate: constant sources
     # ------------------------------------------------------------
-    S_lcl_cond = CMNonEq.conv_q_vap_to_q_lcl_icl_MM2015(
-        lcl, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T,
-    )
+    S_lcl_cond = CMNonEq.conv_q_vap_to_q_lcl(opts.cloud_liquid_formation, mp, tps, micro, thermo)
     D = S_lcl_cond / max(q_min, q_lcl)
     is_source = S_lcl_cond >= zero(FT)
     e1 += ifelse(is_source, S_lcl_cond, zero(FT))
     M11 += ifelse(is_source, zero(FT), D)
 
-    S_icl_dep = CMNonEq.conv_q_vap_to_q_lcl_icl_MM2015(
-        icl, tps, q_tot, q_lcl, q_icl, q_rai, q_sno, ρ, T,
-    )
+    S_icl_dep = CMNonEq.conv_q_vap_to_q_icl(opts.cloud_ice_formation, mp, tps, micro, thermo)
     D = S_icl_dep / max(q_min, q_icl)
     is_source = S_icl_dep >= zero(FT)
     e2 += ifelse(is_source, S_icl_dep, zero(FT))
     M22 += ifelse(is_source, zero(FT), D)
 
     # ------------------------------------------------------------
+    # Cloud ice melt: ice → cloud liquid
+    # ------------------------------------------------------------
+    S_melt_icl = CM1.conv_q_icl_to_q_lcl(opts.cloud_ice_melt, mp, tps, micro, thermo)
+    D = S_melt_icl / max(q_min, q_icl)
+    M22 -= D
+    M12 += D
+
+    # ------------------------------------------------------------
     # Autoconversion: donor-based transfer
     # ------------------------------------------------------------
     S_acnv_1M = CM1.conv_q_lcl_to_q_rai(rai.acnv1M, q_lcl, true)
-    S_acnv_2M = isnothing(var) ? S_acnv_1M : CM2.conv_q_lcl_to_q_rai(var, q_lcl, ρ, N_lcl)
-    S_acnv_lcl = ifelse(N_lcl > zero(N_lcl), S_acnv_2M, S_acnv_1M)
+    S_acnv_lcl = if opts.cloud_liquid_autoconversion isa CMP.LiquidAutoconv2M
+        S_acnv_2M = CM2.conv_q_lcl_to_q_rai(mp.autoconv_2M, q_lcl, ρ, N_lcl)
+        ifelse(N_lcl > zero(N_lcl), S_acnv_2M, S_acnv_1M)
+    else
+        S_acnv_1M
+    end
     D = S_acnv_lcl / max(q_min, q_lcl)
     M11 -= D
     M31 += D
 
-    S_acnv_icl = CM1.conv_q_icl_to_q_sno_no_supersat(sno.acnv1M, q_icl, true)
+    S_acnv_icl = CM1.conv_q_icl_to_q_sno(opts.snow_autoconversion, mp, tps, micro, thermo)
     D = S_acnv_icl / max(q_min, q_icl)
     M22 -= D
     M42 += D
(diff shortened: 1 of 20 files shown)


src/BulkMicrophysicsTendencies.jl
```diff
@@ -178,25 +182,28 @@ This is a pure function of local thermodynamic state, suitable for:
...
     # --- Autoconversion (cloud → precipitation) ---
 
     # Cloud liquid → rain
-    # Use 2M autoconversion when N_lcl > 0 and var provided, otherwise 1M
     S_acnv_1M = CM1.conv_q_lcl_to_q_rai(rai.acnv1M, q_lcl, true)
-    S_acnv_2M = isnothing(var) ? S_acnv_1M : CM2.conv_q_lcl_to_q_rai(var, q_lcl, ρ, N_lcl)
-    S_acnv_lcl = ifelse(N_lcl > zero(N_lcl), S_acnv_2M, S_acnv_1M)
+    S_acnv_lcl = if opts.cloud_liquid_autoconversion isa CMP.LiquidAutoconv2M
+        S_acnv_2M = CM2.conv_q_lcl_to_q_rai(mp.autoconv_2M, q_lcl, ρ, N_lcl)
+        ifelse(N_lcl > zero(N_lcl), S_acnv_2M, S_acnv_1M)
```
sajjadazimi, line 198: Not necessarily related to this PR and also discussed offline; but just to put it in writing somewhere, I think we should move this function to CM1 and also make N_lcl a parameter instead of an argument.

src/BulkMicrophysicsTendencies.jl
sajjadazimi: Now `q_lcl` and `q_icl` are also solved from a coupled 2-by-2 system.

src/BulkMicrophysicsTendencies.jl
```diff
@@ -513,15 +535,17 @@ exponential decays over the substep.
     b3 = invΔt * q_rai
     b4 = lin.e4 + invΔt * q_sno
 
-    q_lcl_new = b1 / a11
-    q_icl_new = b2 / a22
+    # Solve 2×2 system for q_lcl, q_icl (coupled via ice melt M12)
+    det12 = a11 * a22  # a21 = 0
+    q_lcl_new = (b1 * a22 - a12 * b2) / det12
+    q_icl_new = a11 * b2 / det12
```
sajjadazimi, line 540: This is correct, but can also be simplified to:
`q_icl_new = b2 / a22`
`q_lcl_new = (b1 - a12 * q_icl_new) / a11`
both are fine.

sajjadazimi (approved):
Thank you! The code seems much cleaner now. The new ice-to-liquid term in the average bulk tendency solver is treated correctly.

trontrytel:
I bumped a couple of performance tests. I have seen them failing on PRs that were only touching documentation etc.  The failures were mostly from julia 1.12 running on windows, and seem somewhat consistent

src/parameters/Microphysics1MOptions.jl
```diff
@@ -0,0 +1,116 @@
...
+"""Snow melts to rain above freezing."""
+struct SnowMelt <: MicrophysicsOption end
+
+"""
+    Microphysics1MOptions{CLF, CIF, CIM, CLA, SA, RE, SS, SM}
+
+Process configuration for 1-moment microphysics.
+Each field selects a variant for one microphysical process.
+
+Defaults match the baseline (old main branch) behavior.
```
haakon-e, line 69: With cloud ice melt to liquid being a new process, and `cloud_ice_melt::CIM = CloudIceMeltToLiquid()` being default, below, it may be more accurate to remove this statement.

src/parameters/Microphysics1MOptions.jl
```diff
@@ -0,0 +1,116 @@
...
+    EvaporationOnly,
+    SublimationOnly,
+    DepositionSublimation,
+    SnowMelt
+"""
+    MicrophysicsOption
+
+Abstract type for all microphysics process options.
+"""
+abstract type MicrophysicsOption end
```
haakon-e, line 21: Because all option structs share this abstract type, you could accidentally write nonsensical code like:
```julia
Microphysics1MOptions(cloud_liquid_autoconversion = SnowMelt())
```
with no warning until `bulk_microphysics_tendencies` is actually executed and you get a `MethodError`.

Another option would be to provide separate abstract types for each process, guarding against this at model construction time.

src/BulkMicrophysicsTendencies.jl
haakon-e, line 311: We should better document what is required to be added to this function when a new tendency is introduced. Now, you either have to read the source code very closely, or already know what you're doing in order to add new tendencies. 

While most people would hopefully understand that they need to:
- Define a singleton "option" struct
- add the forward dispatch
- add the dispatch site in the forward tendency

adding the matching `M_ij / e_i` entries are opaque.

src/BulkMicrophysicsTendencies.jl
```diff
@@ -704,7 +727,12 @@ Used by both warm-only and warm+ice dispatch methods to reduce code duplication.
     dn_rai_dt = zero(FT)
 
     # --- Condensation of vapor / evaporation of cloud liquid water ---
-    ∂ₜq_lcl_cond = CMNonEq.conv_q_vap_to_q_lcl_icl_MM2015(condevap, tps, q_tot, q_lcl, q_ice, q_rai, zero(q_ice), ρ, T)
+    mp_mock = (; cloud = (; liquid = condevap))
```
haakon-e, line 730: It seems like the need for this may point to some discrepancies between the design of the 1M and 2M schemes that we should think about before applying the pattern more broadly.

At least, the docstring for `conv_q_vap_to_q_lcl` should explain which fields of `mp` it needs

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,56 +35,90 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
+mp = CMP.Microphysics1MParams(Float64;
+    options = CMP.Microphysics1MOptions(
+        cloud_liquid_autoconversion = CMP.LiquidAutoconv2M(),
+    ),
+)
 ```
 """
-@kwdef struct Microphysics1MParams{FT, CP, PP, CE, AP, VL, VA} <: ParametersType
+@kwdef struct Microphysics1MParams{FT, OPT, CP, PP, CE, AP, VL, VA, FR} <: ParametersType
+    options::OPT
```
haakon-e, line 86: This adds 8 type fields to this struct (i.e. the concrete type of `OPT` includes the type space of `Microphysics1MOptions`). Because the full type is printed in stack tracers, this adds hundreds of characters, which may reduce legibility.

haakon-e, line 86: There's a small regression baked into this: if downstream code called `eltype(::Microphysics1MParams)`, they now get `Any` instead of `FT` since `Microphysics1MParams <: ParametersType` and `eltype` for the abstract type is defined as:
```julia
Base.eltype(p::ParametersType) = fieldcount(typeof(p)) > 0 ? eltype(getfield(p, 1)) : Any
```
In practice, it's possible no one uses this pattern, as we have tried to push toward code getting type from the prognostic variables instead of the parameters.

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,56 +35,90 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
+function Microphysics1MParams(toml_dict::CP.ParamDict; options = Microphysics1MOptions())
     (; prescribed_cloud_droplet_number_concentration) = CP.get_parameter_values(
         toml_dict, "prescribed_cloud_droplet_number_concentration", "CloudMicrophysics",
     )
+    # Conditional construction driven by options
+    autoconv_2M = options.cloud_liquid_autoconversion isa LiquidAutoconv2M ?
+                  VarTimescaleAcnv(toml_dict) : nothing
+    frostenberg2023 =
+        options.cloud_ice_formation isa TemperatureDependentCloudIceFormation ?
+        Frostenberg2023(toml_dict) : nothing
```
haakon-e, line 117: This contract, i.e. the correspondence between e.g. `TemperatureDependentCloudIceFormation` in `Microphysics1MOptions` and `Frostenberg2023` in `Microphysics1MParams` is only enforced if this constructor is used. This results in the possibility of constructing a technically valid `Microphysics1MParams` which then error inside e.g. `conv_q_lcl_to_q_rai` because the parameter struct is missing certain fields.

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,56 +35,90 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
+- `options`: Process configuration (default: baseline behavior)
 """
-function Microphysics1MParams(toml_dict::CP.ParamDict; with_2M_autoconv = false)
+function Microphysics1MParams(toml_dict::CP.ParamDict; options = Microphysics1MOptions())
     (; prescribed_cloud_droplet_number_concentration) = CP.get_parameter_values(
         toml_dict, "prescribed_cloud_droplet_number_concentration", "CloudMicrophysics",
     )
+    # Conditional construction driven by options
+    autoconv_2M = options.cloud_liquid_autoconversion isa LiquidAutoconv2M ?
+                  VarTimescaleAcnv(toml_dict) : nothing
```
haakon-e, line 114: Expanding beyond this, it may be cleaner to dispatch, e.g.
```julia
sub_params(::LiquidAutoconv2M, toml_dict) = VarTimescaleAcnv(toml_dict)
sub_params(_, toml_dict) = nothing  # default case
```

src/BulkMicrophysicsTendencies.jl
```diff
@@ -167,7 +167,11 @@ This is a pure function of local thermodynamic state, suitable for:
     ce = mp.collision
     aps = mp.air_properties
     vel = mp.terminal_velocity
-    var = mp.autoconv_2M
+    opts = mp.options
+
+    # Construct state tuples (reused across all process calls)
+    micro = (; q_tot, q_lcl, q_icl, q_rai, q_sno)
+    thermo = (; ρ, T)
```
haakon-e, line 174: We should clearly document somewhere what fields are expected to be contained within these NamedTuples

src/parameters/Microphysics1MOptions.jl
```diff
@@ -0,0 +1,116 @@
...
+
+"""Use a constant relaxation timescale for liquid condensation and evaporation."""
+struct ConstantTimescaleCloudLiquidFormation <: MicrophysicsOption end
+
+"""Use a constant relaxation timescale for ice deposition and sublimation."""
+struct ConstantTimescaleCloudIceFormation <: MicrophysicsOption end
+
+"""Use the INP-dependent Frostenberg (2023) timescale for deposition,
+   with constant timescale for sublimation."""
+struct TemperatureDependentCloudIceFormation <: MicrophysicsOption end
```
haakon-e, line 31: This name seems to suggest a broader class of temperature-dependent schemes, while the docstring clarifies that it is Frostenberg 2023 for deposition and constant for sublimation. I think some other name may more clearly indicate this specific parameterization.

haakon-e (commented):
Great work rethinking the established patterns for managing the individual parameterizations in the 1M scheme. Making them easy to switch and activate/deactivate is the right direction, and the clean YAML-to-singletons path on the Atmos side is genuinely useful.

My main concern is that the PR introduces a fourth name for each parameterization: `temperature_dependent` (YAML), `TemperatureDependentCloudIceFormation` (dispatch singleton), `Frostenberg2023` (parameters), `conv_q_vap_to_q_icl` (function). Developers must keep these aligned when extending CloudMicrophysics and interfacing with ClimaAtmos.

I'd suggest collapsing these to one. Make each parameterization a single struct that (a) carries its own parameters, (b) is callable as the tendency, and (c) is a field of `Microphysics1MParams` directly. So there's no separate options struct. This could be followed all the way through, having the YAML value shadow the user-friendly type name.

A sketch for this suggestion:
```julia
# src/MicrophysicsNonEq.jl
struct TemperatureDependentCloudIceFormation
    σ::FT
    # other parameters ...
end
function (parameterization::TemperatureDependentCloudIceFormation)(mp, tps, micro, thermo)
    # contents of conv_q_vap_to_q_icl ...
    (; σ) = parameterization
    # ...
    S_icl_dep = (qᵥ - qᵥ_sat_ice) / (τ_sub * Γᵢ)
    return S_icl_dep
end

# src/Microphysics1MParams.jl
struct Microphysics1MParams{FT, CIF, ...} 
    cloud_ice_formation::CIF
    # other parameterization structs...
end

# src/BulkMicrophysicsTendencies.jl
function bulk_microphysics_tendencies(
    ::Microphysics1Moment, mp::CMP.Microphysics1MParams, tps, micro, thermo,
)
    micro = UT.clamp_to_nonneg(micro)
    thermo = UT.clamp_to_nonneg(thermo)
    # ...
    S_icl_dep = mp.cloud_ice_formation(mp, tps, micro, thermo)
    dq_icl_dt += S_icl_dep
    # ...
    return ...
end
```
and yaml config:
```yaml
cloud_ice_formation:
  help: "Cloud ice formation timescale option for 1M microphysics [`ConstantCloudIceFormation` (default), `TemperatureDependentCloudIceFormation`]"
  value: "ConstantCloudIceFormation"
```

trontrytel:
Regarding the top comment - I agree about the name proliferation being confusing. I'll try to clean that up to at least make the names match. 

I think the suggested pattern works well for small parameterizations that are self contained. Unfortunately there are other processes that co-depend on many parameters. I don't think this approach would generalize well there. 

Examples of self contained processes:
- Cloud liquid formation — just a timescale τ
- Cloud ice formation — timescale or Frostenberg-specific params

Examples of shared parameters processes:
- Rain evaporation — needs rain terminal velocity, ventilation, particle mass/area
- Snow sublimation — needs snow terminal velocity, ventilation, particle mass/area
- Snow melt — needs snow terminal velocity, ventilation, particle mass/area
- All accretion processes — need particle properties from both interacting species plus collision efficiency

For the shared parameters processes we would then have to either duplicate the parameters to make the structs truly self-contained, or still pass in the full parameter struct. I'm guessing you are leaning to the second option, because your example has `mp.cloud_ice_formation(mp, tps, micro, thermo)` as usage example. This is a little confusing to me because the proposed struct lives in `mp` and also gets `mp` as input? cc @haakon-e

haakon-e:
> Examples of shared parameters processes:
> 
> * Rain evaporation — needs rain terminal velocity, ventilation, particle mass/area
> * Snow sublimation — needs snow terminal velocity, ventilation, particle mass/area
> * Snow melt — needs snow terminal velocity, ventilation, particle mass/area
> * All accretion processes — need particle properties from both interacting species plus collision efficiency

If I understand correctly, there are (roughly) two categories of "shared parameter processes":
1. A parameterization needs information about the particle properties (e.g. PSD parameters, mass, sedimentation velocity)
2. Two parameterizations interact (e.g. accretion/collisions)

For the former case, I would imagine these parameters enter as part of the `mp` namedtuple, and do not take part in dispatch (at least not on the tendency level)

For the latter case, I agree it complicates things. One option would be to dispatch over multiple types simultaneously (we do this e.g. when EDMF + microphysics interact in CA.jl), depending on the exact nature of the process. I would need to think more about this for some of the complex cases.

I also agree that the pattern `mp.something(mp, ...)` is strange. Ideally, we'd be able to separate between "parameterizations" (i.e. the `mp.something` part) and particle properties (the rest of `mp`, e.g. PSD, terminal velocity).

src/parameters/Microphysics1MOptions.jl
trontrytel: We landed somewhere in the middle. For options that offer more than one implementation of them we now have an abstract type: For example `CloudIceFormation` (abstract type) now has `ConstantTimescale` `TemperatureDependent` sub types. For options that only offer one implementation I left them as is. We can revisit later, if we run into confusing errors to often from this.

src/BulkMicrophysicsTendencies.jl
trontrytel: Opened as an issue

src/BulkMicrophysicsTendencies.jl
trontrytel: Fixed. Now it looks like this: `CMP.CloudLiquidFormation(condevap.τ_relax), nothing, tps, micro_mock, thermo_mock,`

src/parameters/Microphysics1MParams.jl
trontrytel: Would grouping them more alleviate the problem? Something like:

```
struct AccretionOptions{CLRA, CLSA, CIRA, CISA, RSA}
    cloud_liquid_rain::CLRA
    cloud_liquid_snow::CLSA
    cloud_ice_rain::CIRA
    cloud_ice_snow::CISA
    rain_snow::RSA
end

struct Microphysics1MOptions{PHASE, AUTO, ACCR}  # 3 params instead of 13
    phase_change::PHASE   # groups CLF, CIF, CIM, RCE, SDS, SM
    autoconversion::AUTO  # groups RA, SA
    accretion::ACCR       # groups the 5 accretion types
end
```

trontrytel: Thanks! I'll be on the lookout for this pattern when I update Atmos to all this carnage

haakon-e: I don't think so. But your latest PR does address the core issue the comment was targeting: the introduction of separate options and param structs. With params and options merged into one struct, this is limited insofar as we have a reasonably complex parameterization for 1M microphysics.

trontrytel: Ok. I'm just going through the rest of the comments here to make sure I didn't miss anything. I'll table the grouping for now. Especially in case if we are really serious about the "tendency bag approach" later

src/parameters/Microphysics1MParams.jl
trontrytel: Somewhat addressed in the last iteration. Might revisit later

src/parameters/Microphysics1MParams.jl
trontrytel: Somewhat addressed in the last iteration. Might revisit later

src/parameters/Microphysics1MOptions.jl
trontrytel: Agreed. Lets worry about it when we add a different temperature dependence here

src/BulkMicrophysicsTendencies.jl
trontrytel: Added in the docstrings

PR #725, opened 2026-06-11
Make the pointwise 2M+P3 tendencies ForwardDiff-differentiable

haakon-e:
Implicit (Rosenbrock-type) microphysics substepping needs Jacobians of the pointwise 2M+P3 tendencies w.r.t. the 8 prognostic species, so `ForwardDiff.jacobian` must work through `bulk_microphysics_tendencies` (`logλ` is held fixed). This PR does some groundwork to enable this.

- Modify `P3State` and `Γ_incl` to accept a mix of `Dual` and floating point arguments.
- Implement `Dual` method for `SF.gamma_inc_inv` (JuliaMath/SpecialFunctions.jl#452) locally. Ideally this should be added to `SpecialFunctions`, and I might do so later, but adding here to move forward locally for now. The implementation uses a mix of analytical and central differences of the forward map.
- Ensure `ice_mass_coeffs` returns one tuple type (needed for efficient AD).
- Type early returns and fallback values across the 2M, NonEq, and P3 leaves by the promotion of all the arguments the result derives from (new `Utilities.promote_typeof`), so a caller mixing `Dual` and float arguments does not get a union-typed, heap-boxed return.
- Handle edge cases for `sgs_weight_function` (lower tail rounds 1 - a to 1, so atanh(-1) = -Inf) and SB2006 autoconversion (vertical tangent of `τ^0.7` at zero rain), which previously could produce NaNs.
- Unrelated fix: `get_distribution_logλ_from_prognostic` called an undefined function.

I might try to differentiate through the `logλ` shape solve later; which needs ∂/∂a of the forward `gamma_inc` (JuliaMath/SpecialFunctions.jl#317).

Tests cover Jacobians at four regimes for both float types against per-row central finite differences, the `gamma_inc_inv` rule into the `q = eps()` tail, the regularised-ratio band, and that the touched functions stay concretely typed under single-`Dual` and all-`Dual` argument mixes.

### DIFF
--- a/src/IceNucleation.jl
+++ b/src/IceNucleation.jl
@@ -266,7 +266,7 @@ Compute the rate of liquid water freezing into ice.
     + `∂ₜq_frz`: Specific mass freezing rate [kg(water) kg⁻¹(air) s⁻¹].
 """
 function liquid_freezing_rate(opt::CMP.RainFreezing, pdf, tps, q, ρ, N, T)
-    FT = eltype(q)
+    FT = float(UT.promote_typeof(q, ρ, N, T))
     T_freeze = TDI.TD.Parameters.T_freeze(tps)
     (; ρw) = pdf  # [kg(water) m⁻³(water)]
     n = N / ρ     # specific number concentration [kg⁻¹(air)]
@@ -298,8 +298,8 @@ function liquid_freezing_rate(opt::CMP.RainFreezing, pdf, tps, q, ρ, N, T)
     # Otherwise, return zero.
     ϵₘ, ϵₙ = UT.ϵ_numerics_2M_M(FT), UT.ϵ_numerics_2M_N(FT)
     cond = (n > ϵₙ) & (q > ϵₘ) & (T < T_freeze - 4)
-    ∂ₜn_frz = ifelse(cond, ∂ₜn_frz, zero(n))
-    ∂ₜq_frz = ifelse(cond, ∂ₜq_frz, zero(q))
+    ∂ₜn_frz = ifelse(cond, ∂ₜn_frz, zero(FT))
+    ∂ₜq_frz = ifelse(cond, ∂ₜq_frz, zero(FT))
 
     return (; ∂ₜn_frz, ∂ₜq_frz)
 end
@@ -351,7 +351,7 @@ function liquid_freezing_rate(
     opt::CMP.RainFreezing, pdf::CMP.CloudParticlePDF_SB2006,
     tps, q, ρ, N, T,
 )
-    FT = eltype(q)
+    FT = float(UT.promote_typeof(q, ρ, N, T))
     T_freeze = TDI.TD.Parameters.T_freeze(tps)
     (; ρw) = pdf
     n = N / ρ     # specific number concentration [kg⁻¹(air)]
@@ -375,8 +375,8 @@ function liquid_freezing_rate(
     # Check non-trivial number/mass and that T < -4 °C.
     ϵₘ, ϵₙ = UT.ϵ_numerics_2M_M(FT), UT.ϵ_numerics_2M_N(FT)
     cond = (n > ϵₙ) & (q > ϵₘ) & (T < T_freeze - 4)  # TODO: Make a parameter.
-    ∂ₜn_frz = ifelse(cond, ∂ₜn_frz, zero(n))
-    ∂ₜq_frz = ifelse(cond, ∂ₜq_frz, zero(q))
+    ∂ₜn_frz = ifelse(cond, ∂ₜn_frz, zero(FT))
+    ∂ₜq_frz = ifelse(cond, ∂ₜq_frz, zero(FT))
 
     return (; ∂ₜn_frz, ∂ₜq_frz)
 end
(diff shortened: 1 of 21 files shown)


haakon-e:

- #725 ◀
    - #741
    - #726
        - #731
            - #732
                - #737
                    - #742
                        - #743
                            - #744

src/P3_size_distribution.jl
```diff
@@ -72,7 +72,7 @@ Compute `log(Iᵏ)` where `Iᵏ` is the following integral:
 See also [`gamma_inc_moment`](@ref)
 """
 function loggamma_inc_moment(D₁, D₂, μ, logλ, k = 0, scale = 1)
-    FT = eltype(logλ)
+    FT = float(UT.promote_typeof(D₁, D₂, μ, logλ))
```
trontrytel: Why float is needed here?

haakon-e: `float` usually does nothing. In the narrow case that the input type is an integer type, `float` returns the corresponding floating point type, e.g.:
```julia
julia> float(Int)
Float64
```
In practice, e.g. `logλ` will always be floating point (since it's a function of the prognostic state that is represented as a floating point type), so I will remove this.

src/IceNucleation.jl
```diff
@@ -266,7 +266,7 @@ Compute the rate of liquid water freezing into ice.
     + `∂ₜq_frz`: Specific mass freezing rate [kg(water) kg⁻¹(air) s⁻¹].
 """
 function liquid_freezing_rate(opt::CMP.RainFreezing, pdf, tps, q, ρ, N, T)
-    FT = eltype(q)
+    FT = float(UT.promote_typeof(q, ρ, N, T))
```
trontrytel, line 269: How expensive it is per element in the list here? Does it save us any to just have q as before?

src/Microphysics2M.jl
```diff
@@ -880,7 +893,8 @@ The method is based on Horn (2012, DOI: [10.5194/gmd-5-345-2012](https://doi.org
 """
 function number_decrease_for_mass_limit((; τ)::CMP.NumberAdjustmentHorn2012, x_min, q, ρ, N)
     # Avoid NaN when both q and x_min are 0
-    N_max = iszero(x_min) ? oftype(q, Inf) : ρ * q / x_min
+    FT = UT.promote_typeof(q, ρ, N)
+    N_max = iszero(x_min) ? FT(Inf) : FT(ρ * q / x_min)
```
trontrytel, line 897: Are we supposed to use ifelse now?

src/Utilities.jl
```diff
@@ -243,6 +258,13 @@ Compatible with dual numbers (AD) and GPUs.
 """
 @inline clamp_to_nonneg(x) = max(zero(x), x)
 
+"""
+    fac(n)
+
+Integer factorial `n!`, valid for `0 ≤ n ≤ 20`.
```
trontrytel, line 264: Should we throw an error for n>20 ?

src/Microphysics2M.jl
haakon-e: I suspect the compiler will internally rewrite these to `ifelse`, but can do it explicitly here for extra safety:)

src/IceNucleation.jl
haakon-e: I'm not quite sure what you mean by expensive? I think the "proper" fix instead of this is to merge #746, as we discussed in the office, where `FT` is derived from the "return type" of the relevant calculation. 

Though that PR is currently implemented as `cond ? f() : otherwise(FT)`, so if we're supposed to rewrite everything with `ifelse` as you mentioned elsewhere, I'd have to update the PR first.

src/Utilities.jl
haakon-e: sure, I can do that. In practice, we'll never use this for numbers even close to that anyways.

src/P3_processes.jl
```diff
@@ -208,8 +208,11 @@ function compute_max_freeze_rate(aps, tps, velocity_params, ρₐ, Tₐ, state)
     # `f_frz = 1`.
     denom = L_f - cp_l * ΔT
     function max_freeze_rate(Dᵢ)
-        Tₐ ≥ T_frz && return zero(Dᵢ)  # No collisional freezing above the freezing temperature
-        denom > 0 || return floatmax(typeof(Dᵢ))
+        # fallback values typed by the promotion of the node and the captured state
+        # (mixed plain/Dual under differentiation)
+        FT = UT.promote_typeof(Dᵢ, ΔT, Δρᵥ_sat, denom)
+        Tₐ ≥ T_frz && return zero(FT)  # No collisional freezing above the freezing temperature
```
trontrytel, line 214: Are we supposed to do ifelse now?

haakon-e: thanks for the catch! I'll update these.

trontrytel (approved):

haakon-e:
welp. Sorry @trontrytel I'll put your feedback into the next PR. I forgot I had auto merge on!

src/IceNucleation.jl
trontrytel: I meant `float(UT.promote_typeof(q, ρ, N, T))` vs `float(UT.promote_typeof(q)` Since I expect it's `q` is the one with the potentially tricky type.

And yeah, I wasn't sure how orthogonal our discussion is compared to the GPU performance PR

src/Utilities.jl
trontrytel: So we think

trontrytel:
> welp. Sorry @trontrytel I'll put your feedback into the next PR. I forgot I had auto merge on!

No worries! I knowingly hit approve

PR #727, opened 2026-06-12
Type 2M sentinel returns by all result-feeding inputs

haakon-e:
Functions that type early returns or sentinel values from a single argument (`FT = eltype(q_tot)`-style) return union-typed, heap-boxed results when a caller mixes plain floats with ForwardDiff Duals across arguments -- as any pointwise Jacobian of the tendencies does (frozen thermodynamic context, Dual species). JET reports nothing and the optimized IR contains no dynamic calls, since the box lives in the abstract return convention (found while chasing per-substep allocations in the Rosenbrock substepping work; tracked there as a value-dependent 32 B/substep).

Adds `Utilities.promote_typeof` and types `rain_evaporation`'s early return by the promotion of all the arguments its rates derive from, with an `@inferred` regression test over both branches. Stacked on #725; further 2M/P3 cases from the same sweep land here as they are confirmed.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -388,6 +388,7 @@ Lightweight numerical utility functions shared across modules.
 ```@docs
 Utilities
 Utilities.clamp_to_nonneg
+Utilities.promote_typeof
 Utilities.ϵ_numerics
 Utilities.ϵ_numerics_2M_M
 Utilities.ϵ_numerics_2M_N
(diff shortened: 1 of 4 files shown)


haakon-e:

- #725
    - #727 ◀
        - #726

haakon-e:
Folded into #725 (tiny fix + tests; not worth a separate PR in the stack).

PR #728, opened 2026-06-14
Make the quadrature scheme the configuration choice

haakon-e:
# Make the quadrature scheme the configuration choice

## Motivation

The P3 size-distribution integrals (deposition and sublimation, melting, riming, ice-rain collection, sedimentation) were configured through a single integer quadrature_order knob. build_quadrature then mapped a magic set of orders, {16, 32, 40, 64}, to Gauss-Legendre and silently fell back to Chebyshev-Gauss for every other order. That coupling makes the knob misleading: the order selects the scheme as a side effect, so a caller who lowers the order from 40 to 36 to trade a little accuracy for speed silently switches integration rules and changes the error characteristics in a way the knob name does not suggest. The two attributes being conflated, the scheme and its order, are independent choices.

## What changed

The quadrature scheme is now the explicit configuration value, passed as a scheme instance that carries its own parameters, the order among them:

    Microphysics2MParams(FT; with_ice = true,
        quadrature = Quadrature.GaussLegendre(40))

build_quadrature is reduced to materializing the chosen scheme in the working float type. Chebyshev-Gauss has closed-form nodes and passes through unchanged; Gauss-Legendre is rebuilt so its stored nodes and weights adopt FT, since a Float64 rule would leak Float64 into Float32 collision integrals. This is host-side and one-shot, and the result is isbits and reused in the GPU hot loop with no per-call construction, as before.

The default is Chebyshev-Gauss at order 100, which reproduces the previous default behavior exactly (the old default order of 100 was not in the Gauss-Legendre magic set, so it already resolved to Chebyshev-Gauss). The documentation now states the scheme tradeoff directly: at matched order Gauss-Legendre is substantially more accurate on the smooth P3 integrands, roughly a factor of 20 lower error on the dominant ice-rain collision integral, while Chebyshev-Gauss remains the conservative default; the cusp-limited ice self-collection diagonal is quadrature-limited under either scheme.

## Breaking change

The quadrature_order keyword argument to Microphysics2MParams and P3IceParams is removed and replaced by the quadrature keyword, which takes a scheme instance. Callers passing quadrature_order must switch to quadrature = Quadrature.ChebyshevGauss(n) or Quadrature.GaussLegendre(n). Callers relying on the implicit order-to-scheme mapping (e.g. expecting order 40 to mean Gauss-Legendre) must now name the scheme explicitly. The default behavior is unchanged.

## Testing

The bulk-tendencies quadrature tests are updated to construct schemes explicitly and to cover both the pass-through (Chebyshev-Gauss) and the FT-rebuild (Gauss-Legendre) paths of build_quadrature.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -116,6 +116,7 @@ Microphysics2M.number_decrease_for_mass_limit
 DistributionTools
 DistributionTools.generalized_gamma_quantile
 DistributionTools.generalized_gamma_cdf
+DistributionTools.gamma_inc_inv
 DistributionTools.exponential_cdf
 DistributionTools.exponential_quantile
 ```
@@ -387,6 +388,7 @@ Lightweight numerical utility functions shared across modules.
 ```@docs
 Utilities
 Utilities.clamp_to_nonneg
+Utilities.promote_typeof
 Utilities.ϵ_numerics
 Utilities.ϵ_numerics_2M_M
 Utilities.ϵ_numerics_2M_N
--- a/docs/src/P3Scheme.md
+++ b/docs/src/P3Scheme.md
@@ -118,6 +118,54 @@ We obtain the following expression for $ρ_d$
   \frac{(β_{va} - 2)(k - 1)}{(1 - F_{rim}) k - 1} - (1 - F_{rim})
 }
 ```
+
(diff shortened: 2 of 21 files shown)


haakon-e:

- #739
    - #725
        - #728 ◀
            - #729

PR #729, opened 2026-06-14
Re-add diagnostic-Nc activation for the 2M scheme

haakon-e:
# Re-add diagnostic-Nc activation for the 2M scheme

## Motivation

The aerosol-activation closure dispatch was dropped from the 2-moment scheme before the P3 merge, leaving the fused 2M tendency with a hardwired zero cloud-droplet number source. The host model could supply its own activation, but there was no in-scheme way to source the prognostic droplet number, which the diagnostic-Nc closure that RCEMIP-I and pre-ARG GCM configurations rely on requires. This PR restores activation, trimmed to the consumer-ready pieces.

## What changed

A small activation-scheme type hierarchy is added: an abstract AbstractActivationScheme with two concrete schemes, NoActivation (the default null source) and DiagnosticNc. Each is a pure immutable parameter struct that fully specifies one closure, and the fused 2M tendency dispatches on the scheme type through a new activation_source helper whose result is folded into the cloud-droplet number tendency dn_lcl_dt.

DiagnosticNc relaxes the prognostic droplet number toward a prescribed target on a fixed physical timescale while cloud mass is present: while cloud liquid mass exceeds a threshold the target is the prescribed number and the tendency is (target - n) / tau, otherwise the target is zero (cloud has evaporated) and the tendency drives n back toward zero on the same timescale. Two design points are worth stating. The relaxation timescale is a fixed physical time, not the integrator timestep: relaxing by 1/dt would produce an infinitely stiff source as an adaptive solver refines the step, which is the same 1/dt-clip pathology avoided elsewhere in this stack. The cloud-existence threshold on liquid mass avoids activating into essentially vapor-only cells, where numerical noise in the liquid mass would otherwise create phantom droplets.

The scheme is wired through WarmRainParams2M as an activation_scheme field defaulting to NoActivation, so existing constructions are unaffected, and surfaced as an activation_scheme keyword on the TOML constructors. The vertical-velocity and pressure ambient inputs are kept as positional arguments to activation_source even though DiagnosticNc ignores them, so the call site can broadcast and the supersaturation-driven closures to come can read them without an interface change.

This mirrors the options pattern of the 1-moment scheme's PrescribedNd, which shapes autoconversion from a prescribed droplet number rather than sourcing a prognostic one: a parameter struct selected by the caller, dispatched on inside the tendency.

## Scope and deferred work

This restores only the diagnostic-Nc tier. The supersaturation-driven activation closures (Twomey, Abdul-Razzak-Ghan) and a prognostic activation-memory model are deferred to a broader activation API; the positional ambient inputs are the seam left for them.

## Testing

The activation_schemes tests cover NoActivation returning an exact zero source, DiagnosticNc relaxation in both the cloud-present and evaporated branches and across the threshold and timescale, activation folding correctly into the fused 2M tendency (the NoActivation default leaves the source at zero, and with DiagnosticNc enabled the droplet-number tendency increases by exactly the activation source), and the activation source typed concretely under mixed argument types.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -67,7 +67,7 @@ Microphysics2M
 
 ```@autodocs
 Modules = [Parameters]
-Pages   = ["Microphysics2M.jl", "Microphysics2MParams.jl"]
+Pages   = ["Microphysics2M.jl", "Microphysics2MParams.jl", "ActivationSchemes.jl"]
 ```
 
 ## Size distributions
@@ -136,6 +136,7 @@ BulkMicrophysicsTendencies.Instantaneous
 BulkMicrophysicsTendencies.InstantaneousVerbose
 BulkMicrophysicsTendencies.LinearizedAverage
 BulkMicrophysicsTendencies.bulk_microphysics_tendencies
+BulkMicrophysicsTendencies.activation_source
 ```
 
 # P3 scheme
--- a/src/BulkMicrophysicsTendencies.jl
+++ b/src/BulkMicrophysicsTendencies.jl
@@ -664,6 +664,38 @@ end
 # --- 2-Moment Microphysics Helper Functions ---
 
 """
+    activation_source(scheme, tps, ρ, T, q_tot, q_lcl, q_ice, n_lcl, w, p)
+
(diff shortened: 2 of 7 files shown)


haakon-e:

- #739
    - #725
        - #728
            - #729 ◀

PR #730, opened 2026-06-14
Generalize the Rosenbrock substep kernel over state dimension

haakon-e:
## What this does

This generalizes the linearized-implicit (Rosenbrock-Euler) substep kernel so it operates on a prognostic state of any static length, in preparation for reusing it across microphysics schemes. The kernel previously hard-coded the eight-component 2-moment + P3 state; it now dispatches on a `StaticArrays.StaticVector{N}`, sizes its identity and dense diagonal matrices from the static length `N`, and takes the channel mask as an argument instead of computing it internally. Mask construction stays scheme-specific, so the kernel itself carries no scheme knowledge and is pure linear algebra over the state vector.

## Why

With this in place a 1-moment Rosenbrock mode, and other schemes, can plug into the same solve by supplying a state FieldVector and a matching channel mask, rather than duplicating the equilibration, channel projection, and update logic. This is a behavior-preserving refactor on its own; the 2-moment + P3 RosenbrockAverage path is unchanged.

## Testing

2-moment + P3 results are bit-for-bit identical to before across the regime, substep-count, and precision matrix (raw IEEE comparison). The existing RosenbrockAverage tests pass unchanged in Float64 and Float32, the hot call stays allocation-free, and the entry is JET-clean.

## Notes

Draft, low in the stack. It adds no new public behavior by itself; the 1-moment mode that consumes this kernel is in the PR above.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -67,7 +67,7 @@ Microphysics2M
 
 ```@autodocs
 Modules = [Parameters]
-Pages   = ["Microphysics2M.jl", "Microphysics2MParams.jl"]
+Pages   = ["Microphysics2M.jl", "Microphysics2MParams.jl", "ActivationSchemes.jl"]
 ```
 
 ## Size distributions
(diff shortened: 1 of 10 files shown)


haakon-e:

- #725
    - #726
        - #728
            - #729
                - #730 ◀
                    - #731
                        - #732
                            - #733

haakon-e:
Folded into #726 — the kernel generalization is small and part of the RosenbrockAverage infrastructure, so it does not warrant a separate stack entry. The change is preserved as a commit in #726.

PR #731, opened 2026-06-14
feat(Rosenbrock): 1M donor / coupled / exact modes on the unified framework

haakon-e:
## What this adds

Adds the 1-moment (`Microphysics1Moment`) path to the `RosenbrockAverage` framework from the previous PR. It introduces three presets:
- `rosenbrock_donor()`: `DonorJacobian`, the donor-based linearization `M`, reproducing `LinearizedAverage` (they agree to round-off in Float64).
- `rosenbrock_coupled()`: `CoupledDonorJacobian`, the donor matrix plus the vapor-competition (WBF) coupling, and corrects the sign of the snow source from cloud liquid.
- `rosenbrock_exact()`: the exact ForwardDiff Jacobian with the growth term taken explicitly and a saturation adjustment. Several 1M kernels were modified to accept dual numbers so ForwardDiff can differentiate the tendency.

## Donor versus exact

The exact Jacobian costs about 2.5× the donor matrix per call. The donor matrix omits cross-couplings that in mixed-phase conditions exceed the diagonal, so at a given number of substeps the exact Jacobian is more accurate there; `coupled` restores the dominant vapor-competition coupling at the donor option's cost.

## Testing

1M tests: `rosenbrock_donor() ≡ LinearizedAverage()`, the keyword constructor, all three presets return finite tendencies, and convergence against a fine explicit reference; zero-allocation and JET checks in Float64 and Float32.

PR #732, opened 2026-06-14
feat(Rosenbrock): post-solve per-process verbose diagnostics

haakon-e:
## What this adds

Adds post-solve per-process tendency diagnostics for the `RosenbrockAverage` modes (1M and 2M+P3) as a `Verbose{M}` wrapper around any mode, reporting the contribution of each microphysical process as realized by the implicit solve.

## Why the decomposition is exact

Each substep solves a system that is linear in the right-hand side `f`:

```
(I/h − J) Δ = f,   f = Σₚ fₚ          (fₚ = the p-th process tendency)
⇒  Δ = Σₚ Δₚ   with   (I/h − J) Δₚ = fₚ
```

so the per-process increments `Δₚ` sum to the substep increment `Δ`. With `nsub > 1` each substep is re-evaluated at the state it starts from, and the per-process increments and the clamp term are accumulated across substeps, so the relation holds for the averaged tendency. The positivity clamp `max(x + Δ, 0)` is the only nonlinear step; it is returned as a single correction term `c` that is not attributable to any one process:

```
Σₚ Δₚ + c = realized net increment.
```

The clamp is a no-op for the donor option (its rates are floored) and binds only rarely for the exact option, since the saturation adjustment already bounds growth. With the clamp disabled the decomposition is exactly `Σₚ Δₚ = net`, which is a reason it could be made optional.

## Testing

Exactness tests confirm the per-process increments plus the clamp correction reconstruct the net to round-off (≈ 1e-13 in Float64) and that the verbose net equals the non-verbose net, for both schemes and precisions. The non-verbose path is unchanged and allocation-free. The verbose path currently allocates and is intended for offline diagnostics; an allocation-free version (a preallocated per-process accumulator, as in the earlier `LinearizedAverage` verbose variant) is a possible refinement.


haakon-e:

- #741
    - #726
        - #731
            - #732 ◀
                - #737
                    - #742
                        - #743
                            - #744

PR #733, opened 2026-06-14
Add implicit-temperature Rosenbrock averaged-tendency mode

haakon-e:
## What this adds

This adds a Rosenbrock averaged-tendency mode that treats temperature implicitly, for both the 1-moment and 2-moment + P3 configurations. The species-only mode advances the local temperature explicitly between substeps from the latent heat of the realized increment, so the dependence of the tendency on temperature sits outside the species Jacobian and the vapor-to-latent-heat-to-temperature-to-saturation feedback is operator-split, which rings about the saturation limit at coarse substeps. Promoting temperature into the implicitly-solved state removes that splitting.

## How

The state is augmented with temperature as a final component (a nine-component 2-moment + P3 state and a five-component 1-moment state), and a joint species-and-temperature tendency appends the latent-heating rate, which is the rate form of the existing between-substep temperature update, so an explicit integration of the augmented tendency reproduces the explicit-temperature trajectory. With temperature differentiated, the Jacobian gains the latent-heating row and the saturation-dependence (Clausius-Clapeyron) column, so the thermal feedback is solved inside the same L-stable one-stage step as the species exchange and no separate between-substep temperature update is done. The dimension-generic kernel is reused unchanged at the new state sizes. This also relaxes the temperature argument of the deposition, sublimation, and evaporation rate helpers (G_func_liquid and G_func_ice) so a dual temperature flows through them, with the air-properties element type still keying the guards, so concrete-temperature callers are unchanged. The existing modes are untouched.

## Validation

Reproducing the operator-split ringing diagnostic, the implicit-temperature mode removes the sign-flip oscillation where the explicit-temperature mode rings, and it converges monotonically against a fine augmented-explicit reference for both schemes and both precisions. In an extreme large-substep cold-deposition case the explicit-temperature temperature diverges by hundreds of kelvin while the implicit-temperature temperature stays bounded. One honest limit: a single very large step still shows single-step linearization error, which is one-stage method truncation error rather than operator splitting, and is cured by more substeps.

## Testing

New tests cover trajectory reproduction, convergence, ringing removal, degeneracy and finiteness, zero allocation, and JET, for both schemes. The existing test set passes unchanged, including the paths that exercise the relaxed temperature signatures.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -137,6 +137,7 @@ BulkMicrophysicsTendencies.InstantaneousVerbose
 BulkMicrophysicsTendencies.LinearizedAverage
 BulkMicrophysicsTendencies.RosenbrockAverage
 BulkMicrophysicsTendencies.RosenbrockAverageVerbose
+BulkMicrophysicsTendencies.RosenbrockAverageImplicitT
 BulkMicrophysicsTendencies.bulk_microphysics_tendencies
 ```
 
--- a/src/BMT_rosenbrock.jl
+++ b/src/BMT_rosenbrock.jl
@@ -927,3 +927,336 @@ A diagnostic path: it may allocate; the non-verbose hot loop is untouched.
     clamp_correction = Δx_clamp_sum / Δt
     return merge(net, (; processes, clamp_correction))
 end
+
+#####
+##### Implicit-temperature Rosenbrock-Euler substepping (`RosenbrockAverageImplicitT`)
+#####
+
+"""
+    MicroState2MP3T{FT}
+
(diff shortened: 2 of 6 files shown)


haakon-e:

- #725
    - #726
        - #731
            - #732
                - #733 ◀
                    - #728
                        - #729

haakon-e:
Superseded by #737. The implicit-temperature Rosenbrock mode is folded into `he/rosenbrock-modes` (#737), which collects the experimental Rosenbrock modes into a single branch. Closing here; the work is preserved in #737.

PR #737, opened 2026-06-17
fix(Rosenbrock): saturation adjustment on max(S_ice, S_liq)

haakon-e:
## What this changes

Refines the `EndStateSaturationAdjustment` limiter so the substep increment is held at saturation over the more-supersaturated phase, `max(S_ice, S_liq)`, instead of over ice alone. Below freezing ice carries the larger supersaturation, so the criterion reduces to the previous ice-only limit and the cold deposition-instability fix is unchanged. Above freezing it limits on liquid, so warm cells condense to liquid saturation; an ice-only limit would instead stop vapor depletion before liquid saturation and hold the cell supersaturated over liquid (too little cloud).

## Docs & test

Adds a "mixed-phase saturation criterion" section to `docs/src/RosenbrockNumerics.md` with a simplified 0-D model that exchanges vapor with cloud liquid and ice only, illustrating the Wegener–Bergeron–Findeisen transfer (why the two saturation thresholds matter), and a framework test covering the warm (liquid) and cold (ice) limits.

## Scope and a known refinement

Limiting on `max(S_ice, S_liq)` is a single shared scalar on the whole increment, which is what keeps the limiter stable. It is correct wherever a single phase grows from vapor. When both phases are supersaturated below freezing (a vigorous mixed-phase updraft core), it limits on the ice floor and so condenses the co-present growing liquid past its own saturation in one step rather than transferring it gradually; a fully per-phase floor (limiting only the supersaturation-driven tendencies of each phase) is a documented candidate refinement. The distinction is inactive at fine substeps and does not affect the cold-instability fix.

### DIFF
--- a/docs/src/RosenbrockNumerics.md
+++ b/docs/src/RosenbrockNumerics.md
@@ -34,8 +34,8 @@ substeps from the latent heat of the realized increment.
 
 - **`TendencyLimiter`** — a limiter applied to the realized substep increment.
   - `NoLimiter`.
-  - `EndStateSaturationAdjustment` — scale the increment so the latent-heated end state does not cross ice
-    saturation (see below).
+  - `EndStateSaturationAdjustment` — scale the increment so the latent-heated end state does not cross saturation
+    over its more-supersaturated phase, `max(S_ice, S_liq)` (see below).
 
 Three preset configurations are supported:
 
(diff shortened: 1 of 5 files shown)


haakon-e:

- #741
    - #726
        - #731
            - #732
                - #737 ◀
                    - #742
                        - #743
                            - #744

PR #740, opened 2026-06-22
fix(P3): use per-regime material density in the ice aspect-ratio closure

haakon-e:
### Summary
The oblate aspect ratio `ϕ = c/a = 3√π·m/(4·ρ·A^{3/2})` was evaluated with the size-dependent effective density ρ = m/V_sphere(D) instead of the particle's material density. With the effective density the particle mass cancels, ϕ comes out prolate (>1) for every nonspherical size, and the `min(1, ϕ)` clamp then pinned `ϕ ≡ 1` everywhere — disabling the `cbrt(ϕ)` aspect-ratio fall-speed correction entirely. Use the per-regime material density and drop the clamp.

### What changed (src/P3_particle_properties.jl, docs, tests)
- New `ϕ_material_density(state, D)`: returns `ρ_i` in the small-spherical, dense-nonspherical, and partially-rimed regimes, and `ρ_g` in the graupel regime (`D_gr ≤ D < D_cr`).
- `ϕᵢ` now divides by `ϕ_material_density(state, D)` and no longer clamps; the docstring keeps the one-line formula and links to the docs.
- `ice_density` docstring: removed the stale "Needed for aspect ratio calculation, so we assume zero liquid fraction" clause (ϕᵢ no longer calls it).

### Mechanism / why
`ϕ = 3√π·m/(4·ρ·A^{3/2})` compares the particle's mass-to-area shape factor against the density of the *solid material* it is made of. Substituting the effective density `m/V_sphere(D)` replaces `ρ` with a quantity that itself carries the mass, cancelling `m` and collapsing the closure to a near-constant ≈ 1; the `min(1, ⋅)` clamp then masked the residual `>1` excursions, so ϕ ≡ 1 and `cbrt(ϕ)` ≡ 1. Within each P3 mass regime the material density is constant, so ϕ tracks m/A^{3/2}: spherical in the spherical regimes (ϕ = 1), oblate (ϕ < 1) in the dense-nonspherical regime.

### Residual ϕ > 1 band just above D_th (why no clamp)
ϕ slightly exceeds 1 (peak ≈ 1.2) in a narrow band immediately above D_th. The projected *area* is discontinuous at D_th (it drops from the spherical law (π/4)D² to the nonspherical law γ·D^σ, ≈ 0.88× smaller) while the *mass* stays continuous, so m/A^{3/2} is inflated just past the threshold. `cbrt(ϕ)` stays finite and bounded there (cbrt(1.2) ≈ 1.06), so a clamp is unnecessary and would re-introduce the ϕ ≡ 1 pathology in that band. Reconciling the mass and area laws at D_th (so the area is continuous too) is a separate P3 area power-law item, deliberately left as a follow-on.

### Validation
- Effect: large unrimed ice fall speeds drop by up to ~4× — the intended oblate correction. Mass-weighted velocities drop most (mass weighting emphasises the large nonspherical sizes where ϕ is smallest).

### Caveats / follow-on
- The residual ϕ > 1 band above D_th remains (follow-on: reconcile the area law).


haakon-e:

- #725
    - #726
        - #731
            - #732
                - #737
                    - #740 ◀
                        - #741
                            - #742
                                - #743
                                    - #744

haakon-e:
Folded into #741

PR #742, opened 2026-06-23
perf(BMT): recover the primal tendency from the Rosenbrock Jacobian pass

haakon-e:
### Summary
In the Rosenbrock micro substep the primal tendency `f = g(x)` and the Jacobian `ForwardDiff.jacobian(g, x)` were computed separately, but ForwardDiff's Jacobian pass already evaluates `f` as the dual `.value` and discards it. Recover it with `DiffResults`/`jacobian!`, eliminating one full (quadrature-dominated) tendency evaluation per substep.

### What changed (src/BMT_rosenbrock.jl, Project.toml, BulkMicrophysicsTendencies.jl)
- Both the 2M+P3 and the 1M ExactJacobian substep paths now go through a shared `_tendency_and_jacobian(jacobian, g, x)` method: `r = ForwardDiff.jacobian!(DiffResults.JacobianResult(x), g, x)`, then `f = typeof(x)(DiffResults.value(r))`, `J = DiffResults.jacobian(r)`. The method reconstructs the same `FieldVector` species type as `x` (`MicroState1M` or `MicroState2MP3`), removing the duplicated DiffResults code that previously lived inline in the 2M+P3 loop.
- The analytic Donor / CoupledDonor 1M Jacobians have no `f` by-product, so their `_tendency_and_jacobian` methods keep the separate `f = g(x)`. The non-finite euler fallback also keeps `f = g(x)`.
- Add `DiffResults` to [deps] + [compat]; `import DiffResults` in BulkMicrophysicsTendencies.jl.

### Mechanism / why
`ForwardDiff.jacobian(g, x::StaticArray)` runs one dual evaluation and extracts the Jacobian, throwing away the dual `.value` (= `f`). `jacobian!` into an immutable `JacobianResult` does the identical single pass and additionally records the value, so `f` is recovered for free. `JacobianResult` is immutable for a StaticArray, so the `jacobian!` return value must be captured (it is not mutated in place). Two opaque call sites otherwise prevent common-subexpression elimination, so the saving is real.

### Validation
- `f` and J bit-identical to the separate computation (verified `===` on the reconstructed `MicroState1M` and `MicroState2MP3` and on J); no extra allocations (`@allocated == 0` for both helper paths and the full 2M+P3 `bulk_microphysics_tendencies`); type-stable (`@inferred`).
- ~27% faster per substep (order-independent since J ≈ 2.68× f).
- rosenbrock_1m / mode / framework / verbose tests pass (F32 + F64), including the allocation and "vs fine explicit reference" checks.

### Caveats
- ExactJacobian paths only (by design). `DiffResults` becomes a direct dep (it was transitive via ForwardDiff).

### DIFF
--- a/src/BMT_rosenbrock.jl
+++ b/src/BMT_rosenbrock.jl
@@ -424,10 +424,10 @@ fields as the `Instantaneous` entry (without the activation diagnostic).
     Tsub = T
     for _ in 1:nsub_eff
         g = Instantaneous2MP3Tendency(mp, tps, ρ, Tsub, q_tot, logλ)
-        f = g(x)
         x_prev = x
         if all(isfinite, x)
-            J = _apply_growth(mode.growth, FD.jacobian(g, x))
+            f, J_raw = _tendency_and_jacobian(mode.jacobian, g, x)
+            J = _apply_growth(mode.growth, J_raw)
             z = _species_mask(mode.jacobian, mode.growth)(x)
             d = if all(isfinite, J)
                 _rosenbrock_update(x, f, J, z, h) - x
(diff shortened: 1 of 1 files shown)


haakon-e:

- #741
    - #726
        - #731
            - #732
                - #737
                    - #742 ◀
                        - #743
                            - #744

PR #746, opened 2026-06-24
feat(Utilities): add safe_call for type-inferred conditional fallbacks

haakon-e:
## Summary

Add `Utilities.safe_call`, a higher-order helper for conditional values whose fallback type is inferred from the computation rather than from its inputs:

```julia
safe_call(q_ice > ϵ; otherwise = zero) do
    n_ice / q_ice
end
```

It returns the `do` block when `cond` is `true` and `otherwise(FT)` when it is `false`, where `FT = Core.Compiler.return_type(f, Tuple{})`.

## Motivation

The recurring `zero(promote_typeof(args...))` idiom has a weakness that `safe_call` removes:

- `promote_typeof` types the fallback from the *inputs*, and every derivative-carrying argument must be listed by hand; omitting one returns a `Union`, which is non-concrete when using `ForwardDiff` on the GPU. `safe_call` takes the type from the function's *range*, so it cannot omit an argument.

## Notes for review

- `Core.Compiler.return_type` is internal and inference-directed: the fallback type is correct only where the computation is type-stable; this is required for the code to run on GPUs anyways. Where inference widens (Union type), this will fail.
- No call site is converted here, though I have tested it on `_regularised_ratio` in the same module. With that, it is type-stable and allocation-free through `ForwardDiff` on CUDA, for the default and overridden `otherwise`.

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -388,6 +388,7 @@ Lightweight numerical utility functions shared across modules.
 Utilities
 Utilities.clamp_to_nonneg
 Utilities.promote_typeof
+Utilities.safe_call
 Utilities.ϵ_numerics
 Utilities.ϵ_numerics_2M_M
 Utilities.ϵ_numerics_2M_N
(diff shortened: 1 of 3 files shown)


haakon-e:

- #725
    - #746 ◀

haakon-e:
Closing in favor of branch-free patterns: rate functions are now evaluated unconditionally, with ifelse selecting the result, rather than short-circuiting on a condition (see e.g. #757). This removes the use case for safe_call.

PR #748, opened 2026-06-26
GPU performance: eliminate warp divergence and reduce compile times

tapios:
## Summary

Optimizes CloudMicrophysics for GPU execution by removing sources of warp divergence, replacing expensive `SpecialFunctions.jl` calls with simpler alternatives, and pre-computing constant gamma values at parameter-construction time.

## Key changes

- **Custom `gamma_inc`/`gamma_inc_inv`** in `Utilities.jl`: fixed-iteration, low-divergence replacements for `SpecialFunctions.gamma_inc`; cuts compile time and uses `ifelse` in inner loops to avoid warp divergence.
- **P3 shape solver**: Uses `BrentsMethod` with `FixedIterations` tolerance (8/10 iters for F32/F64); constant iteration count eliminates warp divergence. In GPU experiments, this was the fastest variant (faster than Regula Falsi, for example; however, CPU performance differs). It will be further accelerated in the next release of RootSolvers (coming soon).
- **Replace convergence tolerances by fixed iterations* in other places too. 
- **Closures → callable structs** across P3 size distribution and terminal velocity for type stability and GPU compatibility.
- **Pre-computed gamma constants** in parameter structs (`Blk1MVelTypeRain`, `CloudParticlePDF_SB2006`, `EvaporationSB2006`): moves `SF.gamma`/`SF.loggamma` from hot loops to construction time.
- **Numerical robustness**: `P − P` vs `Q − Q` formulation in incomplete-gamma moment integrals to avoid catastrophic cancellation.
- New `test/gpu_performance.jl` benchmark suite; updated allocation thresholds.
- Achieves >2x speedup of P3 GPU kernel

### DIFF
--- a/src/P3_terminal_velocity.jl
+++ b/src/P3_terminal_velocity.jl
@@ -1,4 +1,18 @@
 
+# Callable returned by `ice_particle_terminal_velocity`: piecewise small/large-ice
+# Chen 2022 velocity with optional aspect-ratio correction.
+struct P3IceParticleVelocityFunctor{FT, VS, VL, S} <: Function
+    v_term_small::VS
+    v_term_large::VL
+    D_cutoff::FT
+    state::S
+    use_aspect_ratio::Bool
+end
+@inline function (f::P3IceParticleVelocityFunctor)(D)
+    vₜ = D <= f.D_cutoff ? f.v_term_small(D) : f.v_term_large(D)
+    return f.use_aspect_ratio ? cbrt(ϕᵢ(f.state, D)) * vₜ : vₜ
+end
+
 """
     ice_particle_terminal_velocity(velocity_params, ρₐ, state::P3State; [use_aspect_ratio])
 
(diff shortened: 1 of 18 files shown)


src/P3_terminal_velocity.jl
```diff
@@ -22,12 +34,15 @@ terminal velocity of an ice particle of maximum dimension `D`.
...
-        end
-    return v_term
+    return P3IceParticleVelocityFunctor(v_term_small, v_term_large, D_cutoff, state, use_aspect_ratio)
+end
+
+struct P3NumberWeightedIntegrand{N, V} <: Function
+    n::N
+    v_term::V
 end
+@inline (f::P3NumberWeightedIntegrand)(D) = f.n(D) * f.v_term(D)
```
haakon-e: this code should already be type stable and GPU compatible. Did you find otherwise? i.e., the rewrite from closures to callable structs

test/performance_tests.jl
```diff
@@ -178,7 +178,7 @@ function benchmark_test(FT)
         (params_P3, L_ice, N_ice, F_rim, ρ_rim),
         220,
     )
-    bench_press(FT, P3.get_distribution_logλ, (state,), 30_000)
+    bench_press(FT, P3.get_distribution_logλ, (state,), 15_000)  # 12 (F64) / 10 (F32) FixedIterations RegulaFalsi, zero warp divergence
```
haakon-e: nice!

src/P3_terminal_velocity.jl
tapios: For performance on local benchmarks, it makes no differnece. However, my understanding is that without the structs, there's a risk that compiler inference fails (e.g., exceeding inference depth budgets) when this code is embedded in the deeply nested call trees of atmos code. Explicit callable structs prevent accidental heap allocations in those situations. 

Happy to remove it though if there's no need for it. (@dennisYatunin).

haakon-e: hmm. I'd be curious to see if this compiles on 1.11 with these changes then, since `exceeding inference depth budgets` currently limits me to running on 1.12

haakon-e (approved):

PR #749, opened 2026-06-26
Define numerics epsilon for 1M microphysics

petebachant:
This should claw back the big drop in performance on June 16th, which appears to be from branch divergence.

image

### DIFF
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -81,7 +81,7 @@ Returns the intercept parameter of the assumed Marshall-Palmer distribution
 - `ρ`: air density (snow only)
 """
 @inline get_n0((; ν, μ)::CMP.ParticlePDFSnow{FT}, q_sno::FT, ρ::FT) where {FT} =
-    q_sno > UT.ϵ_numerics_1M_M(FT) ? μ * (ρ * q_sno)^ν : zero(FT)
+    μ * max(ρ * q_sno, zero(FT))^ν
 @inline get_n0((; n0)::CMP.ParticlePDFIceRain{FT}, args...) where {FT} = n0
 
 """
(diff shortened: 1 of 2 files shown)


src/Microphysics1M.jl
```diff
@@ -134,7 +134,9 @@ average particles. The value is clipped at `r0 * 1e-5` to prevent numerical issu
     (; r0, m0, me, Δm, χm, gamma_coeff) = mass
 
     λ_inv = FT(0)
-    if q > UT.ϵ_numerics(FT) && ρ > UT.ϵ_numerics(FT)
+    if q > UT.ϵ_numerics_1M_M(FT) && ρ > UT.ϵ_numerics(FT)
```
tapios: Why don't we make this branchless altogether? We can use `max(ρ, 0)` and `max(q, 0)` for safe  evaluation of `λ_inv` (though I think at least `q` is already guarded upstream).

tapios:
As a rule, we should avoid branches. With 1e6 grid points, saving some flops on a subset of the points generally does not bring any benefit because there will always be points that need the full evaluation. It seems to me we can eliminate all `if q > ϵ_numerics_1M_M` conditionals (`lambda_inverse` has a safe cutoff, and rates/velocities go to zero with `q`). [There are some physical branches, e.g., for `T > T_freeze` and `S > 0`, and we need these logically. But they can also be made into branchless `ifelse` constructions.]

trontrytel:
> As a rule, we should avoid branches. With 1e6 grid points, saving some flops on a subset of the points generally does not bring any benefit because there will always be points that need the full evaluation. It seems to me we can eliminate all `if q > ϵ_numerics_1M_M` conditionals (`lambda_inverse` has a safe cutoff, and rates/velocities go to zero with `q`). [There are some physical branches, e.g., for `T > T_freeze` and `S > 0`, and we need these logically. But they can also be made into branchless `ifelse` constructions.]

There were some cases where for very small (but still positive numbers) some parameterizations like Chen 2022 terminal velocity were returning bad output. Rather than tending to zero, the code was returning very large numbers. I think we fixed issues were we found them, like here: https://github.com/CliMA/CloudMicrophysics.jl/pull/482/changes
But be on a look out for similar problems

petebachant:
Eliminating branching here appeared to not provide any speedup compared with the threshold change. Closing this since `main` has progressed a bit and I have a new angle to try.

PR #750, opened 2026-06-29
GPU performance optimizations

tapios:
## Purpose 

Improve GPU performance of 1M, 2M, and P3 microphysics by removing branches and reducing transcendental function calls (e.g., inside quadratures and other hot paths).

## Changes
- **1M / NonEq**: replace `if q > ϵ` guards with branchless `ifelse`; simplify `lambda_inverse`; compute the shared size-distribution quantities (λ⁻¹, n₀, v₀ for rain/snow/ice) once per cell and thread them via an `sd` argument instead of recomputing per process.
- **2M**: branchless SB2006 rates and terminal velocities; compute `cbrt(Sc)` / `sqrt(Re)` once in rain evaporation.
- **P3**: fuse the Chen velocity `D^b·exp(-c·D)` into one `exp`; `ϕᵢ`: `aᵢ^1.5 → aᵢ·√aᵢ`; hoist `loggamma` out of the `gamma_inc_inv` loop; add a μ=1 fast-path quantile; split `ice_self_collection` at the `|Δv|=0` cusp; lower the `quadrature_order` default **100 → 16** (auto-selects Gauss-Legendre).
- **Replace rational exponents** by `FT`.  This is an important lesson for our code stack: Internally, for an expression such as `x::Float32 ^ (-2 // 3)`, Julia constructs a `Rational{Int64}` struct on the GPU thread's local stack. To do so, it runs a 64-bit integer Greatest Common Divisor (`divgcd`) routine along with 64-bit overflow checks on the GPU. Then it lowers numerator and denominator to f32 and computes the power--all in all, with a massive overhead. (**Changing the rational exponents alone achieved ~7x speedup for 2M.**)

## GPU Performance: `main` (27bbdeb1) vs `ts/cmp-gpu-opt` (b0e306f5) [based on array of 1024 CMP evaluations in `test/gpu_performance.jl']
| Microphysics Kernel | Precision | `main` Runtime | `ts/cmp-gpu-opt` Runtime | Speedup vs `main` |
|:---|:---|---:|---:|:---|
| **1-Moment Bulk Tendencies** | Float64 | 261.2 μs (83 allocs) | 57.5 μs (43 allocs) | **4.54×** |
| | Float32 | 51.9 μs (45 allocs) | 31.3 μs (42 allocs) | **1.66×** |
| **2-Moment Bulk Tendencies** | Float64 | 1,947.0 μs (298 allocs) | 43.6 μs (48 allocs) | **44.6×** |
| | Float32 | 1,713.0 μs (298 allocs) | 29.9 μs (41 allocs) | **57.3×** |
| **P3 Ice Microphysics** | Float64 | 127.3 ms (294 allocs) | 101.2 ms (294 allocs) | **1.26×** |
| | Float32 | 45.4 ms (294 allocs) | 46.8 ms (294 allocs) | 0.97× |

* All allocations are CPU-side kernel launch tracking overhead from KernelAbstractions/CUDA.jl
* ~6% speedup of 1M bulk microphysics comes from branch removal (at f32 on GPU); most of the rest from caching special function evaluations. 
* A factor 7 of the 2M speedup comes from replacing rational exponents.

### DIFF
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -135,15 +135,17 @@ average particles. The value is clipped at `r0 * 1e-5` to prevent numerical issu
     # mass(size)
     (; r0, m0, me, Δm, χm, gamma_coeff) = mass
 
-    # Branchless: clamp q, ρ to non-negative and floor the n0 denominator (for snow,
-    # n0 ∝ q^ν → 0 as q → 0) so the ratio stays finite instead of 0/0. As q → 0 the
-    # base → 0, so λ_inv → 0 and is clamped to the r0*1e-5 floor below. This matches the
-    # old `if q > ϵ` guard to within that floor for all physically relevant q, with no
-    # warp divergence (cf. PR #749).
+    # Domain sanitization only (NOT a physical threshold): `clamp_to_nonneg` on q, ρ
+    # and the `max(n0, ϵ)` floor keep the ratio finite instead of 0/0 (for snow,
+    # n0 ∝ q^ν → 0 as q → 0). As q → 0 the base → 0, so λ_inv → 0 and is clamped to the
+    # r0*1e-5 floor below, keeping λ_inv (a denominator in the rate formulas) finite so
+    # the *discarded* branch of each caller's `ifelse(q > ϵ_numerics, rate, 0)` gate
+    # cannot produce Inf/NaN. The physical "tracer absent" decision lives in those
+    # caller gates, not here.
     # Note: Julia compiles x^y to exp(y * log(x)); gamma_coeff is pre-computed in the
     # ParticleMass constructor for GPU performance.
-    qp = max(q, zero(FT))
-    ρp = max(ρ, zero(FT))
+    qp = UT.clamp_to_nonneg(q)
+    ρp = UT.clamp_to_nonneg(ρ)
     denom = χm * m0 * max(n0, UT.ϵ_numerics(FT)) * gamma_coeff
     λ_inv = (ρp * qp * r0^(me + Δm) / denom)^(1 / (me + Δm + 1))
     return max(r0 * FT(1e-5), λ_inv)
@@ -361,14 +363,16 @@ end
     return max(0, q_lcl) / (τ * (Nc / 100_000_000)^α)
 end
 
-# Size-distribution / fall-speed quantities shared across the 1-moment process rates.
+# Size-distribution / fall-speed parameters shared across the 1-moment process rates.
 # `lambda_inverse` (a `pow`), snow `get_n0` (a `pow`) and `get_v0` are each reused by
 # several processes for the same hydrometeor species, so computing them once per cell
 # avoids redundant transcendentals that the compiler cannot eliminate across the
-# (inlined) per-process calls. `BulkMicrophysicsTendencies` builds this once and threads
-# it to the process functions below via their optional `sd` argument; standalone callers
-# omit it and the default rebuilds it, and the compiler then drops the fields they don't use.
-@inline function _size_dist_cache(mp, micro, thermo)
+# (inlined) per-process calls. This is the "compute once and pass" pattern:
+# `BulkMicrophysicsTendencies` builds this once per cell and threads it to the process
+# functions below via their optional `sd` argument. The `sd = size_distr_parameters(...)`
+# default is evaluated only for standalone callers that omit it, so there is no repeated
+# computation on the BMT hot path; for such callers the compiler drops the unused fields.
+@inline function size_distr_parameters(mp, micro, thermo)
     (; q_rai, q_sno, q_icl) = micro
     ρ = thermo.ρ
     return (;
@@ -414,7 +418,7 @@ Harrington et al. (1995) and Kaul et al. (2015).
 end
 
 @inline function conv_q_icl_to_q_sno(
-    opt::CMP.WithSupersaturation, mp, tps, micro, thermo, sd = _size_dist_cache(mp, micro, thermo),
+    opt::CMP.WithSupersaturation, mp, tps, micro, thermo, sd = size_distr_parameters(mp, micro, thermo),
 )
     (; q_tot, q_lcl, q_icl, q_rai, q_sno) = micro
     (; ρ, T) = thermo
@@ -708,7 +712,7 @@ delegate to the corresponding low-level Marshall-Palmer kernels.
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_lcl = micro.q_lcl
     q_rai = micro.q_rai
@@ -733,7 +737,7 @@ end
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_lcl = micro.q_lcl
     q_sno = micro.q_sno
@@ -761,7 +765,7 @@ end
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_icl = micro.q_icl
     q_rai = micro.q_rai
@@ -786,7 +790,7 @@ end
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_icl = micro.q_icl
     q_sno = micro.q_sno
@@ -814,7 +818,7 @@ end
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_rai = micro.q_rai
     q_sno = micro.q_sno
@@ -870,7 +874,7 @@ end
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_icl = micro.q_icl
     q_rai = micro.q_rai
@@ -915,7 +919,7 @@ Only evaporation is considered (sub-saturated over liquid); result is clamped �
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     (; q_tot, q_lcl, q_icl, q_rai, q_sno) = micro
     (; ρ, T) = thermo
@@ -977,7 +981,7 @@ Ventilation factor parameterization follows Seifert and Beheng (2006).
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     return min(0, _snow_subl_dep_rate(mp, tps, micro, thermo, sd))
 end
@@ -988,13 +992,13 @@ end
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     return _snow_subl_dep_rate(mp, tps, micro, thermo, sd)
 end
 
 """Internal helper: snow sublimation/deposition physics kernel."""
-@inline function _snow_subl_dep_rate(mp, tps, micro, thermo, sd = _size_dist_cache(mp, micro, thermo))
+@inline function _snow_subl_dep_rate(mp, tps, micro, thermo, sd = size_distr_parameters(mp, micro, thermo))
     (; q_tot, q_lcl, q_icl, q_rai, q_sno) = micro
     (; ρ, T) = thermo
     (; pdf, mass, vent) = mp.precip.snow
@@ -1053,7 +1057,7 @@ Returns the tendency due to cloud ice melt.
     tps,
     micro,
     thermo,
-    sd = _size_dist_cache(mp, micro, thermo),
+    sd = size_distr_parameters(mp, micro, thermo),
 )
     q_icl = micro.q_icl
     (; ρ, T) = thermo
@@ -1086,7 +1090,7 @@ Returns the tendency due to snow melt.
 """
 @inline conv_q_sno_to_q_rai(::Nothing, mp, tps, micro, thermo, sd = nothing) = zero(thermo.T)
 
-@inline function conv_q_sno_to_q_rai(::CMP.SnowMelt, mp, tps, micro, thermo, sd = _size_dist_cache(mp, micro, thermo))
+@inline function conv_q_sno_to_q_rai(::CMP.SnowMelt, mp, tps, micro, thermo, sd = size_distr_parameters(mp, micro, thermo))
     q_sno = micro.q_sno
     (; ρ, T) = thermo
     (; pdf, mass, vent) = mp.precip.snow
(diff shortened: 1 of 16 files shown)


src/P3_size_distribution.jl
```diff
@@ -1,14 +1,14 @@
...
-struct P3LogNumberFunctor{FT, T} <: Function
+struct P3LogNumberFunctor{FT} <: Function
     log_N₀::FT
     μ::FT
-    logλ::T
+    λ::FT
 end
 @inline function (f::P3LogNumberFunctor)(D)
     logD = log(D)
-    return f.log_N₀ + f.μ * logD - exp(f.logλ + logD)
```
trontrytel, line 11: Why is it better to pass here lambda and do the conversion in the function below?

src/Utilities.jl
```diff
@@ -80,7 +84,7 @@ end
 
     # factor = x^a * e^-x / Gamma(a)
     # Using loggamma for numerical stability
-    factor = exp(a * log(x) - x - SF.loggamma(a))
+    factor = exp(a * log(x) - x - lΓa)
```
trontrytel: Can we keep a move verbose name?

src/P3_particle_properties.jl
```diff
@@ -258,17 +258,18 @@ function ice_mass_coeffs(state::P3State, D)
...
-    elseif iszero(F_rim)     # large nonspherical unrimed ice
-        (α_va, β_va)
-    elseif D_th ≤ D < D_gr   # dense nonspherical rimed ice
-        (α_va, β_va)
-    elseif D_gr ≤ D < D_cr   # graupel (rimed)
-        (ρ_g * π / 6, FT(3))
-    else # D_cr ≤ D          # partially rimed ice
-        (α_va / (1 - F_rim), β_va)
-    end
+    unrimed = iszero(F_rim)
```
trontrytel, line 330: For code readability could we keep the verbal description of different regimes either in the comments or by changing the condition names to be more descriptive?

src/P3_particle_properties.jl
```diff
@@ -333,19 +334,19 @@ Return the cross-sectional area of a particle based on where it falls in the
...
-        nonspherical_area(D)
-    elseif D_gr ≤ D < D_cr   # graupel (rimed)
-        spherical_area(D)
-    else # D_cr ≤ D          # partially rimed ice
-        weighted_average(F_rim, spherical_area(D), nonspherical_area(D))
-    end
+    s_area = D^2 * π / 4
+    ns_area = γ * D^σ
+
+    unrimed = iszero(F_rim)
```
trontrytel, line 414: Same as above, can we keep somewhere the comments about the interpretation of the different regimes here

src/P3_processes.jl
```diff
@@ -437,7 +437,14 @@ Computes the bulk collision rate integrands between ice and liquid particles.
...
 A tuple of 8 integrands, see [`∫liquid_ice_collisions`](@ref) for details.
 """
-function ∫liquid_ice_collisions(n_i, ∂ₜM_max, cloud_integrals, rain_integrals, ice_bounds; quad = ChebyshevGauss(100))
+@inline function ∫liquid_ice_collisions(
+    n_i,
+    ∂ₜM_max,
+    cloud_integrals,
+    rain_integrals,
+    ice_bounds;
+    quad = ChebyshevGauss(100),
```
trontrytel, line 459: We still think we need 100 points here?

src/Microphysics2M.jl
```diff
@@ -67,33 +67,43 @@ Return the parameters of the rain drop diameter distribution
...
+    xr_mean = Lᵣ / safe_Nᵣ
     λr = cbrt(π * ρw / xr_mean)
-    N₀r = λr * Nᵣ
+    N₀r = λr * safe_Nᵣ
 
     Dr_mean = 1 / λr  # The inverse of λr is the mean diameter of the raindrops (units: `m`)
-    return (; N₀r, Dr_mean, xr_mean)
+    cond = Nᵣ < UT.ϵ_numerics_2M_N(FT) || qᵣ < UT.ϵ_numerics_2M_M(FT)
+    return (;
+        N₀r = ifelse(cond, zero(Nᵣ), N₀r),
```
trontrytel: This version evaluates the same condition 3 times. Why does it end up being faster?

src/Microphysics2M.jl
```diff
@@ -67,33 +67,43 @@ Return the parameters of the rain drop diameter distribution
...
-    N₀r = clamp(Nᵣ * cbrt(π * ρw / x̃r), N0_min, N0_max)  # Eq. (95)
+    x̃r = clamp(Lᵣ / safe_Nᵣ, xr_min, xr_max)  # Eq. (94)
+    N₀r = clamp(safe_Nᵣ * cbrt(π * ρw / x̃r), N0_min, N0_max)  # Eq. (95)
     λr = clamp(sqrt(sqrt(π * ρw * N₀r / Lᵣ)), λ_min, λ_max)  # Eq. (96)
     xr_mean = clamp(Lᵣ * λr / N₀r, xr_min, xr_max)  # Eq. (97)
 
     Dr_mean = 1 / λr  # The inverse of λr is the mean diameter of the raindrops (units: `m`)
-    return (; N₀r, Dr_mean, xr_mean)
+    cond = Nᵣ < UT.ϵ_numerics_2M_N(FT) && qᵣ < UT.ϵ_numerics_2M_M(FT)
+    return (;
```
trontrytel, line 106: Same as above

src/Microphysics2M.jl
```diff
@@ -776,40 +769,42 @@ function rain_evaporation(
...
 
-    ∂ₜρn_rai = FT(0)
-    ∂ₜq_rai = FT(0)
     S = TDI.supersaturation_over_liquid(tps, q_tot, q_lcl + q_rai, q_icl + q_sno, ρ, T)
 
-    # If there are no raindrops or conditions are supersaturated, no evaporation occurs
-    (N_rai ≤ ϵₙ || S ≥ 0) && return (; ∂ₜρn_rai, ∂ₜq_rai)
-
     (; ν_air, D_vapor) = aps
-    (; av, bv, α, β, ρ0) = evap
```
trontrytel, line 800: Why is it better to not get the `av` and `bv` params here?

src/MicrophysicsNonEq.jl
```diff
@@ -246,15 +244,13 @@ function terminal_velocity(
...
-        # Stokes law: v(D) = C * D^2, valid for D < ~80 μm (Re < 1)
-        v_term = CO.particle_terminal_velocity(vel, ρₐ)
-        # Mean volume diameter from assumed number concentration
-        D = cbrt(FT(6 / π) * ρₐ * q / N_0 / ρw)
-        fall_w = v_term(D)
-    end
-    return fall_w
+    # Stokes law: v(D) = C * D^2, valid for D < ~80 μm (Re < 1)
+    v_term = CO.particle_terminal_velocity(vel, ρₐ)
+    safe_q = max(q, zero(FT))
```
trontrytel: We used to check if greater than epsilon numerics

trontrytel: Same here, we used to compare against epsilon

src/MicrophysicsNonEq.jl
```diff
@@ -263,14 +259,12 @@ function terminal_velocity(
...
-    fall_w = FT(0)
-    if q > UT.ϵ_numerics(FT)
-        v_term = CO.particle_terminal_velocity(vel, ρₐ, ρᵢ)
-        # Mean volume diameter from assumed number concentration
-        D = cbrt(FT(6 / π) * ρₐ * q / N_0 / ρᵢ)
-        fall_w = max(FT(0), v_term(D))
-    end
-    return fall_w
+    v_term = CO.particle_terminal_velocity(vel, ρₐ, ρᵢ)
+    safe_q = max(q, zero(FT))
```
trontrytel: And here

src/Microphysics1M.jl
```diff
@@ -133,12 +135,17 @@ average particles. The value is clipped at `r0 * 1e-5` to prevent numerical issu
...
-        λ_inv = (ρ * q * r0^(me + Δm) / (χm * m0 * n0 * gamma_coeff))^(1 / (me + Δm + 1))
-    end
+    # Branchless: clamp q, ρ to non-negative and floor the n0 denominator (for snow,
+    # n0 ∝ q^ν → 0 as q → 0) so the ratio stays finite instead of 0/0. As q → 0 the
+    # base → 0, so λ_inv → 0 and is clamped to the r0*1e-5 floor below. This matches the
+    # old `if q > ϵ` guard to within that floor for all physically relevant q, with no
+    # warp divergence (cf. PR #749).
+    # Note: Julia compiles x^y to exp(y * log(x)); gamma_coeff is pre-computed in the
+    # ParticleMass constructor for GPU performance.
+    qp = max(q, zero(FT))
```
trontrytel: Same as comments in other files - this PR in some places keeps the comparison with epsilon and in some other compares with zeros. I think we should keep one value with which we compare our small tracers. For consistency sake, so that we don't end up with some mismatches later in Atmos where some small numbers are already ignored and the other are not.

If there is a reason that we need two epsilon values (or a hard zero and an epsilon) I would still prefer to define them in `Utilities` or somewhere and always call from there. This would give us a way to easily change the value in all places, rather than having Claude do the detective work

src/Microphysics1M.jl
```diff
@@ -380,37 +405,40 @@ Harrington et al. (1995) and Kaul et al. (2015).
...
-@inline function conv_q_icl_to_q_sno(opt::CMP.NoSupersaturation, mp, tps, micro, thermo)
+@inline function conv_q_icl_to_q_sno(opt::CMP.NoSupersaturation, mp, tps, micro, thermo, sd = nothing)
     (; τ, q_threshold, k) = opt.acnv1M
     q_icl = micro.q_icl
     return CO.logistic_function_integral(q_icl, q_threshold, k) / τ
 end
 
-@inline function conv_q_icl_to_q_sno(opt::CMP.WithSupersaturation, mp, tps, micro, thermo)
+@inline function conv_q_icl_to_q_sno(
+    opt::CMP.WithSupersaturation, mp, tps, micro, thermo, sd = _size_dist_cache(mp, micro, thermo),
```
trontrytel: Did Claude say this is more performant than computing `size_distr_cache` somewhere at the beginning of BMT and simply passing the precomputed struct?

src/Microphysics1M.jl
```diff
@@ -358,6 +361,28 @@ end
...
 end
 
+# Size-distribution / fall-speed quantities shared across the 1-moment process rates.
+# `lambda_inverse` (a `pow`), snow `get_n0` (a `pow`) and `get_v0` are each reused by
+# several processes for the same hydrometeor species, so computing them once per cell
+# avoids redundant transcendentals that the compiler cannot eliminate across the
+# (inlined) per-process calls. `BulkMicrophysicsTendencies` builds this once and threads
+# it to the process functions below via their optional `sd` argument; standalone callers
+# omit it and the default rebuilds it, and the compiler then drops the fields they don't use.
+@inline function _size_dist_cache(mp, micro, thermo)
```
trontrytel: Could we change the name to `size_distr_parameters`? Or at least get rid of the underscore?

src/Microphysics1M.jl
```diff
@@ -491,28 +536,41 @@ end
...
-    end
-    return accr_rate
+@inline function accretion_rain_sink(
+    rain::CMP.Rain,
+    ice::CMP.CloudIce,
+    vel::CMP.Blk1MVelTypeRain{FT},
+    E::FT,
+    q_icl::FT,
+    q_rai::FT,
+    ρ::FT,
```
trontrytel, line 570: Should we be passing the size_distr_cache here also?

src/Microphysics1M.jl
```diff
@@ -457,26 +492,36 @@ Internal low-level kernel. Prefer the option-dispatched API.
...
-    end
-    return accr_rate
+@inline function accretion(
+    cloud::CMP.CloudCondensateType,
+    precip::CMP.PrecipitationType,
+    vel::Union{CMP.Blk1MVelTypeRain{FT}, CMP.Blk1MVelTypeSnow{FT}},
+    E::FT,
+    q_clo::FT,
+    q_pre::FT,
+    ρ::FT,
```
trontrytel, line 523: Should we be passing the cached size distr params here also?

trontrytel (approved):
Thank you. Left some small comments to understand the improvement pattern. And a question about keeping the commonly defined epsilon throughout the ifelse conditions.

src/P3_size_distribution.jl
tapios: Storing `λ = exp(logλ)` once makes the slope term a single `λ*D` multiply in the quadrature loop instead of `exp(logλ+logD)` per call. I added a comment.

src/Utilities.jl
tapios: Changed to `loggamma_a`

src/P3_particle_properties.jl
tapios: Restored comments on regimes.

src/P3_particle_properties.jl
tapios: Also restored comments on regimes

src/P3_processes.jl
tapios: Unclear. This is Chebychev quadrature with an endpoint singularity, so should need more points than the Gauss-Legendre quadrature. Left a comment with a TODO to run convergence checks.

src/Microphysics2M.jl
tapios: The predicate is only evaluated once in `cond` and then reused by 3 selects. Faster because it removes warp divergence.

src/Microphysics2M.jl
tapios: `av` and `bv` are folded into `a_vent_0_coeff` and `b_vent_0_coeff` in the `EvaporationSB2006` constructor, so they are no longer needed here.

src/MicrophysicsNonEq.jl
tapios: I changed this to `safe_q = UT.clamp_to_nonneg(q)`. (It's only needed to keep the cbrt safe in the discarded ifelse branch, which has `q > UT.ϵ_numerics(FT)`.)

tapios: Also changed to `safe_q = UT.clamp_to_nonneg(q)`. Same as above.

src/MicrophysicsNonEq.jl
tapios: Same as above.

src/Microphysics1M.jl
tapios: Agreed. I cleaned this up. There were conflated uses: physical and numerical thresholds. I kept one central `ϵ_numerics*` threshold per moment class and used `clamp_to_nonneg` for numerical safekeeping. Added a comment to make this clearer.

src/Microphysics1M.jl
tapios: You are hurting Gemini's feelings... More seriously, BMT builds `sd` once per grid point. `sd = size_distr_parameters(…)' is only used for standalone callers. Added comment to clarify.

src/Microphysics1M.jl
tapios: Done

src/Microphysics1M.jl
tapios: These are standalone convenience wrappers not used inside the BMT broadcasts. They go through the option-dispatched methods below that already have `sd`. Left it as it is so we can call the kernels standalone.

src/Microphysics1M.jl
tapios: As above, they are standalone wrappers not used on the hot path. Easier to leave so we can call them standalone.

tapios:
> Thank you. Left some small comments to understand the improvement pattern. And a question about keeping the commonly defined epsilon throughout the ifelse conditions.

Thanks so much for the thorough review!

PR #752, opened 2026-07-02
ci: refresh dev-guides sync workflow from template

imreddyTeja:
Replaces the hand-maintained `update_dev_guides.yml` with the current template from CliMA/DeveloperGuides.

This fixes a permissions bug in the workflow that prevented the monthly sync job from opening PRs correctly.


imreddyTeja:
Also repaired the broken subtree metadata: the original vendoring was never linked correctly (`git subtree pull` failed with "was never added"). Removed and re-added the `docs/dev-guides` subtree per the template's documented recovery steps, so the monthly sync will work going forward.

PR #757, opened 2026-07-02
refactor(P3,2M): regime selector, aspect-ratio types, and collision-path fixes

haakon-e:
Mechanical refactors and small fixes split out from the P3 quadrature work (#741), so that PR can focus on the numerics. No public behavior changes except where noted.

- `regime_value` is a single branchless selector for the five P3 mass/area regimes, replacing the duplicated `if/elseif/else` conditions in `ice_area`, `ice_mass_coeffs`, and `ϕ_material_density`. Branch values are promoted to a common type, so the selection stays concretely typed when branches mix parameter-typed and state-typed values (`F_rim`, `ρ_g` under differentiation) and allocates nothing. `ice_mass_coeffs` selects the power-law scale and exponent separately, so `regime_value` only ever selects scalars.
- `ϕᵢ` uses the per-regime material density (value change: the aspect-ratio correction is lower in the nonspherical regimes; reference values updated). The [aspect-ratio section of the P3 documentation](https://clima.github.io/CloudMicrophysics.jl/dev/P3Scheme/) derives the closure.
- The terminal-velocity aspect-ratio treatment is a type stored on `ParametersP3` (`aspect_ratio = Oblate()` or `NoAspectRatio()`), replacing the `use_aspect_ratio::Bool` keyword on the terminal-velocity functions. Every consumer reads it from `state.params`, so it is set once at parameter construction and the velocity functions take no aspect-ratio argument.
- `Chen2022VelocityCurve` is defined next to `Chen2022_vel_coeffs` and constructs directly from the velocity parameters and air density; `particle_terminal_velocity` delegates to it. The raw coefficient tuples remain available through `Chen2022_vel_coeffs` for the consumers that integrate them analytically (the closed-form rain-ice collision path and the 1M/2M bulk fall speeds).
- The 2M mass-limit number tendencies are consolidated into `number_tendency_from_mass_limits`, removing the redundant increase/decrease functions.
- The collision-integral closures capture `FT`-typed values instead of the `FT` type object; a closed-over `DataType` lowers `FT(x)`/`zero(FT)` to runtime dispatch on Julia ≤ 1.11, which allocated in the collision integrals.
- `Chen2022_exponential_pdf` computes its integer-factorial normalization with `fac` instead of a hand-evaluated `ifelse` chain (`Base.factorial` reads a host-memory table, which is not GPU-compatible).
- `fac` throws `DomainError` for `n` outside `[0, 20]` instead of silently overflowing.
- The air density is floored to non-negative in the Chen velocity coefficients; a slightly negative sub-domain density otherwise throws in their fractional powers.
- The `gamma_inc` shape assertion uses a single string literal, split over source lines with a line continuation (runtime string concatenation does not compile to GPU device code).


haakon-e:

- #757 ◀
    - #741

haakon-e:
TODO (in a future PR, will write more detailed issue): 

unify the terminal-velocity parameterization. This PR co-locates `Chen2022VelocityCurve` with `Chen2022_vel_coeffs` in `Common`, but the parameter tables (`CMP.Chen2022VelType`) and the aspect-ratio treatment (`CMP.AspectRatio`, stored on `ParametersP3`) remain in separate locations, and the curve stores parallel `ai`/`bi`/`ci` tuples rather than per-term `(a, b, c)` groups. `Chen2022VelocityCurve` exists for type stability: a concretely-typed callable that `velocity_integral_bounds`, the collision-rate kernel, and the closed-form rain path can evaluate and query without branch-dependent closure types. This consolidation of the parameterization (one location, per-term coefficient grouping, one composed ice-velocity object) comes in a follow-up PR.

src/Common.jl
```diff
@@ -345,6 +348,38 @@ end
...
+    Chen2022VelocityCurve(velocity_params, ρₐ)
+    Chen2022VelocityCurve(velocity_params, ρₐ, ρᵢ)
+    Chen2022VelocityCurve(ai, bi, ci)
+
+Callable holding the Chen 2022 terminal-velocity coefficients `(ai, bi, ci)`,
+from [`Chen2022_vel_coeffs`](@ref) evaluated at air density `ρₐ` (and apparent
+ice density `ρᵢ` for the small/large-ice tables). Evaluating it at a diameter
+`D` returns `∑ₖ aₖ D^bₖ exp(-cₖ D)` [m/s].
+"""
+struct Chen2022VelocityCurve{N, FT}
```
haakon-e, line 361: I don't think this is the "final form" of how I want to handle the velocity internally, but was the easiest way to ensure GPU compatibility across Julia versions.

src/Microphysics2M.jl
```diff
@@ -853,52 +852,6 @@ Uses the same approximation pattern as
     return (; ∂N_rai, ∂q_rai)
 end
 
-"""
```
haakon-e, line 856: I had written a combined implementation a while ago, just deleting old duplicated code here.

src/P3_particle_properties.jl
```diff
@@ -401,52 +415,59 @@ Return the cross-sectional area of a particle based on where it falls in the
...
+See also [`ϕ_material_density`](@ref) and the
+[aspect-ratio section of the P3 documentation](@ref "Aspect ratio") for the
+spheroid derivation and the residual `ϕ > 1` band above `D_th`.
 """
 @inline function ϕᵢ(state::P3State, D)
     FT = eltype(D)
+    mᵢ = ice_mass(state, D)
     aᵢ = ice_area(state, D)
-    vol = CO.volume_sphere_D(D)
+    ρ = ϕ_material_density(state, D)
```
haakon-e, line 468: this was the bug in the aspect ratio.

src/P3_processes.jl
```diff
@@ -390,18 +390,22 @@ B_rim is computed by quadrature
     v_l = CO.particle_terminal_velocity(vel.rain, ρₐ)
     v_i = ice_particle_terminal_velocity(vel, ρₐ, state)
     D_min, D_max = bounds_r
+    # Close over `FT`-typed values, not the `FT` type: closing over a `DataType`
+    # turns `FT(π)`/`zero(FT)` into runtime dispatch.
+    zero_rates = (zero(FT), zero(FT), zero(FT))
+    πFT = FT(π)
     function liquid_integrals(Dᵢ)
         if iszero(N₀r) || !(D_max > D_min)
-            return (zero(FT), zero(FT), zero(FT))
```
haakon-e, line 395: FT inside of the inner function was a reason Julia 1.11 failed. Apparently Julia 1.12 is resistant to this type of variable "boxing"

trontrytel: Good to know. Thanks!

src/P3_particle_properties.jl
```diff
@@ -304,6 +304,33 @@ function weighted_average(f_a, a, b)
     return f_a * a + (1 - f_a) * b
 end
 
+"""
+    regime_value(state::P3State, D, small, unrimed, dense_rimed, graupel, partially_rimed)
+
+Select the value for the P3 mass/area regime that the maximum dimension `D` falls in:
```
trontrytel, line 310:
```suggestion
Returns the value for the P3 mass/area regime that the maximum dimension `D` falls in:
```
Select could be misleading and suggest that the user needs to select something?

trontrytel (approved):
LGTM!

Issue #766, opened 2026-07-17
Split process options from parameters for Microphysics1MOptions


nefrathenrici:
Right now the 1-moment option types in `src/parameters/Microphysics1MOptions.jl` pick which process to run and hold the parameters for the given process. This means that you can't choose a process without a full parameter set, complicating the atmosphere model interface downstream.

Types that carry parameters (found by claude): CloudLiquidFormation, ConstantTimescale, Kessler1M, NoSupersaturation, CloudLiquidRainAccretion, CloudLiquidSnowAccretion, CloudIceRainAccretion, CloudIceSnowAccretion, RainSnowAccretion

Example: Kessler1M currently holds parameters:
```julia
struct Kessler1M{AC} <: RainAutoconversion
    acnv1M::AC          # τ, q_threshold, k
end
```
After the proposed refactor, the struct is just a singleton and the parameters are stored in `process_params` in `Microphysics1MParams`
```julia
struct Kessler1M <: RainAutoconversion end
@kwdef struct Microphysics1MParams{OPT, PProc, CP, PP, AP, VL} <: ParametersType
    options::OPT
    process_params::PProc   # NEW
    cloud::CP
    precip::PP
    air_properties::AP
    terminal_velocity::VL
end
```

nefrathenrici:
This is proposed to resolve https://github.com/CliMA/ClimaAtmos.jl/issues/4709

nefrathenrici:
Closed by #768

PR #767, opened 2026-07-21
 Move parameters out 1M options types into Microphysics1MParams 

nefrathenrici:
Changes
- Made the 1M option types singletons.
- Moved 1M options parameters to new field `process_params` in `Microphysics1MParams`. `process_params_for(option, toml_dict)` reads the parameter values for each selected process.
- The 1M tendency functions now read values from `mp.process_params.<slot>` instead of off the option struct. Dropped the now-unused opt argument names.
- Separated two condensation/deposition timescale kernels out (`_conv_q_vap_to_q_lcl_const`, `_conv_q_vap_to_q_icl_const`) so the 2M scheme can reuse them. Temperature-dependent ice stays inlined.
- Added defaults to `Microphysics1MOptions`.

Some benchmark tests fail on the 1.12 runner, I would like to increase the time limit since there is no type-stability change.

### DIFF
--- a/docs/bibliography.bib
+++ b/docs/bibliography.bib
@@ -373,19 +373,6 @@ url  ="http://dx.doi.org/10.1039/C3FD00035D"
   year = {2016}
 }
 
-@Article{KnopfAlpert2013,
-  author = "Knopf, Daniel A. and Alpert, Peter A.",
-  title  = "A water activity based model of heterogeneous ice nucleation kinetics for freezing of water and aqueous solution droplets",
-  journal = "Faraday Discuss.",
-  year = "2013",
-  volume = "165",
-  issue = "0",
-  pages = "513-534",
-  publisher = "The Royal Society of Chemistry",
-  doi = "10.1039/C3FD00035D",
-  url = "http://dx.doi.org/10.1039/C3FD00035D",
-}
-
 @article{Koop2000,
 author = {Koop, Thomas and Luo, Beiping and Tsias, Athanasios and Peter, Thomas},
 title = {Water activity as the determinant for homogeneous ice nucleation in aqueous solutions},
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -340,7 +340,7 @@ the option stored in `Microphysics1MOptions`.
 using the prescribed cloud droplet number concentration.
 
 # Arguments
-- `option`: `nothing`, `Kessler1M(...)`, or `PrescribedNd(...)`
+- `option`: `nothing`, `Kessler1M()`, or `PrescribedNd()`
 - `mp`: 1-moment microphysics parameters
 - `tps`: thermodynamics parameters (unused, kept for uniform interface)
 - `micro`: microphysics state `(; q_tot, q_lcl, q_icl, q_rai, q_sno)`
@@ -351,15 +351,15 @@ using the prescribed cloud droplet number concentration.
 """
 @inline conv_q_lcl_to_q_rai(::Nothing, mp, tps, micro, thermo) = zero(micro.q_lcl)
 
-@inline function conv_q_lcl_to_q_rai(opt::CMP.Kessler1M, mp, tps, micro, thermo)
+@inline function conv_q_lcl_to_q_rai(::CMP.Kessler1M, mp, tps, micro, thermo)
     q_lcl = micro.q_lcl
-    (; τ, q_threshold, k) = opt.acnv1M
+    (; τ, q_threshold, k) = mp.process_params.rain_autoconversion
     return CO.logistic_function_integral(q_lcl, q_threshold, k) / τ
 end
 
-@inline function conv_q_lcl_to_q_rai(opt::CMP.PrescribedNd, mp, tps, micro, thermo)
+@inline function conv_q_lcl_to_q_rai(::CMP.PrescribedNd, mp, tps, micro, thermo)
     q_lcl = micro.q_lcl
-    (; τ, α, Nc) = opt.autoconv
+    (; τ, α, Nc) = mp.process_params.rain_autoconversion
     return max(0, q_lcl) / (τ * (Nc / 100_000_000)^α)
 end
 
@@ -403,7 +403,7 @@ for use in simulations without supersaturation (e.g., with saturation adjustment
 Harrington et al. (1995) and Kaul et al. (2015).
 
 # Arguments
-- `opt`: `nothing`, `NoSupersaturation(...)`, or `WithSupersaturation(...)`
+- `opt`: `nothing`, `NoSupersaturation()`, or `WithSupersaturation()`
 - `mp`: 1-moment microphysics parameters
 - `tps`: thermodynamics parameters
 - `micro`: microphysics state `(; q_tot, q_lcl, q_icl, q_rai, q_sno)`
@@ -411,18 +411,18 @@ Harrington et al. (1995) and Kaul et al. (2015).
 """
 @inline conv_q_icl_to_q_sno(::Nothing, mp, tps, micro, thermo, sd = nothing) = zero(micro.q_icl)
 
-@inline function conv_q_icl_to_q_sno(opt::CMP.NoSupersaturation, mp, tps, micro, thermo, sd = nothing)
-    (; τ, q_threshold, k) = opt.acnv1M
+@inline function conv_q_icl_to_q_sno(::CMP.NoSupersaturation, mp, tps, micro, thermo, sd = nothing)
+    (; τ, q_threshold, k) = mp.process_params.snow_autoconversion
     q_icl = micro.q_icl
     return CO.logistic_function_integral(q_icl, q_threshold, k) / τ
 end
 
 @inline function conv_q_icl_to_q_sno(
-    opt::CMP.WithSupersaturation, mp, tps, micro, thermo, sd = size_distr_parameters(mp, micro, thermo),
+    ::CMP.WithSupersaturation, mp, tps, micro, thermo, sd = size_distr_parameters(mp, micro, thermo),
 )
     (; q_tot, q_lcl, q_icl, q_rai, q_sno) = micro
     (; ρ, T) = thermo
-    r_ice_snow = opt.r_ice_snow
+    r_ice_snow = mp.process_params.snow_autoconversion.r_ice_snow
     (; pdf, mass) = mp.cloud.ice
     aps = mp.air_properties
     FT = eltype(ρ)
@@ -693,7 +693,7 @@ delegate to the corresponding low-level Marshall-Palmer kernels.
 `nothing` variants return zero without computing anything.
 
 # Arguments
-- `opt`: accretion option type (carries collision efficiency) or `nothing`
+- `opt`: accretion option type (selects the process) or `nothing`
 - `mp`: `Microphysics1MParams`
 - `tps`: thermodynamics parameters
 - `micro`: microphysics state `(; q_tot, q_lcl, q_icl, q_rai, q_sno)`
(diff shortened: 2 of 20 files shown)


docs/bibliography.bib
```diff
@@ -373,19 +373,6 @@ @article{Kirkby2016
   year = {2016}
 }
 
-@Article{KnopfAlpert2013,
```
haakon-e, line 376: why remove this? is it unused?

src/parameters/Microphysics1MOptions.jl
```diff
@@ -371,68 +238,176 @@ $(DocStringExtensions.FIELDS)
...
+    snow_deposition_sublimation = process_params_for(o.snow_deposition_sublimation, td),
+    snow_melt = process_params_for(o.snow_melt, td),
+    cloud_liquid_rain_accretion = process_params_for(o.cloud_liquid_rain_accretion, td),
+    cloud_liquid_snow_accretion = process_params_for(o.cloud_liquid_snow_accretion, td),
+    cloud_ice_rain_accretion = process_params_for(o.cloud_ice_rain_accretion, td),
+    cloud_ice_snow_accretion = process_params_for(o.cloud_ice_snow_accretion, td),
+    rain_snow_accretion = process_params_for(o.rain_snow_accretion, td),
+)
+
+# ═══════════════════════════════════════════════════════════════════
```
haakon-e: I personally think it's fine to skip deprecation.

src/parameters/Microphysics1MParams.jl
```diff
@@ -68,13 +70,14 @@ mp = CMP.Microphysics1MParams(Float64)
...
-    cloud_ice_formation = CMP.TemperatureDependent(toml_dict),
+    cloud_ice_formation = CMP.TemperatureDependent(),
     cloud_ice_melt = nothing,
 )
 ```
 """
-@kwdef struct Microphysics1MParams{OPT, CP, PP, AP, VL} <: ParametersType
+@kwdef struct Microphysics1MParams{OPT, PPR, CP, PP, AP, VL} <: ParametersType
     options::OPT
+    process_params::PPR
```
haakon-e, line 86: small nit: I think it would be more intuitive if the naming between `options` and `process_params` was uniform. That is, name both either:
- `options`, `option_params`
- `processes`, `process_params`

I have a slight preference towards the latter, but bike shedding like this can also wait until another day.

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,17 +35,19 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
-are stored directly.
+Process selection (which variant of each process runs) lives in `options`. The
+parameter values each selected variant needs (relaxation timescales,
+autoconversion thresholds, collision efficiencies) live in `process_params`,
+whose fields mirror `options`. Shared parameters (particle size distributions,
+air properties, terminal velocities) are stored directly.
 
 # Fields
-- `options::OPT`: Microphysics1MOptions — process configuration (carries process-specific parameters)
+- `options::OPT`: Microphysics1MOptions — process selection (no parameters)
```
haakon-e:
```suggestion
- `options::OPT`: Microphysics1MOptions — process selection
```

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,17 +35,19 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
+- `process_params::PPR`: per-process parameter data matching `options` (or `nothing` for parameter-free/disabled processes)
```
haakon-e: what does 

> `nothing` for [...] disabled processes

mean?

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,17 +35,19 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
 
-Process-specific parameters (relaxation timescales, autoconversion thresholds,
-collision efficiencies) live inside the option types in `options`.
-Shared parameters (particle size distributions, air properties, terminal velocities)
-are stored directly.
+Process selection (which variant of each process runs) lives in `options`. The
+parameter values each selected variant needs (relaxation timescales,
+autoconversion thresholds, collision efficiencies) live in `process_params`,
+whose fields mirror `options`. Shared parameters (particle size distributions,
+air properties, terminal velocities) are stored directly.
```
haakon-e: Perhaps an example would help make this a bit clearer? The language is a bit dense and abstract right now, I think, for something that shouldn't be that hard to grasp.

src/MicrophysicsNonEq.jl
```diff
@@ -108,11 +108,15 @@ Morrison & Milbrandt (2015), https://doi.org/10.1175/JAS-D-14-0065.1.
...
 @inline conv_q_vap_to_q_lcl(::Nothing, mp, tps, micro, thermo) = zero(thermo.T)
-@inline function conv_q_vap_to_q_lcl(
-    opt::CMP.CloudLiquidFormation, mp, tps::TDI.PS, micro, thermo)
+@inline conv_q_vap_to_q_lcl(
+    ::CMP.CloudLiquidFormation, mp, tps::TDI.PS, micro, thermo) =
+    _conv_q_vap_to_q_lcl_const(
+        mp.process_params.cloud_liquid_formation.τ_relax, tps, micro, thermo)
+
+# Kernel for `conv_q_vap_to_q_lcl`with a constant relaxation timescale `τ`.
+@inline function _conv_q_vap_to_q_lcl_const(τ, tps::TDI.PS, micro, thermo)
```
haakon-e, line 117: why do we have inner `_`-functions for this and `_conv_q_vap_to_q_icl_const`?

haakon-e (approved):
Looks good. some minor inline comments attached.

The biggest piece that it would be useful if this PR documented (at least in the PR description) is why this change is desired. Some motivation is provided in https://github.com/CliMA/CloudMicrophysics.jl/issues/766 which appear to primarily point to ergonomics:

> [currently,] you can't choose a process without a full parameter set, complicating the atmosphere model interface downstream.

If I understood some offline conversations correctly, this change could also be beneficial for performance. It would be useful to spell that out (and test it) in this PR too.

docs/bibliography.bib
nefrathenrici: It's a duplicate of an identical entry that still exists in the bibliography.

src/parameters/Microphysics1MOptions.jl
nefrathenrici: Sounds good, I can make a breaking release.

src/parameters/Microphysics1MParams.jl
nefrathenrici: A process is "disabled" when its option field is set to nothing. "Parameter-free" options  like CloudIceMelt() also map to nothing. I can word this more clearly.

src/MicrophysicsNonEq.jl
nefrathenrici: The `_..._const` kernels hold the shared constant-timescale physics so callers like the 2MP3 or the parcel model can reuse it without supplying Microphysics1MParams

src/parameters/Microphysics1MParams.jl
nefrathenrici: Agreed processes / process_params is clearer, I can make the change.

nefrathenrici: Actually, it may be a larger effort to get everything named consistently. I think it should be done in a later PR.

src/parameters/Microphysics1MOptions.jl
```diff
@@ -70,265 +74,128 @@ abstract type SnowDepositionSublimation <: MicrophysicsOption end
...
 
 """
-    CloudLiquidFormation{FT} <: MicrophysicsOption
+    CloudLiquidFormation <: MicrophysicsOption
 
 Constant relaxation timescale for liquid condensation and evaporation.
-
-# Fields
-$(DocStringExtensions.FIELDS)
+Parameters (`τ_relax`) are stored in `process_params.cloud_liquid_formation`.
```
haakon-e: many docstrings mention where some parameter is stored. could this be made a bit more explicit by mentioning the struct storing them, e.g.,
```Julia
stored in `process_params.cloud_liquid_formation` in [`Microphysics1MParams`](@ref)
```
or similar

src/parameters/Microphysics1MParams.jl
```diff
@@ -35,17 +35,23 @@ Base.show(io::IO, mime::MIME"text/plain", x::PrecipPhaseParams1M) =
...
+`process_params.rain_autoconversion`. Turning a process off with `nothing` (e.g.
+`options.cloud_ice_melt = nothing`) gives it a `nothing` slot in
+`process_params`, as does any option that needs no parameters. Shared parameters
+(particle size distributions, air properties, terminal velocities) are stored
+directly.
 
 # Fields
-- `options::OPT`: Microphysics1MOptions — process configuration (carries process-specific parameters)
+- `options::OPT`: Microphysics1MOptions — process selection
+- `process_params::PPR`: parameter data for each selected process, mirroring `options` (`nothing` when a process is disabled with `nothing` or needs no parameters)
```
haakon-e: nit: add newline

src/parameters/Microphysics1MParams.jl
haakon-e: I agree, let's make that suggestion a separate PR. But it'd also be good to hear from @trontrytel if she has thoughts about this bike shed's color:)

src/MicrophysicsNonEq.jl
haakon-e: thanks for the explanation!

haakon-e (approved):

src/parameters/Microphysics1MParams.jl
trontrytel: Yes on consistency. No preferences on the color of the bikeshed

src/Microphysics1M.jl
```diff
@@ -351,15 +351,15 @@ using the prescribed cloud droplet number concentration.
 """
 @inline conv_q_lcl_to_q_rai(::Nothing, mp, tps, micro, thermo) = zero(micro.q_lcl)
 
-@inline function conv_q_lcl_to_q_rai(opt::CMP.Kessler1M, mp, tps, micro, thermo)
+@inline function conv_q_lcl_to_q_rai(::CMP.Kessler1M, mp, tps, micro, thermo)
     q_lcl = micro.q_lcl
-    (; τ, q_threshold, k) = opt.acnv1M
+    (; τ, q_threshold, k) = mp.process_params.rain_autoconversion
```
trontrytel, line 356: Sorry, I'm a little late into this. I thought we wanted the `process_params` struct to be flat? Since we are no longer dispatching, there is no need for the additional layer of `.rain_autoconversion` ?

nefrathenrici: The per-process key is still needed, but not for dispatch. `process_params` holds the params for every enabled process at once and some of these parameters have the same name. To flatten this structure, every parameter would need to have a unique name.

src/parameters/Microphysics1MParams.jl
nefrathenrici: Great, I added the changes to https://github.com/CliMA/CloudMicrophysics.jl/pull/768

src/Microphysics1M.jl
trontrytel: Yes. I thought that was the idea? I don't think there are that many with doubled names tbh. I thought it would be easier on the compiler to have a flat struct here, rather than more nesting?

PR #768, opened 2026-07-24
Add deprecations to avoid breaking interface

nefrathenrici:
The previous PR #767 would require a minor (breaking) release because it removed the toml-based constructors for 1M options ( e.g. `CloudLiquidFormation(::CP.ParamDict)`). This PR adds the constructors back in with deprecations.


trontrytel:
So thats why ClimaAtmos no longer works for me with the new CM :)

Thank you for fixing it!

trontrytel (approved):
Would be good to add a test for it. It may be deprecated, but thats how Atmos is using the library now?

PR #769, opened 2026-07-28
Guard 1M process rates to avoid kernel spilling

petebachant:
This is a 15% speedup for AMIP (6% against ClimaCore v0.15.0), bringing the `set_microphysics_tendency_cache` kernel down from 13.8% to 6.9% of the overall GPU time.

## TODO

- [x] Merge in and test against latest `main`
- [ ] Understand the changes before requesting a review
- [x] Try to make this speedup measurable inside this package with a lighter weight test


petebachant:
@dennisYatunin any thoughts on this approach as a way to detect register spilling more cheaply in higher level packages?

dennisYatunin:
I think this is an excellent approach! I had Claude make something similar while working on ClimaCore (a generic tool for testing compilation on GPUs without any available devices, available [here](https://github.com/CliMA/DeveloperGuides/tree/main/.claude/skills/fix-ci)), which I've been meaning to generalize into unit tests like you added here. We'll probably want to centralize this across CliMA, either bundling it with ClimaCore or moving it into a new package. Being able to catch GPU issues before they appear in downstream packages will be really helpful for us.

trontrytel:
Should we merge it then?

petebachant:
> Should we merge it then?

Sure, if you don't mind reviewing, this is probably worth merging, since it does stop the register spill.

petebachant:
Re-tested against ClimaCore 0.15.2 and it still eliminates the register spilling. Would be good to get the team's opinions on readability/understandability.

petebachant:
After the recent 15% drop in performance for the flagship AMIP, this change is now only a ~1% benefit, so I'm going to close until a better solution is found.

trontrytel:
Sorry, I was on holidays. Even if we don't want the code changes, do we want to add the tests?

petebachant:
> Sorry, I was on holidays. Even if we don't want the code changes, do we want to add the tests?

No worries! I'm playing around with moving these checks down into ClimaCore, but if they're still valuable here, I'll open up a new PR.

PR #770, opened 2026-07-30
Rain formation as a function of air velocity

trontrytel:
This PR adds rain autoconversion function that depends on air vertical velocity. The idea is to differentiate between the stratiform (less turbulent, slower) and convective (more turbulent, fast) regime. 

Screenshot 2026-09-04 at 12 15 47 PM

This requires adding vertical velocity to bulk microphysics tendencies. So once approved this will be a breaking release for Atmos.

### DIFF
--- a/src/BulkMicrophysicsTendencies.jl
+++ b/src/BulkMicrophysicsTendencies.jl
@@ -128,7 +128,7 @@ linearized operator construction consume this output.
 Constructs two `NamedTuple`s that are passed to all process functions
 (see `Microphysics1M` module docs for the full convention):
 - `micro = (; q_tot, q_lcl, q_icl, q_rai, q_sno)` — specific humidities (kg/kg)
-- `thermo = (; ρ, T)` — air density (kg/m³) and temperature (K)
+- `thermo = (; ρ, T, w)` — air density (kg/m³), temperature (K) and vertical velocity [m/s]
 
 Naming convention: `S_process_species1_species2`
  - process: physical mechanism (phase_change, acnv, accr, melt, accr_melt, accr_freeze)
@@ -140,7 +140,7 @@ processes are pre-routed by temperature, so consumers never need `is_warm`.
 """
 @inline function _microphysics_source_terms(
     ::Microphysics1Moment, mp::CMP.Microphysics1MParams, tps,
-    ρ, T, q_tot, q_lcl, q_icl, q_rai, q_sno,
+    ρ, T, w, q_tot, q_lcl, q_icl, q_rai, q_sno,
 )
     # Clamp negative inputs to zero (robustness against numerical errors)
     ρ = UT.clamp_to_nonneg(ρ)
@@ -155,7 +155,7 @@ processes are pre-routed by temperature, so consumers never need `is_warm`.
 
     # Construct state tuples (reused across all process calls)
     micro = (; q_tot, q_lcl, q_icl, q_rai, q_sno)
-    thermo = (; ρ, T)
+    thermo = (; ρ, T, w)
 
     # Size-distribution / fall-speed quantities (λ⁻¹, n₀, v₀ for rain/snow/ice) are
     # pow/exp-heavy and shared by several processes per species: compute once and pass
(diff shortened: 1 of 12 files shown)


src/Microphysics1M.jl
```diff
@@ -383,10 +384,44 @@ end
...
+end
+
+@inline function rain_autoconversion_timescale(::CMP.PrescribedNd, mp, w = 0)
+    (; τ, α, Nc) = mp.process_params.rain_autoconversion
+    return τ * (Nc / 100_000_000)^α
+end
+
+@inline function rain_autoconversion_timescale(::CMP.VelocityDependent, mp, w)
+    (; τ_slow, τ_fast, w_0) = mp.process_params.rain_autoconversion
+    f = w^4 / (w^4 + w_0^4)
```
szy21, line 416: any physical reason for the fourth power?

szy21 (approved):

src/Microphysics1M.jl
```diff
@@ -383,10 +384,44 @@ end
...
+    return τ_slow + (τ_fast - τ_slow) * f
+end
```
szy21, line 418: My claude suggests using `_consistent_params` here to so that you get an error instead of silently getting a wrong answer when you e.g. use Kessler1M with PrescribedNd parameters. So something like:
```
@inline function rain_autoconversion_timescale(opt::CMP.Kessler1M, mp, w)
    pp = _consistent_params(mp.process_params.rain_autoconversion, CMP.Acnv1M, opt, :rain_autoconversion)
    return pp.τ
end
```
But it's only used in ClimaAtmos diagnostics? So up to you whether it needs to be more strict.

test/gpu_tests.jl
```diff
@@ -696,6 +707,25 @@ function test_gpu(FT)
...
+        (; output, ndrange) = setup_output(3, FT, bad_value)
+        ql = ArrayType([FT(1e-3), FT(1e-3), FT(0)])
+        w_arr = ArrayType([FT(0), FT(5), FT(3)])
+
+        kernel! = test_1_moment_veldep_acnv_kernel!(backend, work_groups)
+        kernel!(mp_1m_vd, tps, output, ql, w_arr; ndrange)
+        out = Array(output)
+
+        # Sanity checks
+        TT.@test !any(isequal(out, FT(bad_value)))
```
szy21, line 721: Also from claude: this is comparing an array to a number so it is always true (same for the other checks in this file). Changing it to `!any(isequal(FT(bad_value)), out)` should work.

src/Microphysics1M.jl
trontrytel: Fixed! My Claude says hi to your Claude.

test/gpu_tests.jl
trontrytel: Fixed!

src/Microphysics1M.jl
trontrytel: Just wanted something smooth that still provides a fast transition between the two regimes. I added some comments in the docs

trontrytel:
This is a breaking release for Atmos, because it requires BMTs to have vertical velocity as input. I already have a branch in Atmos that updates this. But with all the other issues we are having there right now, I'll hold off till next week with merging it. I don't want to add to the reverting PRs

src/BulkMicrophysicsTendencies.jl
```diff
@@ -637,13 +638,14 @@ active microphysical processes, including regime changes near freezing.
...
+    w,
     q_tot,
     q_lcl,
     q_icl,
     q_rai,
     q_sno,
-    Δt,
-    nsub = 1,
+    Δt::AbstractFloat,
+    nsub::Integer = 1,
```
trontrytel, line 648: Claude was worried that if we don't pass `w` in `Atmos` somehow, it would silently proceed using `q_tot` for `w`, `q_lcl` for `q_tot`, etc and assuming `nsub=1`. Typing the last two arguments prevents that error.

sajjadazimi (approved):
Thank you! It looks good to me! My only comment is that we could merge both new closures into one type. It might be useful to have different thresholds and different time scales for different regimes. Any of them can be deactivated by setting their corresponding fast and slow parameters equal (for example if I only want different time-scales I can set thresholds equal).

szy21:
Yes, I also think it would be better to merge the two types

trontrytel:
Sounds good. I collapsed all three variants: Kessler, variable timescale and variable threshold into one. By default the ClimaParams will just give you the old Kessler. This is still a breaking release, because I need to pass w to Atmos.

PR #775, opened 2026-08-06
Bulk tendencies: limit vapor to condensate sources

sajjadazimi:
## Purpose 
Bulk tendencies: cap vapor to condensate sources so q_v stays ≥ min(q_sat_liq, q_sat_ice)

----

### DIFF
--- a/src/BulkMicrophysicsTendencies.jl
+++ b/src/BulkMicrophysicsTendencies.jl
@@ -409,6 +409,19 @@ exponential decays over the substep.
 
     invΔt = one(FT) / Δt
 
+    # Cap vap→condensate sources jointly so the substep cannot drive
+    # `q_v` below `min(q_sat_liq, q_sat_ice)`. Preserves relative rates.
+    q_sat_min = min(
+        TDI.saturation_vapor_specific_content_over_liquid(tps, T, ρ),
+        TDI.saturation_vapor_specific_content_over_ice(tps, T, ρ),
+    )
+    q_v = q_tot - q_lcl - q_icl - q_rai - q_sno
+    α = min(
+        one(FT),
+        max(zero(FT), q_v - q_sat_min) * invΔt /
+        max(lin.e1 + lin.e2 + lin.e4, eps(FT)),
+    )
+
     # A = I/Δt - M
     a11 = invΔt - lin.M11
     a12 = -lin.M12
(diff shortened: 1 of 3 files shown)


src/BulkMicrophysicsTendencies.jl
```diff
@@ -421,12 +434,12 @@ exponential decays over the substep.
     a43 = -lin.M43
     a44 = invΔt - lin.M44
 
-    # rhs = e + q_0/Δt
+    # rhs = e + q_0/Δt (vap→condensate `e` terms scaled by `α` above)
     # e3 = 0 by the 1m model
-    b1 = lin.e1 + invΔt * q_lcl
-    b2 = lin.e2 + invΔt * q_icl
+    b1 = α * lin.e1 + invΔt * q_lcl
```
trontrytel, line 439: We could consider making the coefficient different for liquid and ice part? I.e proportional to the difference over liquid for liquid, and over ice for ice. That would keep it in line with the WBF process that puts different rate for condensation vs deposition. Or is it important for the numerics that it is the same?

src/BulkMicrophysicsTendencies.jl
```diff
@@ -409,6 +409,19 @@ exponential decays over the substep.
...
+    # Cap vap→condensate sources jointly so the substep cannot drive
+    # `q_v` below `min(q_sat_liq, q_sat_ice)`. Preserves relative rates.
+    q_sat_min = min(
+        TDI.saturation_vapor_specific_content_over_liquid(tps, T, ρ),
+        TDI.saturation_vapor_specific_content_over_ice(tps, T, ρ),
+    )
+    q_v = q_tot - q_lcl - q_icl - q_rai - q_sno
+    α = min(
+        one(FT),
+        max(zero(FT), q_v - q_sat_min) * invΔt /
```
trontrytel, line 421: Our source term is proportional to `q_v - q_sat` But does it mean our limiter has to be too? Or should we limit by total amount of vapor not going negative?

src/BulkMicrophysicsTendencies.jl
trontrytel: (So that we would end up limiting the 3 processes at a different rate each. But maybe its not worth it)

trontrytel (approved):
Left some comments. Mostly about if the limiter should be proportional to `q_vap` or `q_vap - q_vap_sat`.

src/BulkMicrophysicsTendencies.jl
sajjadazimi: Discussed offline

src/BulkMicrophysicsTendencies.jl
sajjadazimi: Discussed offline

PR #780, opened 2026-08-24
1M correctness and robustness fixes

tapios:
- **Fix a factor `2^α` in the snow aspect-ratio average**: `ϕ_av` used the
  radius-based PSD slope for a diameter-based power law, inflating shape-based
  (`Oblate`/`Prolate`) snow terminal velocities by ~26%. 
- **Make the 1M rates ForwardDiff-differentiable**: removed the
  `::FT ... where {FT}` annotations for consistency with ForwardDiff and other Clima repos,
  with `zero(rate)` fallbacks and `ϵ_numerics` filters; the  integer-hardening happens now 
  inside `Utilities.ϵ_numerics`. New AD tests check rates against central differences in
  both precisions, plus mixed-Dual return-type sweeps in `return_type_tests.jl`.
- **Suppress snow deposition above freezing** (sublimation stays active):
  closes an unphysical (but tiny) vapor→ice growth path reachable during transient
  supersaturation spikes. 
- **Fail clearly on option/parameter mismatches**: passing e.g.
  `WithSupersaturation()` with parameters built for `NoSupersaturation` now
  throws an actionable `ArgumentError` instead of an obscure field error.
- Comment/doc fixes: Chen et al. (2022) table references corrected to
  B1/B3/B5, `cp_l` in the melt factor.

**Results change** only for the shape-based snow terminal velocity (−26%) and for
snow deposition in ice-supersaturated air above freezing (now 0).

Merge this *before* #779

### DIFF
--- a/src/Microphysics1M.jl
+++ b/src/Microphysics1M.jl
@@ -82,11 +82,13 @@ Returns the intercept parameter of the assumed Marshall-Palmer distribution
 - `q_sno`: snow specific content (snow only)
 - `ρ`: air density (snow only)
 """
-@inline function get_n0((; ν, μ)::CMP.ParticlePDFSnow{FT}, q_sno::FT, ρ::FT) where {FT}
-    safe_q_sno = max(q_sno, UT.ϵ_numerics(FT))
-    return ifelse(q_sno > UT.ϵ_numerics(FT), μ * (ρ * safe_q_sno)^ν, zero(FT))
+@inline function get_n0((; ν, μ)::CMP.ParticlePDFSnow, q_sno, ρ)
+    ϵ = UT.ϵ_numerics(q_sno)
+    safe_q_sno = max(q_sno, ϵ)
+    n0 = μ * (ρ * safe_q_sno)^ν
+    return ifelse(q_sno > ϵ, n0, zero(n0))
 end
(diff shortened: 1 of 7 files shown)


src/Microphysics1M.jl
```diff
@@ -224,69 +236,76 @@ Fall velocity of individual particles is parameterized:
...
+    λ_inv,
+)
     (; χv, ve, Δv, gamma_term) = vel
     (; r0, me, Δm, χm, gamma_coeff) = mass
 
     # gamma_term = SF.gamma(me + ve + Δm + Δv + 1) (pre-computed in vel)
     # gamma_coeff = SF.gamma(me + Δm + 1) (pre-computed in mass)
     fall_w = χv * v0 * (λ_inv / r0)^(ve + Δv) * gamma_term / gamma_coeff
-    return ifelse(q > UT.ϵ_numerics(FT), fall_w, zero(FT))
+    return ifelse(q > UT.ϵ_numerics(q), fall_w, zero(fall_w))
```
trontrytel, line 261: In ClimaAtmos, what made the simulation stable for small q and variable velocity, was to assume a constant but non-zero velocity for q smaller than 1e-8. I would therefore rather not increase the epsilon here, and not clip the velocity to zero

tapios: The semantics is here unchanged, though, isn't it?

trontrytel (approved):

PR #781, opened 2026-08-31
Prescribed N_ice for vapor -> ice formation in non-eq microphysics

trontrytel:
This PR adds vapor to cloud ice timescale based on prescribed number concentration. The number concentration we already had in parameters, because we use it in sedimentation. 

I also fixed a bug in Frostenberg that was using number concentration per m3 rather than kg (doesn't really improve the results)

I also also fixed a dispatch issue when you want to switch off cloud liquid snow collisions. 

To be tested in AMIP, but deep convection tests show that with prescribed N we can get much better liquid - ice partitioning.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ MLJ = "add582a8-e3ab-11e8-2d5e-e98b27df1bc7"
 EmulatorModelsExt = ["DataFrames", "MLJ"]
 
 [compat]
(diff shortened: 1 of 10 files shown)


szy21:
Does it make more sense to have deposition and sublimation timescale the same (both depend on N_ice), or to have a fixed sublimation timescale as the Frostenberg?

szy21 (approved):

PR #783, opened 2026-09-02
Cloud water freezing

trontrytel:
This PR:
- Adds homogenous freezing of cloud droplets for T < T homogeneous ice nucleation
- Adds heterogeneous freezing of cloud liquid following Bigg but with parameters for cloud droplets
- Adds a limiter that prevents vapor to liquid condensation for temperatures below T homogeneous ice nucleation


sajjadazimi (approved):
Thanks! I'm not familiar with the freezing parameterizations but the code and the bulk microphysics solver addition look good.

src/BulkMicrophysicsTendencies.jl
sajjadazimi: Could you remove this comment in this PR. det is positive but not for the reason mentioned here. It is positive because a33*a44 is guaranteed to be larger than a34*a43.

trontrytel:
Downstream test is broken because I need to make a release and update ClimaAtmos. Will do it today

PR #786, opened 2026-09-09
Fix/implicit noneq relaxation

trontrytel:
## Purpose

Stabilize the 1-moment linearized substep solver when the vapor ↔ cloud-condensate relaxation timescale is much shorter than the substep. For `cloud_ice_formation = PrescribedIceNumber` (with  `cloud_ice_sedimentation_number_concentration = 5e8 m⁻³`) the relaxation timescale τ = 1.6–7 s. We use 60 s substeps in ClimaAtmos AMIP runs).

## The problem

In `_linearize` / `_linearized_implicit_step`, vapor → condensate sources are constants over the substep. Source terms are capped so that the substep cannot take more vapor than the excess over ice saturation at the substep start temperature. Condensate sinks are implicit decays `-D q` capped by the available condensate. When Δt_sub/τ ≫ 1 both caps become the constraint and their asymmetry creates a cycle:

1. deposition removes the full excess; the latent heating raises q_sat,ice by (Γ−1)·excess (20–45 % at 250–270 K), so the substep ends sub-saturated;
2. sublimation with τ ≈ 1.5 s then removes essentially all the ice, although the vapor deficit is a fraction of it, so the parcel overshoots to supersaturation and cools;
3. repeat.

Offline calling `bulk_microphysics_tendencies(LinearizedAverage(), …)` repeatedly on a supercooled liquid cloud at 265 K (Δt = 180 s, nsub = 3, Float32, AMIP parameters) gives an undamped ±1.2 K / ±4e-4 kg kg⁻¹ oscillation per step; `TemperatureDependent` (τ_sub = 300 s, Frostenberg τ_dep) relaxes monotonically on the same state because with Δt_sub/τ ≈ 0.2 the caps never activate.

## The fix

Each vapor ↔ cloud-condensate phase change is a relaxation toward an equilibrium, `S = (q* − q)/τ`, whose exact integral over a substep is

```
Δq = S Δt / (1 + Δt/τ),        q* = q + S τ
```

Instead of feeding the instantaneous `S` as a constant source (or a `−D q` sink) for the whole substep, the solver now feeds this pre-integrated value for both sources and sinks. `S` already contains `1/Γ` and the available-condensate bound, so `q*` is the saturation-adjusted state including the first-order latent-heat feedback (sources) and at most the available condensate (sinks). The substep can never cross it for any Δt/τ. Other processes on the same row (autoconversion, accretion, freezing, melt) are left untouched. For Δt/τ ≪ 1 the previous scheme is recovered (factor 1/(1+Δt/τ), e.g. −17 % per 60 s substep for the 300 s `TemperatureDependent` sublimation timescale).

Commits:
1. `src/MicrophysicsNonEq.jl`: new `τ_vap_to_q_lcl(opt, …)` / `τ_vap_to_q_icl(opt, …)` returning the relaxation timescale (without Γ) used by the matching `conv_q_vap_to_q_*` method (`Inf` when disabled; for `TemperatureDependent` the sublimation or Frostenberg deposition timescale, matching the tendency). `src/BulkMicrophysicsTendencies.jl`: `_microphysics_source_terms` also returns the two timescales; `_linearize(…, Δt)` adds `S/(1 + Δt/τ)` to `e` for both phase changes; `_linearized_implicit_step` applies the α vapor cap to the positive parts of the drives only. Regression test for the stiff regime added; two pre-existing structural tests unchanged.
2. `src/MicrophysicsNonEq.jl`: the evaporation/sublimation bound is made Γ-consistent, `min(−sat_excess, Γ q)/(τ Γ)` instead of `min(−sat_excess, q)/(τ Γ)`, i.e. the rate is bounded by `q/τ` and the relaxation target is exactly `max(q − deficit/Γ, 0)`. Previously, in air whose deficit exceeded the condensate, the target left a fraction `1/Γ` of the condensate behind. Only states with `q < deficit < Γ q` change. This commit is independent of the first and can be dropped if you prefer to keep the old bound.
3. Docs: `docs/src/BulkTendencies.md`, `docs/src/MicrophysicsNonEq.md`, and the new functions in `docs/src/API.md`.

## Validation

- Offline probe with the AMIP parameter file (Δt = 180 s, nsub = 3, Float32): the 265 K case now converges monotonically (ΔT per 180 s step +0.18, +0.08, +0.05, +0.02 K); all cold states between 205 and 271 K are damped; the substep trajectory no longer crosses ice saturation.
- Added a stiff relaxation test.
- Coarse AMIP runs for a year with this fix and produces reasonable results.

## Notes for reviewers

- Behavior change for the slow options (`ConstantTimescale`, `TemperatureDependent`): per-substep phase change is reduced by the factor 1/(1+Δt_sub/τ) (≈0.83 for τ = 300 s, Δt_sub = 60 s). Results will differ slightly; the previous treatment was the Δt/τ → 0 limit.
- Independent of numerics: with `cloud_ice_sedimentation_number_concentration = 5e8 m⁻³` the deposition timescale is ≤ 7 s at any T < 0 °C, i.e. `PrescribedIceNumber` acts as an instantaneous saturation adjustment to ice and glaciates supercooled liquid clouds within a step or two (maximal-rate WBF). A dedicated deposition ice-number parameter (or a temperature-dependent one) would decouple this from sedimentation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sajjadazimi (commented):
Allowing negative `e` doesn't guarantee positivity of condensates after solve. When `S` is negative condensate mass can end up at `-|S| dt` after the solve (from all the other processes it can decrease to close zero and from `e` it can go below zero). I think it's safe to only apply the pre-integrated value when it's positive. When the source is negative the rate is something times q anyways (`q^*` is zero for condensate sinks); in this case the linear solver will integrate this rate coupled with other processes to keep positive (in contrast to isolated integration that can make condensate values negative). We can talk about it off-line.

trontrytel:
Closing and continuing in https://github.com/CliMA/CloudMicrophysics.jl/pull/787

PR #787, opened 2026-09-11
Add MM2015 solver for fast vapor-cloud processes

trontrytel:
## Purpose

This PR fixes the numerical instability of the 1-moment non-equilibrium cloud formation when the relaxation timescale is much shorter than the substep of the linearized solver. With `cloud_ice_formation = PrescribedIceNumber` and the default `cloud_ice_sedimentation_number_concentration = 5e8 m⁻³`, the deposition/sublimation timescale is 1.6–7 s against 60 s substeps in the coarse AMIP configuration. The coarse AMIP run crashed with NaNs after ~2 days, while `TemperatureDependent` (τ ≥ 300 s) with the same freezing code was stable.

## Cause

In `_linearized_implicit_step` the instantaneous vapor → condensate rate was applied as a constant over the substep (capped at ice saturation using the substep-start temperature) and the sublimation rate as a decay capped by the available condensate rather than by the vapor deficit. When Δt/τ ≫ 1 both caps were active and their asymmetry produced a period-2 deposition/sublimation flip-flop: deposit all the excess, latent heating leaves the parcel subsaturated, sublimate nearly all the ice, overshoot to supersaturation, repeat.

## Changes

- `src/MicrophysicsNonEq.jl`: new `τ_vap_to_q_lcl(opt, …)` / `τ_vap_to_q_icl(opt, …)` returning the relaxation timescale (without Γ) used by the matching `conv_q_vap_to_q_*` method (`Inf` when disabled). The evaporation/sublimation bound becomes `min(−sat_excess, Γ q)/(τ Γ)`, i.e. the rate is bounded by `q/τ` and the relaxation target is exactly `max(q − deficit/Γ, 0)`.
- `src/BulkMicrophysicsTendencies.jl`: `_microphysics_source_terms` also returns the two timescales. In `_linearize(…, Δt)` the transfer over a substep is the time average of the exact relaxation, `Δq = S Δt φ(Δt/τ)` with `φ(x) = (1 − e^{−x})/x` (Morrison & Milbrandt 2015, Appendix C, eqs. C6–C7; new helper `_relaxation_average_factor`), so the substep never crosses the equilibrium and the instantaneous rate is recovered for Δt ≪ τ. A source enters `e` as a non-negative constant; a sink enters `M` as an implicit decay with coefficient `D = |Δq| /
 (max(q + Δq, q_min) Δt)`, which removes exactly `|Δq|` when acting alone. Following review feedback, sinks are kept implicit so that condensate positivity is structural: every added `M` entry is a non-negative decay and `I/Δt − M` is an M-matrix, hence non-negative inputs give non-negative condensate for any combination of processes, independent of the tendency clip. A plain `S/q` decay was not used because with τ of a few seconds it removes essentially all condensate at the slightest subsaturation and re-creates the oscillating behavior. 
- Docs (`BulkTendencies.md`, `MicrophysicsNonEq.md`, `API.md`) and tests: limits of the averaging factor; stiff deposition/sublimation approach ice saturation without crossing it; a 10 % subsaturated ice cloud with autoconversion and snow closes the deficit, stays non-negative and does not overshoot saturation; two pre-existing structural tests updated to the new signature.

## Validation

- Offline (AMIP parameters, Δt = 180 s, 3 substeps, Float32): the 265 K case converges monotonically (ΔT +0.16, +0.07, +0.06, +0.01 K per step); all states between 205 and 271 K are damped; subsaturated ice clouds with competing sinks close the deficit exactly with no sign changes; no negative condensate anywhere.
- Tests: `bulk_tendencies_tests.jl` 191/191 (Float64, Float32), `microphysics_noneq_tests.jl` 39/39, `type_stability_tests.jl` 130/130, `return_type_tests.jl` (mixed Dual/plain) pass; JuliaFormatter clean.
- Coupled coarse AMIP (ClimaCoupler, `PrescribedIceNumber`, Float32, dt = 180 s, per-step NaN check): a full 366-day year completes with no NaN at 1.99 SYPD on one A100, where the unpatched code fails at 2.2 days. These runs used an earlier iteration of this branch that differed only in how sinks enter the solver; the final form is identical to it on all offline test cases.

## Notes for reviewers

- Results change slightly for the slow options: the transfer per substep is `φ(Δt/τ)` times the instantaneous rate (0.91 for τ = 300 s and 60 s substeps).
- Two properties shared with the existing donor-based linearization: an implicit sink coefficient also acts on mass entering the row in the same substep (freezing into a sublimating ice pool, melting into an evaporating liquid pool), and rain evaporation / snow sublimation keep their constant-rate treatment.
- Physics, separate from this PR: with 5e8 m⁻³ the deposition timescale is ≤ 7 s at any T < 0 °C, so `PrescribedIceNumber` glaciates supercooled clouds within a step or two (liquid fraction 0.5 near 272 K instead of ~258 K in the previous default). A sensitivity sweep over the prescribed number is under way.
- Will need to test if the new version is also stable in AMIP
- A more complete follow-up would adopt the joint supersaturation formulation of MM15 Appendix C, eqs. C1–C7: a single relaxation variable with the combined liquid and ice timescale and the WBF transfer as an explicit liquid-to-ice term, so the two phases' competition for vapor is partitioned within a substep rather than across substeps. Carrying temperature as an unknown is not required for this, since the latent-heat feedback of each phase change is already in Γ and φ. Its one remaining benefit, fusion heat shifting the same-substep saturation targets, is at most a few tenths of a kelvin per substep and is damped by the implicit sinks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/BulkMicrophysicsTendencies.jl
```diff
@@ -262,6 +268,20 @@ here appears with a fixed sign — no `ifelse` branching.
...
+to its initial rate: `φ(x) = (1 - exp(-x)) / x` with `x = Δt/τ` (Morrison &
+Milbrandt 2015, Appendix C, eqs. C6-C7). `φ → 1` for `Δt ≪ τ` (the instantaneous
+rate is recovered) and `φ → τ/Δt` for `Δt ≫ τ` (the transfer saturates at the
+equilibrium amount `S τ`, so the substep never overshoots the equilibrium).
+"""
+@inline function _relaxation_average_factor(x)
+    # `expm1` keeps full accuracy for small `x`; only `x == 0` (τ = Inf) needs the guard.
+    return ifelse(x > zero(x), -expm1(-x) / max(x, eps(typeof(x))), one(x))
+end
+
```
sajjadazimi: Since we need `x phi(x)` maybe it's better to add this and avoid division by zero.

sajjadazimi (approved):
Thanks!

PR #788, opened 2026-09-14
Consistent 2M rain fall speed at zero q or N

haakon-e:
Return zero mass and number fall speeds when either q or N is negligible


trontrytel (approved):
What ended up working in Atmos was actually setting sedimentation velocity to be a non-zero constant for small q_rain: https://github.com/CliMA/ClimaAtmos.jl/pull/4744/changes#r3754247487

That was to prevent the implicit solver from going unstable from the big changes in dw_terminal/dq_rain for small q_rain values. I.e. the flux is negligent, but the derivative was not.

sajjadazimi (approved):
For stability reasons (to avoid shock waves in advection solver) it's useful to have a minimum nonzero terminal velocity. In 1M the assumed minimum terminal velocity is in the order of the terminal velocity of the smallest raindrops (~50 um). Here in 2M maybe an idea is to use the velocity of smallest or largest raindrops, depending on which variable is negligible. For example if only N is negligible then raindrops are probably very large.

haakon-e:
Hmm, interesting analogy from 1M. I'll leave this unmerged for now while doing a bit more testing to see if there are any similar issues like this for me.

PR #789, opened 2026-09-14
Physics-scaled presence tests + AD-safe incomplete gamma

haakon-e:
## Purpose
Replace generic episilon presence conditionals that depend on the floating point type with fixed physical scales.

Improve robustness of some special functions.

## Content
Make `gamma_inc`, `gamma_inc_inv` AD-compatible, and reduce the `Float32` iteration count from 20 to 14 (tested that this is safe).

Add `guarded_quotient`, `nearest_admissible_b` and `clamp_to_nonneg` as the shared
primitives for presence tests and the rime density cone.

Remove `sgs_weight_function` (its functionality is replaced by `guarded_quotient`)

Make `DistributionTools` functions return `NaN` on invalid arguments rather than throwing a `DomainError` (which is not GPU-friendly). Impacts: `generalized_gamma_quantile`, `generalized_gamma_cdf`, `generalized_gamma_Mn`, `exponential_cdf`, and `exponential_quantile`.

Added docs to describe some of the changes in this PR.


src/DistributionTools.jl
```diff
@@ -44,7 +44,10 @@ Calculate the quantile (inverse cumulative distribution function) for a
 function generalized_gamma_quantile(ν, μ, B, Y)
     # Compute the inverse of the regularized incomplete gamma function
     z = UT.gamma_inc_inv((ν + 1) / μ, Y, 1 - Y)
-    return (z / B)^(1 / μ)
+    # Negative base => DomainError of fractional power. Return NaN instead.
+    base = z / B
+    base = ifelse(base ≥ 0, base, oftype(base, NaN))
```
trontrytel, line 49: Do we want to return NaN? Or guard against negative or zero numbers?

docs/src/API.md
```diff
@@ -400,9 +400,10 @@ Utilities.gamma_inc
 Utilities.gamma_inc_inv
 Utilities.rime_mass_fraction
 Utilities.rime_density
-Utilities.sgs_weight_function
```
trontrytel, line 403: Was that ever used in Atmos? If yes, we may need a breaking release here.

src/Utilities.jl
```diff
@@ -20,15 +20,14 @@ export fac
...
-arguments — non-concrete, heap-boxed, and silent.
+Similar to `promote_type`, but accepts concrete variables rather than
+their types. Useful to guarantee a single, concrete return type for functions
+that accept mixed input types (e.g. `Float32` and `ForwardDiff.Dual{Float32}`)
 """
 @inline promote_typeof(args...) = Base.promote_typeof(args...)
 export unrolled_logsumexp
-export sgs_weight_function, rime_mass_fraction, rime_density
+export rime_mass_fraction, rime_density
+export guarded_quotient, nearest_admissible_b
```
trontrytel, line 30: Those are being exported because you are planning to use them in a followup PR?

src/Utilities.jl
```diff
@@ -416,106 +418,95 @@ function unrolled_logsumexp(x)
...
+# Extended help
 
-Mirrors the `sgs_weight_function` in `ClimaAtmos.jl/src/utils/variable_manipulations.jl`.
+The cutoff is independent of `eps(typeof(q_ice))`, which spans nine orders of magnitude
+between `Float32` and `Float64` and would place the result at a different physical scale in each.
+It sits below every admissible ice content, so a small `q_ice` paired with a
+proportionally small `q_rim` returns its true fraction.
+"""
+@inline rime_mass_fraction(q_rim, q_ice, ϵ = oftype(q_ice, 1e-24)) =
+    clamp(_regularised_ratio(q_rim, q_ice, ϵ), zero(q_ice), one(q_ice))
```
trontrytel, line 450: We also have `F_rim = min(UT.rime_mass_fraction(ρq_rim, ρq_ice), one(FT) - eps(FT))` in `P3_particle_properties`. But maybe not needed?

trontrytel (approved):

docs/src/API.md
haakon-e: No. ClimaAtmos had its own version of this function (I copied it from there), but it turns out that the smooth weighting works really poorly for `q/N`-type fractions. I'm happy to make it a breaking release if that's safest, but really, I don't think anyone has used this.

src/Utilities.jl
haakon-e: that's right.

docs/src/API.md
trontrytel: If nobody was using it, then it's fine. No worries

